PR #29, opened 2018-10-10
Remember the Weekday

wjmelements:
#### Changes
This trims an unnecessary call to getWeekday by keeping the result on the stack.
The compiler might already optimize this because `getWeekday` is a pure function.
Reviewers @terryli0095


wjmelements:
Tests pass for 6f9c0cf

wjmelements:
I have verified that this reduces gas cost. As of 0.4.25, pure functions are not yet optimized in this way.

PR #30, opened 2018-10-10
fix list formatting in README

zachlysobey:
I *think* that this was intended to be a markdown list.
I just added a little whitespace to make it happen.


wjmelements:
I think you're right, though I'm changing these instructions in https://github.com/trusttoken/trueUSD/pull/27

PR #32, opened 2018-10-15
Commit Deployed Contract

wjmelements:
#### Changes
This adds a flattened copy of the deployed contract under `contracts/deployed`.
Reviewers @terryli0095


zachlysobey:
I'm curious @wjmelements, is this a *generated* file? As a rule of thumb, I think it is usually preferable to *not* version control generated files...

zachlysobey:
- I totally agree that this is very useful!
- is there danger of this code becoming out-of-date?
- would it make sense to instead provide instructions or a link to get it directly from etherscan.io?

But... I defer to you on this of course :)

PR #34, opened 2018-10-15
update license field in package.json

zachlysobey:
I noticed that the package.json listed "ISC" as the license (the `npm init` default), but there is an included MIT license. I've updated the package.json to reference that LICENSE file per suggestion here: https://docs.npmjs.com/files/package.json\#license


zachlysobey:
Closing b/c no one is paying attention apparently :)

PR #55, opened 2019-01-17
0x100000 Redemption Addresses

wjmelements:
#### Changes
This allocates many redemption addresses for us to use instead of having us increment a counter one at a time.
Which ones are active is maintained by the registry.
We can modify the constant by upgrading if we need more.
Reviewers @terryli0095


wjmelements:
There is a merge conflict with #54

PR #56, opened 2019-01-17
flatten-all

wjmelements:
#### Changes
This creates and runs a script, `flatten-all`.
Reviewers @terryli0095


wjmelements:
Holding this until #57

PR #59, opened 2019-01-26
Patch RefundPrice Typo

wjmelements:
#### Changes
I missed a row when doing the replace.
Reviewers @terryli0095


wjmelements:
test passes

PR #60, opened 2019-01-26
GT 2

wjmelements:
#### Changes
Instead of checking NOT EQ 0 we will check GT 2, which is marginally cheaper.
This should behave the same otherwise.
A one-sided check is also safer for unexpected negative lengths.
Reviewers @terryli0095


wjmelements:
You would hope that the optimizer would implement unsigned `!= 0` as `>0` but there's only one way to be sure.

PR #62, opened 2019-02-01
Remove duplicate SSTORES

wjmelements:
#### Changes
Thanks @terryli0095 for finding a way to do this without assembly.
Turns out solidity already zeroes array members when you subtract length.
Therefore we do not need to set zeroes beforehand.
Reviewers @terryli0095


wjmelements:
https://etherscan.io/tx/0xf24a6df4376fda4a15a96f2e9d9dd817e89f2ee6f355785b0021fcce717a3ffb
https://etherscan.io/tx/0x45f89d5a92251df5c77d7440c787fbca8eb042c99e909ad89027a3591b9ee0b3

PR #70, opened 2019-02-20
Profile transferFrom

wjmelements:
#### Changes
This adds measurements for 16 transferFrom scenarios, and for burn transfers.
Does not test autosweep.
Reviewers @terryli0095


terryli0095:
This is sweet

PR #74, opened 2019-02-21
Update to Registry#16

wjmelements:
#### Changes
This updates the registry submodule.
Reviewers @terryli0095


contracts/mocks/OtherMocks.sol
```diff
@@ -1,5 +1,7 @@
 pragma solidity ^0.4.23;
 
+import "../../registry/contracts/mocks/RegistryMock.sol";
```
wjmelements: This is a hack to get the RegistryMock artifact

PR #89, opened 2019-04-10
rm Unused Flag Var and use uint256 for outvar in subAllowance

wjmelements:
#### Changes
Removing an unused variable saves gas, because solc
Moving the comparison out of the internal function also saves gas, because solc. We were already doing this for balances.

Reviewers @terryli0095


terryli0095:
what does the compiler even do....

PR #90, opened 2019-04-10
Flatten Proxy

wjmelements:
#### Changes
This flattens the proxy into one file, which allows optimization of the fallback function.
The optimization removes internal functions, which have considerable overhead.
It also uses returndatasize multiple times instead of keeping a stack variable. returndatasize is very cheap; cheaper than dup.
The other internal functions can stay because they do not affect the cost of the fallback function.

Reviewers @terryli0095


terryli0095:
wait hold on let me read more

wjmelements:
I'll rerun the flattener with #89.

PR #93, opened 2019-04-19
implementation public view

wjmelements:
#### Changes
In #90 I accidentally botched the visibility of the implementation getter for the proxy.
Reviewers @terryli0095


wjmelements:
Surprisingly, this does affect the bytecode.

PR #96, opened 2019-04-29
JUMPI Sheep

wjmelements:
#### Changes
This modifies the Sheep contract so that they use JUMPI to assert instead of REVERT to revert.
We do not care about the gas of those who would improperly call our sheep contracts.
This saves 2416 gas in sponsorGas2, and 2 (1) gas in gasRefund39.
I also add tests to verify that the Sheep assert if you call them. 
Reviewers @terryli0095

### DIFF
--- a/GasProfile.json
+++ b/GasProfile.json
@@ -21,12 +21,12 @@
   "emptyApprovalReducingToBurn": 34596,
   "reduceApprovalEmptyingToBurn": 34532,
   "emptyApprovalEmptyingToBurn": 24676,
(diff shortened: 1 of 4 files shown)


wjmelements:
The 2 gas savings for us in the refund comes from changing PUSH1 to PC, which saves 1 gas, and removing JUMPDEST, which saves 1 gas.

The 4-byte reduction in sheep size saves 800 gas per sheep. The remaining 16 comes from removing the `sheep` local variable from the assembly.

terryli0095:
tested

PR #104, opened 2019-08-01
[WIP] Fallback Registrar

wjmelements:
#### Changes
This begins creation of the fallback registrar.
TODO testing
Reviewers @terryli0095


wjmelements:
I don't think this implementation would work if called from a constructor context. It would be nice if we could let that work.

wjmelements:
Perhaps if we had both register and deregister we could skip the codesize check.

PR #130, opened 2020-03-03
iEarn financial opportunity

Marik-D:


contracts/TrueReward/IEarnFinancialOpportunity.sol
```diff
@@ -1,4 +1,84 @@
...
+
+    function configure(
+        yTrueUSDInterface _yToken,
+        TrueRewardBackedToken _token
+    ) public onlyOwner {
+        yToken = _yToken;
+        token = _token;
+    }
+
+    function proxyOwner() public view returns(address) {
```
wjmelements: This won't work

contracts/TrueReward/IEarnFinancialOpportunity.sol
```diff
@@ -1,4 +1,84 @@
...
+        return OwnedUpgradeabilityProxy(address(this)).proxyOwner();
```
wjmelements: Also, you shouldn't staticcall into yourself if you don't have to; it's gas-intensive

contracts/TrueReward/IEarnFinancialOpportunity.sol
```diff
@@ -1,4 +1,84 @@
...
+    }
+
+    function deposit(address _account, uint256 _amount) external returns(uint256) {
+        require(token.transferFrom(_account, address(this), _amount), "transfer from failed");
+        require(token.approve(address(yToken), _amount), "approve failed");
+        
+        uint256 balanceBefore = yToken.balanceOf(address(this));
+        yToken.deposit(_amount);
```
wjmelements: it would surprise me if this did not return the number of shared minted

terryli0095: it does not

terryli0095: we can do the calculation on our end, but calling balance of is potentially safer

PR #168, opened 2020-04-11
deploy

hal909:


contracts/deploy/UpgradeHelper2
```diff
@@ -0,0 +1,112 @@
+pragma solidity ^0.5.13;
```
Marik-D, line 1: no extension on this file

migrations/3_deploy_truerewards.js
```diff
@@ -0,0 +1,66 @@
+var TrueUSD = artifacts.require("TrueUSD");
```
Marik-D, line 1: Probably worth removing this file since all the code is commented out

contracts/deploy/UpgradeHelper2
Marik-D: Remove this file if not needed

scripts/deploy_testnet.js
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * We use waffle to deploy our contracts from scratch.
+ * For upgrades, use deploy/upgrade.js
+ * Use the config object to set paramaters for deployment
+ */
+
+// this might be unsafe. We want to use ethereum accounts instead of hardcode pk
+const config = {
+    rpc: 'localhost:7585',
+    accountPrivateKey: 'ac74a462679b69b28f5c3c124eda5cc521a0d3d5ca9b0877f9e500ed94c24414'
```
Marik-D, line 13: probably we shouldn't have private keys commited in code

Marik-D: we can just use program arguments to pass it for now

PR #180, opened 2020-04-17
Liquidation and award pool tests for assured financial opportunity

Marik-D:


test-ts/AssuredFinancialOpportunity.test.ts
```diff
@@ -215,7 +218,116 @@ describe('AssuredFinancialOpportunity', () => {
     })
   })
 
-  it.skip('liquidation', async () => {
+  it('liquidation', async () => {
```
Marik-D, line 221: add partial liquidation tests

PR #196, opened 2020-04-30
refactor financial opportunity

hal909:


contracts/TrueCurrencies/AssuredFinancialOpportunityStorage.sol
```diff
@@ -18,8 +18,8 @@ contract AssuredFinancialOpportunityStorage {
     // we change the adjustment factor
     uint adjustmentFactor;
 
-    // minPerTokenValue can never decrease
-    uint minPerTokenValue;
+    // mintokenValue can never decrease
```
Anze1m, line 21:
```suggestion
    // minTokenValue can never decrease
```

contracts/TrueCurrencies/AssuredFinancialOpportunityStorage.sol
```diff
@@ -18,8 +18,8 @@ contract AssuredFinancialOpportunityStorage {
...
+    uint mintokenValue;
```
Anze1m, line 22:
```suggestion
    uint minTokenValue;
```

Anze1m: Btw where is this value actually set?

PR #205, opened 2020-05-07
whitelist

hal909:


contracts/TrueCurrencies/utilities/TokenFaucet.sol
```diff
@@ -56,12 +56,19 @@ contract TokenFaucet {
     }
 
     function faucet(uint256 _amount) external {
+        //
```
krzysztof-jelski, line 59: Maybe worth naming the attribute we set here.

contracts/TrueCurrencies/utilities/TokenFaucet.sol
```diff
@@ -56,12 +56,19 @@ contract TokenFaucet {
     }
 
     function faucet(uint256 _amount) external {
+        // 
         registry.setAttributeValue(msg.sender, 0x6861735061737365644b59432f414d4c00000000000000000000000000000000, 1);
+        // whitelist trueRewards
+        registry.setAttributeValue(msg.sender, 0x6973547275655265776172647357686974656c69737465640000000000000000, 1);
```
krzysztof-jelski, line 62: Perhaps extract constant.

PR #253, opened 2020-07-07
Migrate to solidity 0.6, merge with trusttokens

vanruch:


test-ts/aave/.waffle.json
```diff
@@ -0,0 +1,11 @@
+{
+    "compilerVersion": "v0.5.13+commit.5b0b510c",
+    "compilerOptions": {
+        "optimizer": {
```
krzysztof-jelski, line 4: As it's for tests only, we can probably do without optimizer.

contracts/TrueReward/mocks/LendingPoolCoreMock.sol
```diff
@@ -1,20 +1,24 @@
...
 
 import "../ILendingPoolCore.sol";
-import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
 contract LendingPoolCoreMock is ILendingPoolCore {
     uint256 reserveNormalizedIncome = 1*10**27;
 
-    function getReserveNormalizedIncome(address _reserve) external view returns (uint256) {
+    function getReserveNormalizedIncome(address _reserve) override external view returns (uint256) {
```
krzysztof-jelski, line 10:
```suggestion
    function getReserveNormalizedIncome(addres) override external view returns (uint256) {
```

contracts/TrueCurrencies/TrueRewardBackedToken.sol
```diff
@@ -138,10 +138,10 @@ contract TrueRewardBackedToken is RewardTokenWithReserve {
         // require TrueReward is enabled
         require(trueRewardEnabled(msg.sender), "TrueReward already disabled");
         // get balance
-        uint rewardBalance = rewardTokenBalance(msg.sender, opportunity());
+        uint256 rewardBalance = rewardTokenBalance(msg.sender, opportunity());
 
         // remove reward distribution
-        _removeDistribution(opportunity());
+        _removeDistribution();
```
hal909, line 144: Is there a reason this function is changed? we will want to support multiple opportunities in the future. / is this relevant to this pull request

contracts/TrueCurrencies/TrueRewardBackedToken.sol
```diff
@@ -323,16 +326,14 @@ contract TrueRewardBackedToken is RewardTokenWithReserve {
...
     }
 
     /**
      * @dev Remove reward distribution for a financial opportunity
      * Remove
      */
-    function _removeDistribution(address finOp) internal {
-        delete _rewardDistribution[msg.sender][0];
-        _rewardDistribution[msg.sender].length--;
+    function _removeDistribution() internal {
```
hal909, line 336: can we create a separate PR for this change? try to keep this relevant to only migration & mono repo.

contracts/TrueCurrencies/TrueUSD.sol
```diff
@@ -28,7 +28,36 @@ contract TrueUSD is TrueRewardBackedToken, DelegateERC20 {
         return "TUSD";
     }
 
-    function canBurn() internal pure returns (bytes32) {
+    function canBurn() internal override pure returns (bytes32) {
         return "canBurn";
     }
+
+    function _transferAllArgs(
```
hal909, line 35: Is this required for the migration to 0.6 or can we avoid having to implement these functions again? If we can avoid it, add these to a new PR with the other changes to TrueUSD / TrueRewardBackedToken

contracts/TrueReward/AssuredFinancialOpportunity.sol
```diff
@@ -187,16 +188,14 @@ contract AssuredFinancialOpportunity is FinancialOpportunity, AssuredFinancialOp
     function setRewardBasis(uint32 newBasis) external onlyOwner {
         minTokenValue = _tokenValue();
 
-        adjustmentFactor = adjustmentFactor
-            .mul(_calculateTokenValue(rewardBasis))
-            .div(_calculateTokenValue(newBasis));
+        adjustmentFactor = adjustmentFactor.mul(_calculateTokenValue(rewardBasis)).div(_calculateTokenValue(newBasis));
```
hal909, line 191: Is there a benefit to making this one line? might be easier to read how it was before. Not a major issue though.

contracts/TrueReward/mocks/ATokenMock.sol
```diff
@@ -1,10 +1,11 @@
-pragma solidity 0.5.13;
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.6.10;
 
-import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
+import { ERC20Mock } from "./ERC20Mock.sol";
```
hal909, line 4: How is the mock different than the actual ERC-20?

contracts/TrueReward/utilities/IExponentContract.sol
```diff
@@ -1,9 +0,0 @@
-pragma solidity 0.5.13;
```
hal909, line 1: We probably want to preserve the IExponentContract and have FractionalExponents extend from this.. We want to be able to swap out the existing exponent contract if we decide to use a new one.

hal909: and make it an interface

contracts/deploy/DeployHelper.sol
```diff
@@ -167,42 +160,19 @@ contract DeployHelper {
...
         liquidator = Liquidator(address(liquidatorProxy));
 
-        liquidator.configure(
-            address(registry),
-            address(trueUSDProxy),
-            address(trustTokenProxy),
-            outputUniswapAddress,
-            stakeUniswapAddress
-        );
+        liquidator.configure(address(registry), address(trueUSDProxy), address(trustTokenProxy), outputUniswapAddress, stakeUniswapAddress);
```
hal909, line 169: Any reason for making these one line?

test-ts/aave/.waffle.json
hal909: I think we want the tests to reflect the same gas usage as with the optimizer, so we can record gas costs in tests. Ultimately we want to replace the gas profiler with a new one , but for now I think it makes sense to just keep this. tests in typescript don't use the Gas Profile though.

contracts/TrueCurrencies/TrueRewardBackedToken.sol
vanruch: The compiler started to warn loudly about unused arguments in newest version. And we don't really have any other place where we are ready for changing for multiple finOps

contracts/TrueCurrencies/TrueUSD.sol
vanruch: Sadly yes, it's required. TUSD extends `TrueRewardBackedToken` and `DelegateERC20` both of which implement have  `_transferAllArgs`. Solidity 0.6 wouldn't compile without this change

contracts/TrueReward/AssuredFinancialOpportunity.sol
vanruch: Linter did it.

contracts/TrueReward/mocks/ATokenMock.sol
vanruch: openzeppelin ERC20 doesn't have virtual balanceOf, otherwise they are same

contracts/deploy/DeployHelper.sol
vanruch: Changed prettify settings a bit, looks good now

vanruch:
> 1. Let's use lowercase folder names in the contracts subfolder:
👍 
> 2. Separate the truffle tests into sub-folders relative to the contract folders. For example:
Maybe not in this PR. It's enormous enough already
> 3, 4, 5, 6, 7
👍

PR #269, opened 2020-07-24
Add script to register SAFT accounts

vanruch:


scripts/register_saft_addresses.ts
```diff
@@ -0,0 +1,60 @@
...
+import { TrustTokenFactory } from '../build/types/TrustTokenFactory'
+
+export const txnArgs = { gasLimit: 2_000_000, gasPrice: 20_000_000_000 }
+
+export interface SaftAccount {
+  address: string,
+  amount: string,
+}
+
+const toTrustToken = (amount: string) => utils.parseUnits(amount, 8)
```
Anze1m, line 21: `toTrustToken`, but from what? Maybe sth like `weiToTrustToken`?

scripts/register_saft_addresses.ts
```diff
@@ -0,0 +1,60 @@
...
+const sum = (numbers: utils.BigNumber[]) => numbers.reduce((cumSum, value) => cumSum.add(value), constants.Zero)
```
Anze1m, line 22: `cumSum` doesn't sound fortunate. Maybe simply sth like this?
```suggestion
const sum = (numbers: utils.BigNumber[]) => numbers.reduce((a, b) => a.add(b), constants.Zero)
```

scripts/register_saft_addresses.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const registerSaftAccounts = async (registry: TimeLockRegistry, trustToken: TrustToken, accounts: SaftAccount[]) => {
+  const totalAllowance = sum(accounts.map(({ amount }) => toTrustToken(amount)))
+  const tx = await trustToken.approve(registry.address, totalAllowance, txnArgs)
+  await tx.wait()
+  console.log('Transfers approved')
+  let { nonce } = tx
+  const pendingTransactions = []
+  for (const { address, amount } of accounts) {
+    pendingTransactions.push((await registry.register(address, toTrustToken(amount), { ...txnArgs, nonce: nonce + 1 })).wait()
```
Anze1m, line 32:
```suggestion
    pendingTransactions.push((await registry.register(address, toTrustToken(amount), { ...txnArgs, nonce: nonce++ })).wait()
```
and remove line 36?

scripts/register_saft_addresses.ts
```diff
@@ -0,0 +1,60 @@
...
+    .map((line) => ({
+      address: line.split(',')[0].trim(),
+      amount: line.split(',')[1].trim(),
+    }))
+
+const readAccountList = (filePath: string) => parseAccountList(fs.readFileSync(filePath).toString())
+
+if (require.main === module) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const deployedAddresses = require(`./deploy/${process.argv[2]}.json`)
```
Anze1m, line 54: Why not read it with `fs`?

test-ts/utils/parseTT.ts
```diff
@@ -1,3 +1,3 @@
 import { utils } from 'ethers'
 
-export const parseTT = (amount: number) => utils.bigNumberify(amount).mul(10 ** 8)
+export const parseTT = (amount: utils.BigNumberish) => utils.parseUnits(amount.toString(), 8)
```
Anze1m, line 3: Isn't it essentially the same as `toTrustToken`? Maybe we should remove `toTrustToken` and use this one instead?

Anze1m: And still maybe try to come up with more descriptive name

scripts/register_saft_addresses.ts
vanruch: cumsum is widely used https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/cumsum, https://numpy.org/doc/stable/reference/generated/numpy.cumsum.html

the word might be funny though

scripts/register_saft_addresses.ts
vanruch: The problem is, can we call it wei when we don't have 18 digits precision? It's kinda weird that community hasn't come with any universal name for minimal token unit

scripts/register_saft_addresses.ts
vanruch: idk, i think your version is a bit harder to read

scripts/register_saft_addresses.ts
vanruch: require also does JSON parsing for us

test-ts/utils/parseTT.ts
vanruch: Yeah, the problem is I didn't want to import from test and there's no a good place for utils function otherwise. Maybe moving to script/utils is a good idea

Anze1m: We may have it in some main, non-test directory and import from there.

scripts/register_saft_addresses.ts
Anze1m: Maybe... 😛

scripts/register_saft_addresses.ts
Anze1m: It does, but it also forces us to add linter exception. Reading all files in the same manner and then parsing them imo would be prettier.

scripts/register_saft_addresses.ts
Anze1m: Still `(a, b) => a.add(b)` imo looks better.

vanruch: I never remember which argument is which in reduce. But in this case the order doesn't matter, so it's ok I guess

scripts/register_saft_addresses.ts
Anze1m: Right... Quantum sounds like a proper word, but using it might actually cause even more confusion as it's not a common idea. So maybe let's keep `toTrustToken`, but move it to some shared utils directory and import from there. And actually have the version with `BigNumberish` instead of `string`.

PR #315, opened 2020-10-08
trusttoken multiple distributions

hal909:


contracts/trusttoken/TimeLockedToken.sol
```diff
@@ -95,10 +95,9 @@ abstract contract TimeLockedToken is ERC20, ClaimableContract {
      */
     function registerLockup(address receiver, uint256 amount) external onlyTimeLockRegistry {
         require(balanceOf[msg.sender] >= amount, "insufficient balance");
-        require(distribution[receiver] == 0, "distribution already set");
 
-        // set distribution to lockup amount
-        distribution[receiver] = amount;
+        // add amount to locked distribution
+        distribution[receiver] = distribution[receiver].add(amount);
```
vanruch, line 100: Looks like comment on line 92 should be removed.

vanruch:
Looks good, but could you add some tests for the case when distribution is added during n-th period (n>1). It probably will pass but it's good to be sure

vanruch:
#324 points to dev

PR #322, opened 2020-10-20
Add whitelist to loan transfers

vanruch:


contracts/truefi/LoanToken.sol
```diff
@@ -15,15 +15,19 @@ contract LoanToken is ILoanToken, ERC20 {
     uint256 public override apy;
 
     uint256 public start;
+    address public funder;
```
vanruch, line 18: Maybe `lender` is a better name?

hal909 (approved):

PR #331, opened 2020-10-26
Return total debt value

barrutko:


contracts/truefi/TrueLender.sol
```diff
@@ -111,9 +112,32 @@ contract TrueLender is Ownable {
...
+    function valueFor(ILoanToken loanToken) public view returns (uint256) {
+        (uint256 amount, uint256 apy, uint256 duration) = loanToken.getParameters();
+        uint256 start = loanToken.start();
+        uint256 passed = block.timestamp.sub(start);
+        if (passed > duration) {
+            passed = duration;
+        }
+
+        uint256 helper = amount.mul(apy).mul(passed).mul(loanToken.balanceOf(address(this)));
+        uint256 interest = helper.div(360 days).div(10000).div(loanToken.totalSupply());
```
vanruch, line 128: Just have it in one expression.

contracts/truefi/TrueLender.sol
```diff
@@ -111,9 +112,32 @@ contract TrueLender is Ownable {
         pool.borrow(amount);
         currencyToken.approve(address(loanToken), amount);
         loanToken.fund();
+        loans.push(loanToken);
         emit Funded(address(loanToken), amount);
     }
 
+    function valueFor(ILoanToken loanToken) public view returns (uint256) {
+        (uint256 amount, uint256 apy, uint256 duration) = loanToken.getParameters();
```
vanruch, line 120: Mb it's worth adding a check that loan is funded and return 0 otherwise? It looks like unnecessary, but just in case logic changes somewhere in contract.

vanruch: And overall, checking for LT total supply looks like a good idea to avoid div by 0 exception

contracts/truefi/TrueLender.sol
```diff
@@ -111,9 +112,32 @@ contract TrueLender is Ownable {
...
+            passed = duration;
+        }
+
+        uint256 helper = amount.mul(apy).mul(passed).mul(loanToken.balanceOf(address(this)));
+        uint256 interest = helper.div(360 days).div(10000).div(loanToken.totalSupply());
+
+        return amount.add(interest);
+    }
+
+    function value() external view returns (uint256) {
```
vanruch, line 133: totalLoanValue?

contracts/truefi/TrueLender.sol
```diff
@@ -111,9 +112,32 @@ contract TrueLender is Ownable {
...
+        uint256 totalValue;
+        for (uint256 index = 0; index < loans.length; index++) {
+            totalValue = totalValue.add(valueFor(loans[index]));
+        }
+        return totalValue;
+    }
+
     function reclaim(ILoanToken loanToken) external onlyOwner {
```
vanruch, line 141: I think we should remove loans from array here as well

contracts/truefi/TrueLender.sol
barrutko: I had it like this, but then I get `stack too deep` error.

contracts/truefi/TrueLender.sol
barrutko: I'd stick to the simple, one word name on the interfaces.

contracts/truefi/TrueLender.sol
barrutko: Let's add this in next PR.

contracts/truefi/TrueLender.sol
barrutko: Generally loan is added to the array when it's funded. I guess checking if the supply is not 0 should be enough.

vanruch: Better check if balance of not 0

vanruch (approved):

PR #343, opened 2020-11-01
Move method evaluating loan value to LoanToken

barrutko:


contracts/truefi/LoanToken.sol
```diff
@@ -108,6 +108,22 @@ contract LoanToken is ILoanToken, ERC20 {
         return (amount, apy, duration);
     }
 
+    function value(address holder) external override view returns (uint256) {
```
vanruch, line 111: I think passing holder address as a param doesn't make much sense here. I'd look moge logical if we passed here LoanToken amount and it'd return it's value in TUSD

vanruch (approved):

PR #361, opened 2020-11-03
Improve importing compilation products

barrutko:


package.json
```diff
@@ -32,6 +32,7 @@
   "devDependencies": {
     "@typechain/ethers-v5": "^1.0.0",
     "@types/chai": "^4.2.11",
+    "@types/expect": "^24.3.0",
```
vanruch, line 35: what's this?

vanruch (approved):

package.json
barrutko: Actually I can remove it

barrutko: Leftovers from attempt to use `ts-mocha`

PR #379, opened 2020-11-24
Introduce TRU reward multiplier

hal909:


contracts/truefi/TrueRatingAgency.sol
```diff
@@ -393,7 +406,8 @@ contract TrueRatingAgency is ITrueRatingAgency, Ownable {
             // calculate reward
             // prettier-ignore
             uint256 reward = toTrustToken(interest.mul(
-                distributor.remaining()).div(distributor.amount()));
+                distributor.remaining()).div(distributor.amount()))
+                .mul(rewardMultiplier);
```
vanruch, line 410: Maybe it's better to do `.mul(rewardMultiplier)` before `.div(distributor.amount()))` to win a bit of precision

vanruch (approved):

contracts/truefi/TrueRatingAgency.sol
jakub-wojciechowski: Yes, it's always better to divide in the last step.

contracts/truefi/TrueRatingAgency.sol
```diff
@@ -396,8 +409,10 @@ contract TrueRatingAgency is ITrueRatingAgency, Ownable {
 
             // calculate reward
             // prettier-ignore
-            uint256 reward = toTrustToken(interest.mul(
-                distributor.remaining()).div(distributor.amount()));
+            uint256 reward = toTrustToken(interest
+                .mul(distributor.remaining())
+                .mul(rewardMultiplier)
+                .div(distributor.amount()));
```
vanruch, line 415: move ); to next line and remove prettier-ignore
Should make it more readable

PR #386, opened 2020-12-02
Enhance CRV Uniswap integration

barrutko:


contracts/truefi/TrueFiPool.sol
```diff
@@ -338,24 +338,30 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
     }
 
     /**
-     * @dev Collect CRV tokens minted by staking at gauge and sell them on Uniswap
+     * @dev Collect CRV tokens minted by staking at gauge
+     */
+    function collectCrv() external onlyOwner {
```
vanruch, line 343: Is onlyOwner necessary here?

barrutko: I'd keep it under control just in case.

vanruch (approved):

PR #395, opened 2020-12-09
Prevent borrower from repaing with surplus

MiksuJak:


contracts/truefi/LoanToken.sol
```diff
@@ -301,6 +301,9 @@ contract LoanToken is ILoanToken, ERC20 {
      * @param _amount amount of currencyToken to repay
      */
     function repay(address _sender, uint256 _amount) external override onlyAfterWithdraw {
+        if (_amount > debt) {
```
vanruch, line 304: Revert in this case. It's better to throw mistake than quitely fix it and lead to user's confusion.

barrutko: I agree. Especially considering the fact, that this amount will not move unexpectedly and frontend app can fetch it and pass accurately.

MiksuJak:
Requested changes have just come. 
May I also suggest changing a condition in close() function?
`
function close() external override onlyOngoing {
        require(start.add(term) <= block.timestamp, "LoanToken: Loan cannot be closed yet");
        if (_balance() >= debt) {
            status = Status.Settled;
        } else {
            status = Status.Defaulted;
        }

        emit Closed(status, _balance());
    }
`
Considering the fact, that now repaying with surplus is impossible, we can strictly check if contract balance equals the debt.
From my initial research, this might require to rewrite a few tests.

vanruch:
> Requested changes have just come.
> May I also suggest changing a condition in close() function?
> 
> ```
> function close() external override onlyOngoing {
>         require(start.add(term) <= block.timestamp, "LoanToken: Loan cannot be closed yet");
>         if (_balance() >= debt) {
>             status = Status.Settled;
>         } else {
>             status = Status.Defaulted;
>         }
> 
>         emit Closed(status, _balance());
>     }
> ```
> 
> Considering the fact, that now repaying with surplus is impossible, we can strictly check if contract balance equals the debt.
> From my initial research, this might require to rewrite a few tests.

It's still possible, by doing a simple transfer. We cannot prevent this really

vanruch (approved):

PR #397, opened 2020-12-09
End distributions when empty() is called

vanruch:


contracts/truefi/distributors/LinearTrueDistributor.sol
```diff
@@ -126,10 +126,13 @@ contract LinearTrueDistributor is ITrueDistributor, Ownable {
 
     /**
      * @dev Withdraw funds (for instance if owner decides to create a new distribution)
-     * Distributes remaining funds before withdrwaing
+     * Distributes remaining funds before withdrawing
+     * Ends current distribution
      */
     function empty() public override onlyOwner {
         distribute();
+        distributed = 0;
```
barrutko, line 134: Should distributed be set to 0?

contracts/truefi/distributors/ArbitraryDistributor.sol
```diff
@@ -67,12 +69,15 @@ contract ArbitraryDistributor is IArbitraryDistributor, Ownable {
     function distribute(uint256 _amount) public override onlyBeneficiary {
         remaining = remaining.sub(_amount);
         require(trustToken.transfer(msg.sender, _amount));
+
+        emit Distributed(_amount);
```
barrutko, line 73: 👍

barrutko (approved):
`distributed = 0;` looks a bit confusing, but apart from that looks fine 👌

contracts/truefi/distributors/LinearTrueDistributor.sol
vanruch: If it's not, this line would throw `totalAmount.sub(distributed);`

PR #399, opened 2020-12-10
Restrict loan funding to lender

vanruch:


contracts/truefi/LoanFactory.sol
```diff
@@ -40,6 +42,10 @@ contract LoanFactory is ILoanFactory, Initializable {
         currencyToken = _currencyToken;
     }
 
+    function setLender() external {
+        lender = 0x16d02Dc67EB237C387023339356b25d1D54b0922;
```
barrutko, line 46: Shouldn't we have some more parametrised setter?

vanruch: No, that would be dangerous as it could be front-ran. And we don't have any kind of ownership to protect from it

barrutko (approved):

PR #401, opened 2020-12-10
Fix non-deterministic test

barrutko:


MiksuJak (approved):
I even counted amount of zeros. Shouldn't we switch to math power notation in tests, so they are more readable?

PR #408, opened 2020-12-14
Increase readability of many zeros numbers

MiksuJak:


test/truefi/TrueFarm.test.ts
```diff
@@ -35,12 +35,14 @@ describe('TrueFarm', () => {
   let farm2: TrueFarm
   let start: number
 
+  const toBigNumberString = (base: number, magnitude: number) => base.toString() + '0'.repeat(magnitude)
```
MiksuJak, line 38: Waiting for suggestions, for shorter yet descriptive name for this function.

vanruch: ethers gives us `utils.parseUnits`, `utils.parseEther` functions and I think, it's good enough to use it everywhere

MiksuJak: It is meant to be used inside parseEther/ parseTru. parseEther('100000000000') still remains quite unreadable

PR #412, opened 2020-12-15
Governance Upgrade

yubo-ruan:
### Upgrade Notes: 

- [x] **Truffle compile passed**

- [ ] **Testing**

### VoteToken

- [x] use solidity 0.6.10 version
- [x] override _transfer function (add _moveDelegates)
- [x] fix bugs with keccak256(abi.encodePacked())
- [x] abstract contract VoteToken is TimeLockedToken 
- [x] uint96(balanceOf[delegator]) explicit conversion
- [x] unlockedBalance(delegator)

### ProxyStorage

**Add additional storage of** 

- [x] mapping delegates
- [x] struct Checkpoint
- [x] mapping checkpoints
- [x] mapping numCheckpoints
- [x] mapping nonces

### Timelock.sol

- [x] using solidity ^0.6.10
- [x] fix fallback() external payable { }

### TrustToken

- [x] Inherit from VoteToken

### GovernorAlpha.sol

- [x] name change to  "TrustToken Governor Alpha"
- [x] using solidity ^0.6.10
- [x] add MIT License
- [x] quorumVotes changes to 58000000e8      (4% of Tru)
- [x] proposalThreshold changes to 14500000e8      (1% of Tru)
- [x] fix TrustTokenInterface and trustToken instance

### Notice:
GovernorAlpha.sol, Timelock.sol, and part of VoteToken are forked from [Compound](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/Comp.sol)

### DIFF
--- a/contracts/governance/GovernorAlpha.sol
+++ b/contracts/governance/GovernorAlpha.sol
@@ -1,22 +1,28 @@
-// SPDX-License-Identifier: MIT
+// COPIED FROM https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/GovernorAlpha.sol
+// Copyright 2020 Compound Labs, Inc.
+// Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+// 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+// 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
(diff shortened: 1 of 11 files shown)


vanruch:
Great work! 🚀 

Just have some notes
1. We do development on `dev` branch, so you should probably rebase and change base in PR
2. I'm a bit concerned about licencing. You copied a lot of code from Compound and changed licence from BSD-4 to MIT. We should double check if that's not a problem and whether we need to add things like original authors in comments
3. Please list all changes that were made to original Comp contracts (may be here in PR, may be in code comments) and write tests for them if necessary
4. I would prefer to avoid upgrading TrustToken contract. Contract upgrades always bring some risk. For example MKR has no mentions of governance in token code. I don't really know how we could avoid it while still using Comp governance though
5. Do we need these `@title, @author` comments before each contract? IMO, they don't bring any additional value and only take up space. There's always git blame :) 
6. All tests should be in TS, not JS

barrutko:
I agree with everything Ivan said.

Additionally:
1. I'd add tests for all newly introduced smart contracts. If some of them are taken from other codebases, their tests should be also ported here.
2. You should contact someone with admin rights and rather open PR within the repo - not from the fork, so the CI can run.
3. I'd enhance revert messages structure to be more consistent with the rest of the contracts. (Let's stick to `"ContractName: Actual reason of the revert"`).
4. I think that governance should be located in separate subdirectory. So I'd change contracts' location from `contracts/trusttoken/governance` to just `contracts/governance`.
5. We might want to avoid using `pragma experimental ABIEncoderV2;`. AFAIK it causes a lot of pain with contract verification.

yubo-ruan:
update to governance upgrade new PR

PR #417, opened 2020-12-16
Resolve pending test

MiksuJak:


MiksuJak:
Previously the owner was checking his approval, which is infinite by default, so none of those tests could result in false.

barrutko (approved):
Noice. 🚀

vanruch (approved):
tenor

PR #418, opened 2020-12-17
Treat year as 365 days

MiksuJak:


barrutko (commented):
This should be also changed (`TrueLender.sol`)
```
    function loanIsCredible(
        uint256 apy,
        uint256 term,
        uint256 yesVotes,
        uint256 noVotes
    ) public view returns (bool) {
        return apy.mul(term).mul(yesVotes).div(360 days) >= noVotes.mul(riskAversion);
    }
```

MiksuJak:
Also a thing about **loanIsCredible**. That's the one function that is actually missing in tests. Maybe it is worth adding some.

barrutko (approved):

PR #419, opened 2020-12-17
Fix for rarely failing tests

MiksuJak:


barrutko (approved):
Imo looks good, but it would be nice if @vanruch could also have a look.

vanruch (changes requested):
We are mixing two completely different types of measuring closeliness of numbers. What's worse, in one case it gets more precise with growing eps and in another - with declining. I would propose to split it into two different functions: rename `expectCloseTo` to `expectScaledCloseTo` and make `expectCloseTo` a simple check of `a-b`

MiksuJak:
> We are mixing two completely different types of measuring closeliness of numbers. What's worse, in one case it gets more precise with growing eps and in another - with declining. I would propose to split it into two different functions: rename `expectCloseTo` to `expectScaledCloseTo` and make `expectCloseTo` a simple check of `a-b`

Right, sounds reasonable. Apparently simple check `a-b` does not work for bigger numbers, that's why we need two methods.

vanruch (approved):

PR #424, opened 2020-12-21
Simplify submission restrictions

MiksuJak:


MiksuJak:
Also, how do we handle
`mapping(address => bool) public allowedSubmitters;`
It is not used anymore, but deleting it may interfere with proxy storage.

PR #427, opened 2020-12-23
Stop using infinite curve approval

MiksuJak:


contracts/truefi/TrueFiPool.sol
```diff
@@ -211,9 +211,10 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
 
     /**
      * @dev Function to approve curve gauge to spend y pool tokens
+     * @param amount Amount to approve curveGauge for
      */
-    function approveCurve() external onlyOwner {
-        _curvePool.token().approve(address(_curveGauge), uint256(-1));
+    function approveCurve(uint256 amount) internal {
```
MiksuJak, line 216: Do we even need this method? Tests seem to pass fine without it

barrutko (approved):

PR #451, opened 2021-01-26
Introduce StkTru token

vanruch:


test/governance/StkTruToken.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    stkToken = await deployContract(StkTruTokenFactory)
+    await stkToken.initialize(trustToken.address, tusd.address, oracle.address)
+
+    await trustToken.mint(owner.address, amount)
+    await trustToken.approve(stkToken.address, amount)
+  })
+
+  describe('Oracle', () => {
+    it('only owner can change oracle', async () => {
+      await expect(stkToken.connect(staker).setOracle(staker.address)).to.be.revertedWith('only owner')
```
MiksuJak, line 52:
```suggestion
      await expect(stkToken.connect(staker).setOracle(staker.address)).to.be.revertedWith('Ownable: Caller is not the owner')
```
Don't we want to have full revert messages in tests?

vanruch: These are some prehistoric contracts we don't want to change

test/governance/StkTruToken.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        await trustToken.mint(stkToken.address, parseTRU(10))
+        await tusd.mint(stkToken.address, parseEth(10))
+        await stkToken.stake(amount)
+        await timeTravel(provider, stakeCooldown)
+        await stkToken.unstake(amount)
+        expect(await trustToken.balanceOf(owner.address)).to.equal(amount.add(parseTRU(10)))
+        expect(await tusd.balanceOf(owner.address)).to.equal(parseEth(10))
+      })
+    })
+
```
MiksuJak, line 114: Maybe we should also make a test, that checks if cooldown works properly.
User stakes -> cooldown passes -> stakes again and immediately unstakes.
That should revert.

hal909 (commented):
- make stkTru a Farm
- add distributor for stkTru with 36,000 TRU distributed per day
- use open oracle price feed OR time weighted price average from SushiSwap
- make farm update function claim TFI-LP

contracts/governance/StkTruToken.sol
```diff
@@ -0,0 +1,89 @@
...
+        initalized = true;
+    }
+
+    function setOracle(ITruPriceOracle _oracle) external onlyOwner {
+        oracle = _oracle;
+        emit OracleChanged(_oracle);
+    }
+
+    function stake(uint256 amount) external {
+        unlockTime[msg.sender] = block.timestamp + MIN_STAKE_TIME;
```
hal909, line 45: rather than storing the unlock time, we want to have a function to initiate cooldown. Initiating cooldown will set unlcokTime[msg.sender] to block.timestamp + MIN_STAKE_TIME. Staking more should increase the cooldown. We don't have a definite formula for how to increase the cooldown, so It think it's fine to just reset the cooldown to max if someone stakes more after initiating cooldown.

See stkAave: https://etherscan.io/address/0x4da27a545c0c5b758a6ba100e3a049001de870f5#code

contracts/governance/StkTruToken.sol
```diff
@@ -0,0 +1,89 @@
...
+        require(trustToken.transferFrom(msg.sender, address(this), amount));
+
+        emit Stake(msg.sender, amount, amountToMint);
+    }
+
+    function unstake(uint256 amount) external nonReentrant {
+        require(balanceOf[msg.sender] >= amount, "StkTruToken: Insufficient balance");
+        require(unlockTime[msg.sender] <= block.timestamp, "StkTruToken: Stake is locked");
+
+        uint256 truAmount = trustToken.balanceOf(address(this)).mul(amount).div(totalSupply);
```
hal909, line 59: since we will have TRU rewards and TRU in the staking pool, we need to calculate & transfer those separately here rather than use trustToken.balanceOf.

contracts/governance/StkTruToken.sol
```diff
@@ -0,0 +1,89 @@
...
+import {VoteToken} from "./VoteToken.sol";
+import {ClaimableContract} from "../trusttoken/common/ClaimableContract.sol";
+import {ITruPriceOracle} from "./interface/ITruPriceOracle.sol";
+
+contract StkTruToken is VoteToken, ClaimableContract, ReentrancyGuard {
+    using SafeMath for uint256;
+
+    uint256 public constant MIN_STAKE_TIME = 14 days;
+
+    IERC20 public trustToken;
```
hal909, line 17: can we rename trustToken to just tru? since we are renaming the token to TrueFi.

contracts/governance/StkTruToken.sol
```diff
@@ -0,0 +1,89 @@
...
+    IERC20 public tusd;
```
hal909, line 18: need to use tfusd instead of tusd. tfusd is the new name for TFI-LP

contracts/governance/StkTruToken.sol
```diff
@@ -0,0 +1,89 @@
...
+
+    function setOracle(ITruPriceOracle _oracle) external onlyOwner {
+        oracle = _oracle;
+        emit OracleChanged(_oracle);
+    }
+
+    function stake(uint256 amount) external {
+        unlockTime[msg.sender] = block.timestamp + MIN_STAKE_TIME;
+
+        uint256 amountToMint = totalSupply == 0 ? amount : totalSupply.mul(amount).div(value());
```
hal909, line 47: I'm unsure if this is how we want to calculate amountToMint. Could you explain this line? Why are we using the value of the pool in USD to calculate how much to mint?

contracts/governance/StkTruToken.sol
```diff
@@ -0,0 +1,89 @@
...
+        _burn(msg.sender, amount);
+
+        require(trustToken.transfer(msg.sender, truAmount));
+        require(tusd.transfer(msg.sender, tusdAmount));
+
+        emit Unstake(msg.sender, amount, truAmount, tusdAmount);
+    }
+
+    function value() public view returns (uint256) {
+        return trustToken.balanceOf(address(this)).add(oracle.usdToTru(tusd.balanceOf(address(this))));
```
hal909, line 71: Unsure about this calculation as well. What is the goal here?

contracts/governance/StkTruToken.sol
```diff
@@ -0,0 +1,89 @@
...
+    function rounding() public pure returns (uint8) {
+        return 8;
+    }
+
+    function name() public override pure returns (string memory) {
+        return "Staked TrueFi";
+    }
+
+    function symbol() public override pure returns (string memory) {
+        return "StkTRU";
```
hal909, line 87: lowercase: stkTRU

contracts/governance/VoteToken.sol
```diff
@@ -12,24 +12,30 @@
...
-    function delegateBySig(address delegatee, uint nonce, uint expiry, uint8 v, bytes32 r, bytes32 s) public override {
+    function delegateBySig(
+        address delegatee,
+        uint256 nonce,
+        uint256 expiry,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) public override {
         //OLD: bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, keccak256(bytes(name)), getChainId(), address(this)));
```
hal909, line 37: delete comment

hal909 (commented):
Good start. Please refer to the spec: https://github.com/trusttoken/truefi-spec/blob/master/TrueFi2.0.md

contracts/governance/StkTruToken.sol
vanruch: We want to mint amount of tokens that would represent staker's share in the stake. That is
Before
Stake value (in TRU) = V
Supply = S
Amount (in TRU) = A
Minted = X
After:
Value = V+A
Supply = S+X

X/(S+X)*(V+A)=A
X(V+A)=AS+AX
XV+XA=SA+XA
XV=SA
X = SA/V

contracts/governance/StkTruToken.sol
vanruch: But if TRU rewards are in staking pool as well, there shouldn't be any problem

hal909: the reasoning is, un-staking is different than claiming. Stakers can incentive claim TRU at any time, but need to activate cooldown to un-stake staked TRU

hal909 (approved):

PR #454, opened 2021-01-26
Staking in a farm also calls _claim()

BenjaminCosman:
By far the most (gas-)expensive part of interacting with the farm is the update() modifier, which needs to be called before staking, unstaking, and claiming rewards. Notice that exit() already exists to save on gas by bundling together unstaking and claiming, so that update only needs to be called once. To save on gas, staking and claiming rewards should be similarly bundled.

### DIFF
--- a/contracts/truefi/TrueFarm.sol
+++ b/contracts/truefi/TrueFarm.sol
@@ -92,6 +92,7 @@ contract TrueFarm is ITrueFarm, Initializable {
      * @param amount Amount of tokens to stake
      */
     function stake(uint256 amount) external override update {
+        _claim();
         staked[msg.sender] = staked[msg.sender].add(amount);
         totalStaked = totalStaked.add(amount);
         require(stakingToken.transferFrom(msg.sender, address(this), amount));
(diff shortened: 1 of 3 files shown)


hal909:
it might make sense to check if the account has any claimable rewards first, or check if the account has claimable rewards in the _claim() function. This will optimize staking so that staking with claimable balance will automatically claim. This will increase staking cost for new stakers by calling SLOAD, but still will be beneficial for users staking that have a claimable balance. @vanruch

vanruch:
@BenjaminCosman thanks for this, this could help saving some gas ⛽. And I agree with @hal909 - we need to make claim only if there's anything to be claimed, otherwise we'd perform empty transfer for every new staker.

BenjaminCosman:
Updated accordingly.

vanruch (approved):

barrutko (commented):
Would be nice to have a test that checks if adding more stake to already existing pile of staked tokens triggers the claim. I know it's pretty obvious one, but it's important to cover all wanted and expected behaviour with tests, so we won't mess with it accidentally in the future.

BenjaminCosman:
Test added.

barrutko:
One last thing. All development happens on dev branch. You need to change base to dev and resolve all conflicts that will occur.

PR #455, opened 2021-01-27
Introduce TrueRatingAgencyV2

barrutko:


contracts/governance/interface/IVoteToken.sol
```diff
@@ -1,10 +1,13 @@
...
 interface IVoteToken {
     function delegate(address delegatee) external;
     function delegateBySig(address delegatee, uint nonce, uint expiry, uint8 v, bytes32 r, bytes32 s) external;
     function getCurrentVotes(address account) external view returns (uint96);
     function getPriorVotes(address account, uint blockNumber) external view returns (uint96);
-}
\ No newline at end of file
+}
+
+interface IVoteTokenWithERC20 is IVoteToken, IERC20 {}
```
hal909, line 13: any reason we need this?

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+    // ================ WARNING ==================
+    // ===== THIS CONTRACT IS INITIALIZABLE ======
+    // === STORAGE VARIABLES ARE DECLARED BELOW ==
+    // REMOVAL OR REORDER OF VARIABLES WILL RESULT
+    // ========= IN STORAGE CORRUPTION ===========
+
+    mapping(address => bool) public allowedSubmitters;
+    mapping(address => Loan) public loans;
+
+    IBurnableERC20 public trustToken;
```
hal909, line 70: rename trustToken to tru

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+    IVoteTokenWithERC20 public stakedTrustToken;
```
hal909, line 71: rename stakedTrustToken to stkTRU

hal909, line 71: Also, does it have to be IVoteTokenWithERC20? Why not IERC20? Are we using any of the VoteToken functions in this contract?

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+        require(stakedTrustToken.transfer(msg.sender, amountToTransfer));
+        emit Withdrawn(id, msg.sender, stake, amountToTransfer, burned);
+    }
+
+    /**
+     * @dev Internal view to convert values to 8 decimals precision
+     * @param input Value to convert to TRU precision
+     * @return output TRU amount
+     */
+    function toTrustToken(uint256 input) internal pure returns (uint256 output) {
```
hal909, line 364: toTru

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+
+    /**
+     * @dev Update total TRU reward for a Loan
+     * Reward is divided proportionally based on # TRU staked
+     * chi = (TRU remaining in distributor) / (Total TRU allocated for distribution)
+     * interest = (loan APY * term * principal)
+     * R = Total Reward = (interest * chi * rewardFactor)
+     * @param id Loan ID
+     */
+    modifier calculateTotalReward(address id) {
```
hal909, line 376: we might simplify rewards but that'd TBD. I don't see a huge issue with the way it currently is, except that it's hard to understand haha. We might have a fixed distribution per loan instead, since we will eventually phase out this feature.

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+     * R = Total Reward = (interest * chi)
+     * R is distributed to voters based on their proportion of votes/total_votes
+     *
+     * Claimable reward = R x (current time / total time)
+     *      * (account TRU staked / total TRU staked) - (amount claimed)
+     *
+     * @param id Loan ID
+     * @param voter Voter account
+     */
+    function claim(address id, address voter) public override onlyFundedLoans(id) calculateTotalReward(id) {
```
hal909, line 416: I think we can either remove the claim function, or call it from the withdraw function. Does it make sense to have this?

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+            require(trustToken.transfer(voter, claimableRewards));
+            emit Claimed(id, voter, claimableRewards);
+        }
+    }
+
+    function claimed(address id, address voter) external view returns (uint256) {
+        return loans[id].claimed[voter];
+    }
+
+    function claimable(address id, address voter) public view returns (uint256) {
```
hal909, line 432: We also can simplify the claim function since we will only ever claim once. We can greatly simplify a lot of the claiming logic.

hal909 (commented):
Looks good so far!! I think we need to overhaul the claim logic to make it as simple as possible.

contracts/truefi/TrueRatingAgencyV2.sol
barrutko: Not RN, but we will delegate to and undelegate voting power from the rater.

contracts/truefi/TrueRatingAgencyV2.sol
barrutko: Let's keep it like this in this PR and change it later. Right now it works the same as the old TRA, so the easiest approach is to copy it over. We'll introduce changes in the future.

contracts/truefi/TrueRatingAgencyV2.sol
barrutko: We call it from withdraw function. I kept it public, so it's technically possible to claim without withdrawing, but we con't need to implement any UI endpoint for that. All rewards will be automatically claimed with the withdrawal anyway.

contracts/truefi/TrueRatingAgencyV2.sol
barrutko: We still need claimable function though. Would be nice to display in the UI what will you get after withdrawing.

vanruch (approved):

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+ * incentive TRU throughout the course of the loan. After the loan's term,
+ * if the voter is correct, they earn a TRU reward plus a portion of the
+ * losing side's vote. A portion of the losing side's TRU is burned.
+ *
+ * Voting Lifecycle:
+ * - Borrowers can apply for loans at any time by deploying a LoanToken
+ * - LoanTokens are registered with the prediction market contract
+ * - Once registered, TRU holders can vote at any time
+ * - If a loan is funded, TRU is locked for the term of the loan
+ * - At the end of the term, payouts are determined based on the loan outcome
```
vanruch, line 35: Comment is outdated

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -0,0 +1,485 @@
...
+     * @param stake Amount of TRU to stake
+     */
+    function no(address id, uint256 stake) external override onlyPendingLoans(id) {
+        vote(id, stake, false);
+    }
+
+    // prettier-ignore
+    /**
+     * @dev Withdraw stake on a loan and remove votes.
+     * Unstaking only allowed for loans that are not Running
```
vanruch, line 327: outdated comment

hal909 (approved):

PR #457, opened 2021-01-27
Shortcircuit claimable() calculation when nothing is staked

BenjaminCosman:
Currently if totalStaked is 0 (which will likely never happen in ordinary operation, but is possible), claimable reverts with a division by zero error. This PR fixes that issue, and should have almost no effect on anything else (gas difference should be tiny, and is zero when called from e.g. Etherscan since it's a view-only function).

### DIFF
--- a/contracts/truefi/TrueFarm.sol
+++ b/contracts/truefi/TrueFarm.sol
@@ -150,6 +150,9 @@ contract TrueFarm is ITrueFarm, Initializable {
      * @return claimable rewards for account
      */
     function claimable(address account) external view returns (uint256) {
+        if (staked[account] == 0) {
+            return claimableReward[account];
+        }
         // estimate pending reward from distributor
         uint256 pending = trueDistributor.nextDistribution();
         // calculate total rewards (including pending)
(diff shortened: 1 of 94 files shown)


barrutko:
Maybe then we should check if `totalStaked ` is zero and return zero in such a case? Presented solution only works if we always perform `claim` within withdrawal. Of course currently we do this, but this might change in the future.
Code looks fine in the current shape, but maybe if it's only about solving division-by-0 problem, we should check totalStaked instead.
Also would be nice to have a test for that ✅

BenjaminCosman:
I will add a test.

But I'm confused: why does my solution only work if we perform claim within withdrawal? If `staked[account]` is 0, then the quantity `staked[account].mul(nextcumulativeRewardPerToken.sub(previousCumulatedRewardPerToken[account])).div(PRECISION)` on line 167 is also 0 so the shortcircuiting of lines 156-167 looks valid. Meanwhile, if `staked[account]` is non-0, then totalStaked is also non-0 (bounded below by the stake in any one account), so the lines 156-167 will run as normal with no chance of div-by-0. Basically this PR solves the div-by-0 in the way that _also_ may result in gas savings, in case gas of claimable() ever matters, by shortcircuiting a potentially-common case (individual has no stake) which is a strict superset of the very-rare case (nobody has a stake).

BenjaminCosman:
Tests added.

vanruch (approved):

barrutko:
One last thing. All development happens on dev branch. You need to change base to dev and resolve all conflicts that will occur.

PR #459, opened 2021-01-29
Introduce liquidator

MiksuJak:
Will need a few modifications when stkToken gets introduced


contracts/truefi/Liquidator.sol
```diff
@@ -0,0 +1,105 @@
...
+    using SafeMath for uint256;
+
+    // ================ WARNING ==================
+    // ===== THIS CONTRACT IS INITIALIZABLE ======
+    // === STORAGE VARIABLES ARE DECLARED BELOW ==
+    // REMOVAL OR REORDER OF VARIABLES WILL RESULT
+    // ========= IN STORAGE CORRUPTION ===========
+
+    ITrueFiPool public _pool;
+    IStakingPool public _stakingPool;
```
vanruch, line 28: rename to `stkTru`

contracts/truefi/Liquidator.sol
```diff
@@ -0,0 +1,105 @@
...
+    IERC20 public _trustToken;
```
vanruch, line 29: rename to `tru`

contracts/truefi/Liquidator.sol
```diff
@@ -0,0 +1,105 @@
...
+contract Liquidator is Ownable {
+    using SafeMath for uint256;
+
+    // ================ WARNING ==================
+    // ===== THIS CONTRACT IS INITIALIZABLE ======
+    // === STORAGE VARIABLES ARE DECLARED BELOW ==
+    // REMOVAL OR REORDER OF VARIABLES WILL RESULT
+    // ========= IN STORAGE CORRUPTION ===========
+
+    ITrueFiPool public _pool;
```
vanruch, line 27: rename to pool

contracts/truefi/LoanToken.sol
```diff
@@ -296,6 +321,15 @@ contract LoanToken is ILoanToken, ERC20 {
         emit Closed(status, _balance());
     }
 
+    /**
+     * @dev Liquidate the loan if it has defaulted
+     */
+    function liquidate() external override onlyDefaulted onlyLiquidator {
+        status = Status.Liquidated;
```
vanruch, line 328: Should liquidate also do reclaim?

contracts/truefi/LoanToken.sol
```diff
@@ -34,6 +34,7 @@ contract LoanToken is ILoanToken, ERC20 {
     uint8 public constant override version = 2;
```
vanruch, line 34: Bump version

contracts/truefi/LoanToken.sol
MiksuJak: So i designed it the way, that if borrower payed some of his debt back, then it would be claimable with loan tokens. The difference will be taken from stkTru (or 10% bound). This way when stkTru balance is low, pool participants will suffer smaller losses.

contracts/truefi/Liquidator.sol
```diff
@@ -0,0 +1,120 @@
...
+        emit OracleChanged(newOracle);
+    }
+
+    /**
+     * @dev Liquidates a defaulted Loan, withdraws a portion of tru from staking pool
+     * then transfers tru to TrueFiPool as compensation
+     * @param loan Loan to be liquidated
+     */
+    function liquidate(ILoanToken loan) external {
+        uint256 defaultedValue = getAmountToWithdraw(loan.debt().sub(loan.balance()));
```
barrutko, line 103: What if loan is closed as defaulted (it was repaid partially), some loan tokens are redeemed and then loan is liquidated? Redeeming some part of loan will make debt look bigger here. We probably should take this scenario into account.

MiksuJak: Very nice catch! ⚾️
Luckily one function change fixes the problem.
```
        uint256 defaultedValue = getAmountToWithdraw(loan.debt().sub(loan.repaid()));
```

hal909 (approved):

PR #464, opened 2021-02-04
Add TRU to LoanToken for easier future modifications

MiksuJak:


contracts/truefi/LoanToken.sol
```diff
@@ -303,11 +307,13 @@ contract LoanToken is ILoanToken, ERC20 {
      */
     function redeem(uint256 _amount) external override onlyClosed {
         uint256 amountToReturn = _amount.mul(_balance()).div(totalSupply());
+        uint256 truToReturn = _amount.mul(tru.balanceOf(address(this))).div(totalSupply());
         redeemed = redeemed.add(amountToReturn);
         _burn(msg.sender, _amount);
         require(currencyToken.transfer(msg.sender, amountToReturn));
+        require(tru.transfer(msg.sender, truToReturn));
```
barrutko, line 314: Maybe some small optimisation like `if(truToReturn)`? This is gonna be very rare case anyway.

contracts/truefi/LoanFactory.sol
```diff
@@ -46,6 +48,10 @@ contract LoanFactory is ILoanFactory, Initializable {
         lender = 0x16d02Dc67EB237C387023339356b25d1D54b0922;
     }
 
+    function setTru() external {
+        tru = 0x4C19596f5aAfF459fA38B0f7eD92F11AE6543784;
```
barrutko, line 52: If we are harcoding it anyway, and change will require update anyway, maybe let's not use setter, but just implement getter returning `tru` address

hal909 (approved):

hal909 (approved):

PR #469, opened 2021-02-05
Remove onlyOwner from reclaim

vanruch:


BenjaminCosman:
This doesn't look safe: I think you need a more robust check that the passed in `loanToken` is _actually_ a known loan (e.g. checking if it is in `_loans`) instead of just asking it (`loanToken.isLoanToken()`). Otherwise, I haven't figured out how exactly I'd attack this yet but it seems like I could maybe do something evil (reentrancy?) in between the computations of `balanceBefore` and `balanceAfter`.

hal909 (approved):

PR #471, opened 2021-02-07
Check downcast

BenjaminCosman:
...or if it was purposefully unchecked to save on gas, then instead you should add a comment (e.g. "This cast is unchecked to save on gas; it is safe because the total supply of each of our tokens can fit in a uint96. If future governance ever greatly increases the total supply, it should make sure to also change this cast to safe96.")

### DIFF
--- a/contracts/governance/VoteToken.sol
+++ b/contracts/governance/VoteToken.sol
@@ -104,7 +104,7 @@ abstract contract VoteToken is ERC20, IVoteToken {
         uint256 _value
     ) internal virtual override {
         super._transfer(_from, _to, _value);
-        _moveDelegates(delegates[_from], delegates[_to], uint96(_value));
+        _moveDelegates(delegates[_from], delegates[_to], safe96(_value, "StkTruToken: uint96 overflow"));
     }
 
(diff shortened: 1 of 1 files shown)


contracts/governance/VoteToken.sol
```diff
@@ -104,7 +104,7 @@ abstract contract VoteToken is ERC20, IVoteToken {
         uint256 _value
     ) internal virtual override {
         super._transfer(_from, _to, _value);
-        _moveDelegates(delegates[_from], delegates[_to], uint96(_value));
+        _moveDelegates(delegates[_from], delegates[_to], safe96(_value, "StkTruToken: uint96 overflow"));
```
hal909, line 107: can we write a test for this?

BenjaminCosman: I can do that. Note that none of the _other_ safe96 casts have corresponding tests - shall I add tests for all of them or just this new one?

barrutko: I think it's fine to leave it like this.

barrutko (approved):

PR #498, opened 2021-02-18
Introduce fee payers whitelist

barrutko:


vanruch (approved):

contracts/governance/StkTruToken.sol
```diff
@@ -74,6 +76,7 @@ contract StkTruToken is VoteToken, ClaimableContract, ReentrancyGuard {
     event Cooldown(address indexed who, uint256 endTime);
     event CooldownTimeChanged(uint256 newUnstakePeriodDuration);
     event UnstakePeriodDurationChanged(uint256 newUnstakePeriodDuration);
+    event FeePayerWhitelistingStatusChanged(address payer, bool status);
```
BenjaminCosman, line 79: Should payer be an `indexed` argument? Not sure it really matters, but it looks like you've been doing that almost everywhere else with addresses.

BenjaminCosman (approved):

contracts/governance/StkTruToken.sol
hal909: this doesn't get called a lot & I think you can only index a certain number of params per contract

PR #514, opened 2021-02-24
fix TRU storage

vanruch:


contracts/governance/common/ClaimableContract.sol
```diff
@@ -1,16 +1,18 @@
...
  * @title ClaimableContract
  * @dev The ClaimableContract contract is a copy of Claimable Contract by Zeppelin.
  and provides basic authorization control functions. Inherits storage layout of
  ProxyStorage.
  */
-contract ClaimableContract {
-    address owner_;
-    address pendingOwner_;
-    bool initalized;
+contract ClaimableContract is ProxyStorage {
```
hal909, line 12: I don't quite understand this change

hal909 (commented):
@vanruch Can you explain what you're doing here? I don't quite understand why ClaimableContract is ProxyStorage (but it's in the governance folder?) I think we need to talk tomorrow to refactor the folder & naming structure so we are consistent from now on.

vanruch:
> @vanruch Can you explain what you're doing here? I don't quite understand why ClaimableContract is ProxyStorage (but it's in the governance folder?) I think we need to talk tomorrow to refactor the folder & naming structure so we are consistent from now on.

Yeah, we now have a bunch of different ProxyStorages and Claimable contracts all with slightly different storage allocations and I'm not sure if we can do anything nice with it now :(

PR #520, opened 2021-03-04
Add TRU and CRV prices to poolValue

vanruch:


.waffle.json
```diff
@@ -3,7 +3,7 @@
     "compilerOptions": {
         "optimizer": {
           "enabled": true,
-          "runs": 20000
+          "runs": 200
```
hal909, line 6: can we keep this at 20,000? Since we removed some function from the pool

vanruch, line 6:
```suggestion
          "runs": 20000
```

contracts/truefi/TrueFiPool.sol
```diff
@@ -181,6 +193,16 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
         _;
     }
 
+    /**
+     * @dev only lender can perform borrowing or repaying
```
hal909, line 197: wrong comment for modifier

contracts/truefi/TrueFiPool.sol
```diff
@@ -223,12 +245,21 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
...
-    function setOracle(ITruPriceOracle newOracle) public onlyOwner {
-        _oracle = newOracle;
-        emit OracleChanged(newOracle);
+    function setTruOracle(ITruPriceOracle newOracle) public onlyOwner {
+        _truOracle = newOracle;
+        emit TruOracleChanged(newOracle);
+    }
+
+    /**
+     * @dev set TrueFi price oracle token address
```
hal909, line 257: fix comment

contracts/truefi/TrueFiPool.sol
```diff
@@ -256,6 +287,14 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
         return _stakeToken.balanceOf(address(this));
     }
 
+    /**
+     * @dev Get total balance of stake tokens
```
hal909, line 291: change this comment to be more specific to curve.fi

contracts/truefi/TrueFiPool.sol
```diff
@@ -256,6 +287,14 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
...
+     * @return Balance of stake tokens in this contract
+     */
+    function crvBalance() public view returns (uint256) {
+        return _minter.token().balanceOf(address(this));
```
hal909, line 295: does this include unclaimed crv balance? should we be including that when calculating the total pool value?

contracts/truefi/TrueFiPool.sol
```diff
@@ -677,4 +735,8 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
 
         return mintedAmount;
     }
+
+    function conservativePriceEstimation(uint256 price) internal pure returns (uint256) {
```
hal909, line 739: add documentation

contracts/truefi/TrueFiPool.sol
vanruch: No it doesn't and I think that including it would be wrong. These tokens are not in the pool yet and it may lead to some unexpected problems if we include it into the value calculation.

hal909 (approved):

PR #523, opened 2021-03-08
Add TrueFarm.claimRestake() to skip transfers

yuchennil:
Avoid wasting gas + user effort on transferring to/from msg.sender when a user wants to claim and immediately restake all claimable reward.

I should add some tests, though I haven't looked into the testing framework yet. Just want to make sure this is on the right track.


vanruch:
Solidity code looks great, but lacks tests. Also, could you please try to close and reopen the PR to try to trigger the CI?

PR #524, opened 2021-03-09
Remove unnecessary functions from TrueFiPool

vanruch:


contracts/truefi/TrueFiPool.sol
```diff
@@ -163,38 +157,6 @@ contract TrueFiPool is ITrueFiPool, ERC20, ReentrancyGuard, Ownable {
...
 
-    /**
-     * @dev Initialize pool
-     * @param __curvePool curve pool address
-     * @param __curveGauge curve gauge address
-     * @param __currencyToken curve pool underlying token
-     * @param __lender TrueLender address
-     * @param __uniRouter Uniswap router
-     */
-    function initialize(
```
barrutko, line 174: Should we really remove the initializer?

vanruch: Why not?

hal909 (approved):

PR #527, opened 2021-03-10
Add git hook to catch private keys

vanruch:
You'll have to call
`git config core.hooksPath .githooks`
to make this work


MiksuJak (approved):
Works well 🦖

barrutko (approved):

PR #528, opened 2021-03-10
Add StkTruToken.claimRestake() to save gas + time

yuchennil:


contracts/governance/StkTruToken.sol
```diff
@@ -401,15 +397,46 @@ contract StkTruToken is VoteToken, StkClaimableContract, ReentrancyGuard {
...
         require(token.transfer(msg.sender, rewardToClaim));
         emit Claim(msg.sender, token, rewardToClaim);
     }
 
+    /**
+     * @dev Internal claim function that returns the transfer value
+     * Claim rewards for a specific ERC20 token to return in a uint256
+     * @param token Token to claim rewards for
+     */
+    function _claim_to_uint256(IERC20 token) internal returns (uint256) {
```
vanruch, line 410: I'm not a fan of having types in function name. We already have it in the function signature. @hal909 @barrutko what do you think?

yuchennil: Thanks for the feedback!

Yeah, I was also uncomfortable with the function name for the same reason and spent some time trying to come up with a better one (e.g., _claim_to_tru, _claim_to_token, _partial_claim, _incomplete_claim, _claim_without_transfer). Same with _stake_from_uint256.

I think what we want to ultimately express is that this internal method does a partial update of the contract state but isn't done yet. The change still needs to be carried forward, either by issuing a transaction out or by updating another part of internal state. Unfortunately at its most general everything in the EVM appears to be a uint256, so I agree the type name doesn't say much.

I couldn't come up with anything more concise and legible, so I just sent this PR out to see what the rest of the team thinks.

What do you think of my second-best candidate: _claim_without_transfer() and _stake_without_transfer()? I generally consider it poor style to describe what a function *doesn't* do, but I'm starting to think in this case it might be the most legible description.

vanruch: _claimWithoutTransfer() looks fine, but I'd say it is better to rename current claim to something like `_claimAndTransfer()` and simply call new one `_claim()`. Similarly, `_stake()` seems like an ok name for internal helper for stake method. 

And one more thing, according to style guide https://docs.soliditylang.org/en/latest/style-guide.html function names should be in mixedCase, and not underscored_case.

Generally, the PR looks great!

vanruch: Oh, and when you'll be pushing changes, sync the branch with the current dev, otherwise integration tests with 1Inch will fail :)

yuchennil: Thanks!

I think I'll go with _claimWithoutTransfer() for now. If we want claim() to keep a consistent meaning across all smart contracts, then I suspect it would be confusing to have it only zero out claimableReward for this specific contract, but elsewhere mean both zeroing out and transferring. Maybe the overall claim() naming can be revisited at a later date, but I suspect it isn't a priority.

Thanks for the style catch! I should have noticed that myself.

Yep, I'll sync again before pushing.

vanruch (approved):

test/governance/StkTruToken.test.ts
```diff
@@ -268,6 +268,40 @@ describe('StkTruToken', () => {
...
+      it('clears claimable balance', async () => {
+        expectScaledCloseTo(await stkToken.claimable(owner.address, tru.address), parseTRU(1000))
+        await stkToken.claimRestakeTRU()
+        expect(await stkToken.claimable(owner.address, tru.address)).to.equal(0)
+      })
+
+      it('stakes claimable balance', async () => {
+        expect(await stkToken.stakeSupply()).to.equal(amount)
+        await stkToken.claimRestakeTRU()
+        expectScaledCloseTo(await stkToken.stakeSupply(), amount.add(parseTRU(1000)))
```
MiksuJak, line 296: Don't actually know if it is necessary, but adding an extra test never did any harm. We could maybe do the same thing with timeTravel and a little increased value? (proportionally to the distributor's distribution of course)

Anyway, the code looks good 🥳

MiksuJak (approved):

contracts/governance/StkTruToken.sol
```diff
@@ -401,15 +397,46 @@ contract StkTruToken is VoteToken, StkClaimableContract, ReentrancyGuard {
...
+            emit Cooldown(msg.sender, block.timestamp.add(cooldownTime));
+        }
+
+        if (delegates[msg.sender] == address(0)) {
+            delegates[msg.sender] = msg.sender;
+        }
+
+        uint256 amountToMint = stakeSupply == 0 ? amount : amount.mul(totalSupply).div(stakeSupply);
+        _mint(msg.sender, amountToMint);
+        stakeSupply = stakeSupply.add(amount);
```
hal909, line 437: it might be better to emit the Stake event here and remove it from the public functions. Same with claimWithoutTransfer. Then you know the event will always be emitted.

contracts/governance/StkTruToken.sol
```diff
@@ -319,6 +304,17 @@ contract StkTruToken is VoteToken, StkClaimableContract, ReentrancyGuard {
...
     }
 
+    /**
+     * @dev Claim TRU rewards, then restake without transferring
+     * Allows account to save more gas by avoiding out-and-back transfers
+     */
+    function claimRestakeTRU() external distribute update(msg.sender) {
+        uint256 amount = _claimWithoutTransfer(tru);
+        _stakeWithoutTransfer(amount);
+        emit Claim(msg.sender, tru, amount);
```
hal909, line 314: remove events here and put them in the internal functions

contracts/governance/StkTruToken.sol
```diff
@@ -401,15 +397,46 @@ contract StkTruToken is VoteToken, StkClaimableContract, ReentrancyGuard {
...
      */
     function _claim(IERC20 token) internal {
-        farmRewards[token].totalClaimedRewards = farmRewards[token].totalClaimedRewards.add(
-            farmRewards[token].claimableReward[msg.sender]
-        );
-        uint256 rewardToClaim = farmRewards[token].claimableReward[msg.sender];
-        farmRewards[token].claimableReward[msg.sender] = 0;
+        uint256 rewardToClaim = _claimWithoutTransfer(token);
         require(token.transfer(msg.sender, rewardToClaim));
         emit Claim(msg.sender, token, rewardToClaim);
```
hal909, line 402: Remove this Claim event

contracts/governance/StkTruToken.sol
```diff
@@ -319,6 +304,17 @@ contract StkTruToken is VoteToken, StkClaimableContract, ReentrancyGuard {
         _claim(token);
     }
 
+    /**
+     * @dev Claim TRU rewards, then restake without transferring
+     * Allows account to save more gas by avoiding out-and-back transfers
+     */
+    function claimRestakeTRU() external distribute update(msg.sender) {
```
hal909, line 311: you can rename this to just claimRestake()

hal909 (changes requested):
Looks good so far! You can move the Claim() and Stake() events to be emitted in the internal functions and remove the emits from the public ones.

test/governance/StkTruToken.test.ts
yuchennil: Thanks!

With this test I'm just making sure the basic functionality of this function gets covered, so a refactor later on doesn't break this without warning us. timeTravel might also be nice, but as long as it's tested more carefully elsewhere, I don't think there's much need to repeat it here.

I'm not too attached to this opinion though, so if you'd like I can add a test case.

yuchennil:
> Looks good so far! You can move the Claim() and Stake() events to be emitted in the internal functions and remove the emits from the public ones.

Thanks for the review! I moved the Claim() and Stake() emits into the internal functions.

I had considered moving these emits inside in my original commit, but at the time I thought it mattered that event emission should occur after transfers, just in case a transfer fails. But I should've realized atomic transactions also means emits inside the transaction succeed iff the transaction overall succeeds.

Under the Checks-Effects-Interactions pattern (https://docs.soliditylang.org/en/v0.5.10/security-considerations.html#use-the-checks-effects-interactions-pattern), should we generally prefer to emit events *before* we transfer? If we count events as effects, not interactions, then it would seem they should come first. My guess is it's a matter of code style that doesn't make a difference, since events don't have on-chain side effects and there aren't any reentrancy concerns.

yuchennil:
I haven't been able to figure out why ci/circleci: upgrade has been failing for this change.

Could there have been a regression, @MiksuJak ? When I run "yarn test:upgrade" locally, I get the following error, which seems to point to #526:

  4 passing (1m)
  1 failing

  1) Pool 1Inch integration
       sells all available CRV:
     RuntimeError: VM Exception while processing transaction: revert TrueFiPool: Unknown caller

MiksuJak:
> I haven't been able to figure out why ci/circleci: upgrade has been failing for this change.
> 
> Could there have been a regression, @MiksuJak ? When I run "yarn test:upgrade" locally, I get the following error, which seems to point to #526:
> 
> 4 passing (1m)
> 1 failing
> 
> 1. Pool 1Inch integration
>    sells all available CRV:
>    RuntimeError: VM Exception while processing transaction: revert TrueFiPool: Unknown caller

You should rebase this branch to fresh dev. @vanruch has resolved all known issues 😎

yuchennil:
> > I haven't been able to figure out why ci/circleci: upgrade has been failing for this change.
> > Could there have been a regression, @MiksuJak ? When I run "yarn test:upgrade" locally, I get the following error, which seems to point to #526:
> > 4 passing (1m)
> > 1 failing
> > 
> > 1. Pool 1Inch integration
> >    sells all available CRV:
> >    RuntimeError: VM Exception while processing transaction: revert TrueFiPool: Unknown caller
> 
> You should rebase this branch to current dev. @vanruch has resolved all known issues

Thanks for taking care of the issues! I've merged dev, and all checks are passing now. @hal909, can you review again?

hal909 (approved):

PR #538, opened 2021-03-22
timelock pause

hal909:


contracts/governance/Timelock.sol
```diff
@@ -58,13 +57,18 @@ contract Timelock is ClaimableContract {
 
         owner_ = msg.sender;
         initalized = true;
-        // OLD: N/A
-        admin_initialized = false;
+        admin_initialized = true;
```
vanruch, line 60: Was setting this to true intentional? Looks like this defeats the purpose of having this property (and also breaks a bunch of tests)

vanruch (approved):

PR #547, opened 2021-03-29
Add an isRepaidInFull() view to LoanToken

yuchenlintt:
We can deduplicate some logic in close(), so I've split this into external and internal versions.

I'm wondering whether it makes more sense to create an outstandingDebt() view that just returns debt.sub(_balance()), instead of taking this boolean approach. An outstandingDebt() view would also be useful in repay()'s overpayment check and #546's repayInFull()'s amount calculation. Is the concern that a UI might round an outstandingDebt() of 0.00000001 TUSD to 0.00 TUSD?

### DIFF
--- a/contracts/truefi/LoanToken.sol
+++ b/contracts/truefi/LoanToken.sol
@@ -308,7 +308,7 @@ contract LoanToken is ILoanToken, ERC20 {
      * @dev Close the loan and check if it has been repaid
      */
     function close() external override onlyOngoing {
-        if (_balance() >= debt) {
+        if (_isRepaidInFull()) {
             status = Status.Settled;
         } else {
             require(start.add(term).add(lastMinutePaybackDuration) <= block.timestamp, "LoanToken: Loan cannot be closed yet");
@@ -374,6 +374,14 @@ contract LoanToken is ILoanToken, ERC20 {
      */
     function repaid() external override view onlyAfterWithdraw returns (uint256) {
         return _balance().add(redeemed);
+    }
+
+    function isRepaidInFull() external override view returns (bool) {
+        return _isRepaidInFull();
+    }
+
+    function _isRepaidInFull() internal view onlyOngoing returns (bool) {
+        return _balance() >= debt;
     }
 
     /**
--- a/contracts/truefi/interface/ILoanToken.sol
+++ b/contracts/truefi/interface/ILoanToken.sol
@@ -57,6 +57,8 @@ interface ILoanToken is IERC20 {
 
     function repaid() external view returns (uint256);
 
+    function isRepaidInFull() external view returns (bool);
+
     function balance() external view returns (uint256);
 
     function value(uint256 _balance) external view returns (uint256);


MiksuJak (approved):

contracts/truefi/LoanToken.sol
```diff
@@ -376,6 +376,14 @@ contract LoanToken is ILoanToken, ERC20 {
         return _balance().add(redeemed);
     }
 
+    function isRepaidInFull() external override view returns (bool) {
```
hal909, line 379: rename isRepaid()

contracts/truefi/LoanToken.sol
```diff
@@ -376,6 +376,14 @@ contract LoanToken is ILoanToken, ERC20 {
...
+        return _isRepaidInFull();
+    }
+
+    function _isRepaidInFull() internal view onlyOngoing returns (bool) {
```
hal909, line 383: short, simple functions, we don't need an internal function. you can remove this and move the logic to the public one

contracts/truefi/LoanToken.sol
```diff
@@ -376,6 +376,14 @@ contract LoanToken is ILoanToken, ERC20 {
         return _balance().add(redeemed);
     }
```
hal909, line 378: add documentation

hal909 (changes requested):
add a test and implement the comments, then this is good

contracts/truefi/LoanToken.sol
yuchenlintt: Sounds good!

contracts/truefi/LoanToken.sol
yuchenlintt: The other one is currently external (like all other functions in the ILoanToken interface), but I'll combine them and make it public.

What are your thoughts on the internal _balance() vs external balance() that just delegates? These are defined immediately below isRepaid().

yuchenlintt: I just tried combining the internal/external isRepaid() functions into a single public function and got this compile error in ILoanToken:

    TypeError: Functions in interfaces must be declared external.

If I compile with a single external definition then (as expected) I get this error:

    DeclarationError: Undeclared identifier. Did you mean "Repaid" or "isRepaid"?
            if (isRepaid()) {

For now I'll keep the internal/external split, unless you know another way to resolve this?

contracts/truefi/LoanToken.sol
yuchenlintt: Done!

yuchenlintt:
Hey @hal909, I just finished addressing your comments, minus the internal/external -> public merge (see above for details). Could you take another look?

Also, have you heard back from product on whether they'd prefer isRepaid() as a bool or would be okay with outstandingDebt() as a uint256?

contracts/truefi/LoanToken.sol
```diff
@@ -394,6 +394,22 @@ contract LoanToken is ILoanToken, ERC20 {
...
 
+    /**
+     * @dev Check whether an ongoing loan has been repaid in full
+     * @return true if and only if this loan has been repaid
+     */
+    function isRepaid() external override view returns (bool) {
+        return _isRepaid();
+    }
```
vanruch, line 403: Do we need both `isRepaid` and `_ isRepaid`? I'd say that making `isRepaid` public and using it internally should be good enough

yuchenlintt: (See comment thread above)

I'd like to get rid of _isRepaid as well, but unfortunately I can't find a way to make it compile.
- If I make isRepaid public then the compiler complains that ILoanToken functions must be external, not public
- If I make isRepaid external then the compiler complains that we can't call it internally from close()

So it looks to me like we have to keep the internal/external split to satisfy the compiler. This also appears to be the workaround we're using for balance()/_balance() (lines 417/425) right now.

vanruch: You can set visibility to external in the interface and public in the implementation.

yuchenlintt: Oh that worked!

Cool, thanks for the Solidity tip. Looks like interfaces here work differently from what I'd expected.

contracts/truefi/LoanToken.sol
yuchenlintt: Ivan helped me figure it out. I hadn't realized the interface could have an external function that you can override with a public implementation.

hal909 (approved):

PR #548, opened 2021-03-29
TrueLenderV2

vanruch:


contracts/truefi2/TrueLender2.sol
```diff
@@ -0,0 +1,219 @@
...
+contract TrueLender2 is ITrueLender2, Ownable {
+    using SafeMath for uint256;
+
+    // ================ WARNING ==================
+    // ===== THIS CONTRACT IS INITIALIZABLE ======
+    // === STORAGE VARIABLES ARE DECLARED BELOW ==
+    // REMOVAL OR REORDER OF VARIABLES WILL RESULT
+    // ========= IN STORAGE CORRUPTION ===========
+
+    mapping(ITrueFiPool2 => ILoanToken2[]) loansOnPool;
```
MiksuJak, line 29: Maybe just `poolLoans`?

contracts/truefi/LoanToken.sol
```diff
@@ -449,4 +448,8 @@ contract LoanToken is ILoanToken, ERC20 {
     ) internal override onlyWhoCanTransfer(sender) {
         return super._transfer(sender, recipient, _amount);
     }
+
+    function version() external virtual override pure returns (uint8) {
```
hal909, line 452: documentation

contracts/truefi2/TrueLender2.sol
hal909: agree, poolLoans is good

contracts/truefi2/TrueLender2.sol
```diff
@@ -0,0 +1,219 @@
...
+import {ITrueFiPool2} from "./interface/ITrueFiPool2.sol";
+import {IPoolFactory} from "./interface/IPoolFactory.sol";
+
+/**
+ * @title TrueLender v2.0
+ * @dev Loans management helper
+ * This contract is a bridge that helps to transfer funds from pool to the loans and back
+ * TrueLender holds all LoanTokens and may distribute them on pool exits
+ */
+contract TrueLender2 is ITrueLender2, Ownable {
```
hal909, line 20: make this Claimable with the new claimable contract

contracts/truefi2/interface/ILoanToken2.sol
```diff
@@ -0,0 +1,14 @@
...
+
+import {ITrueFiPool2} from "./ITrueFiPool2.sol";
+import {ILoanToken} from "../../truefi/interface/ILoanToken.sol";
+
+interface IContractWithPool {
+    function pool() external view returns (ITrueFiPool2);
+}
+
+// Had to be split because of multiple inheritance problem
+interface ILoanToken2 is ILoanToken, IContractWithPool {
```
hal909, line 12: Should we just create a new LoanToken interface? It might be harder to read / understand the inheritance, and we are Deprecating the old loan token anyways.

vanruch: There is still active development of current LoanToken and it's not going to change much. So let's keep it like that to avoid manual merging of changes in old LoanToken into the new one. And later we can make it a single contract.

hal909 (approved):

PR #550, opened 2021-03-30
Split LoanToken.close() into settle() and enterDefault()

yuchenlintt:
This should reduce our hesitation to call settle() (née close()), since there won't be a possibility we accidentally default a loan that has passed its term. As a side benefit I think our tests are a little more legible as well, since after this we can see whether a loan was supposed to settle or default.

A note on naming: default is apparently a reserved keyword in Solidity (though as far as I know it hasn't been assigned a meaning yet), so I chose enterDefault() instead. Feel free to suggest an alternative.

### DIFF
--- a/contracts/truefi/LoanToken.sol
+++ b/contracts/truefi/LoanToken.sol
@@ -163,7 +163,7 @@ contract LoanToken is ILoanToken, ERC20 {
     /**
      * @dev Only when loan is Settled
      */
-    modifier onlyClosed() {
+    modifier onlySettled() {
         require(status >= Status.Settled, "LoanToken: Current status should be Settled or Defaulted");
         _;
     }
@@ -339,10 +339,10 @@ contract LoanToken is ILoanToken, ERC20 {
 
     /**
      * @dev Redeem LoanToken balances for underlying currencyToken
-     * Can only call this function after the loan is Closed
+     * Can only call this function after the loan is Settled
      * @param _amount amount to redeem
      */
-    function redeem(uint256 _amount) external override onlyClosed {
+    function redeem(uint256 _amount) external override onlySettled {
         uint256 amountToReturn = _amount.mul(_balance()).div(totalSupply());
         redeemed = redeemed.add(amountToReturn);
         _burn(msg.sender, _amount);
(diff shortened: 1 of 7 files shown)


contracts/truefi/LoanToken.sol
```diff
@@ -158,7 +163,7 @@ contract LoanToken is ILoanToken, ERC20 {
     /**
      * @dev Only when loan is Settled
      */
-    modifier onlyClosed() {
+    modifier onlySettled() {
```
MiksuJak, line 166: I think modifier's name should remain `onlyClosed()`
`onlySettled()` is not accurate, because `status >= Status.Settled` means settled/defaulted/liquidated. 
In all of those cases redeem should be an option, in case when borrower repaid his debt partially
Also forbidding reclaim when a loan defaulted, would just freeze funds forever and nobody would profit from that

yuchenlintt: My bad, thanks for catching my mistake. I'll revert this change then. I have to read more carefully!

yuchenlintt: I've reverted the name to onlyClosed() and updated the documentation to mention Defaulted + Liquidated as well.

MiksuJak:
We could also update the revert message for `onlyClosed()`. I forgot to do it while implementing Liquidator contract, which is totally my bad. However adding 'or liquidated' to revert message will make it longer and less clear for a quick glance. Any ideas?

yuchenlintt:
> We could also update the revert message for `onlyClosed()`. I forgot to do it while implementing Liquidator contract, which is totally my bad. However adding 'or liquidated' to revert message will make it longer and less clear for a quick glance. Any ideas?

How about this?

    /**
     * @dev Only after loan has been closed: Settled, Defaulted, or Liquidated
     */
    modifier onlyAfterClosed() {
        require(status >= Status.Settled, "LoanToken: Only after loan has been closed");
        _;
    }

Then it would have consistent phrasing with modifier onlyAfterWithdraw() in line 190.

vanruch:
> How about this?
Maybe `onlyAfterClose` is more grammatically correct? Anyway, this error message

contracts/truefi/LoanToken.sol
```diff
@@ -76,11 +76,16 @@ contract LoanToken is ILoanToken, ERC20 {
...
+     * @dev Emitted when loan has been fully repaid
+     * @param returnedAmount Amount that was returned
+     */
+    event Settled(uint256 returnedAmount);
+
+    /**
+     * @dev Emitted when term is over without full repayment
      * @param returnedAmount Amount that was returned before expiry
      */
-    event Closed(Status status, uint256 returnedAmount);
```
hal909, line 83: @vanruch @barrutko do we want to actually change this event?

hal909 (changes requested):
just want @vanruch and @barrutko to approve this then we can go ahead and merge

yuchenlintt:
Thanks for the review!

As per @vanruch + @MiksuJak's suggestions above, I renamed onlyClosed to onlyAfterClose and modified the revert message.

hal909 (approved):

PR #551, opened 2021-03-30
Settle LoanTokens when fully repaid via function call

yuchenlintt:
We discussed this in standup, but I don't recall if we decided to move forward with it.

Pros:
- atomically repay in full and then settle the loan in a single transaction
- reduce the lag of closing loans, whether it's manually initiated or scheduled in a keeper

Cons:
- costs borrowers slightly more gas

### DIFF
--- a/contracts/truefi/LoanToken.sol
+++ b/contracts/truefi/LoanToken.sol
@@ -347,10 +347,28 @@ contract LoanToken is ILoanToken, ERC20 {
      * @param _sender account sending currencyToken to repay
      * @param _amount amount of currencyToken to repay
      */
-    function repay(address _sender, uint256 _amount) external override onlyAfterWithdraw {
+    function repay(address _sender, uint256 _amount) external override {
+        _repay(_sender, _amount);
+    }
+
+    /**
+     * @dev Function for borrower to repay all of the remaining loan balance
+     * Borrower should use this to ensure full repayment
+     * @param _sender account sending currencyToken to repay
+     */
+    function repayInFull(address _sender) external override {
+        _repay(_sender, debt.sub(_balance()));
+    }
+
+    /**
+     * @dev Internal function for loan repayment
+     * @param _sender account sending currencyToken to repay
+     * @param _amount amount of currencyToken to repay
+     */
+    function _repay(address _sender, uint256 _amount) internal onlyAfterWithdraw {
         require(_amount <= debt.sub(_balance()), "LoanToken: Cannot repay over the debt");
-        require(currencyToken.transferFrom(_sender, address(this), _amount));
         emit Repaid(_sender, _amount);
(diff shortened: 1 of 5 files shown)


test/truefi/LoanToken.test.ts
```diff
@@ -371,6 +371,17 @@ describe('LoanToken', () => {
...
+
+    it('emits Closed event', async () => {
+      await loanToken.fund()
+      await withdraw(borrower)
+      const debt = await loanToken.debt()
+      await tusd.connect(borrower).approve(loanToken.address, debt)
+      await tusd.mint(borrower.address, parseEth(300))
+      await expect(loanToken.repayInFull(borrower.address))
+        .to.emit(loanToken, 'Closed')
+        .withArgs(borrower.address, debt)
```
MiksuJak, line 383:
```suggestion
        .withArgs(LoanTokenStatus.Settled, debt)
```

vanruch:
Looks good, let's wait for #550 to be merged, make `settle()` public instead of external and call it on repay

hal909 (approved):

hal909 (changes requested):
fix merge conflicts, then this is good

contracts/truefi/LoanToken.sol
```diff
@@ -373,13 +373,19 @@ contract LoanToken is ILoanToken, ERC20 {
...
      * @param _amount amount of currencyToken to repay
      */
     function _repay(address _sender, uint256 _amount) internal onlyAfterWithdraw {
-        require(_amount <= debt.sub(_balance()), "LoanToken: Cannot repay over the debt");
+        uint256 outstandingDebt = debt.sub(_balance());
+        require(_amount <= outstandingDebt, "LoanToken: Cannot repay over the debt");
         emit Repaid(_sender, _amount);
+
         currencyToken.safeTransferFrom(_sender, address(this), _amount);
+        if (_amount >= outstandingDebt) {
```
hal909, line 386: instead of checking amount, check isRepaid(). This will settle the loan if the total debt is repaid, not just the amount paid in this transaction

yuchenlintt: Thanks for the catch! You're right, I should have gone for clarity over saving an extra isRepaid() function call.

I'll also note that calling settle() after the currency transfer violates the checks-effects-interactions pattern, but I think it's justified here because:
1) the only state change after the transfer is status = Settled, which doesn't care about reentrancy. (Currently we call settle() a long time after repaying anyway.)
2) I think it's dangerous to remove the balance >= debt check from settle(), which we'd need to do if we shifted the call to settle() up before the transfer.
3) If we don't remove the check, then we'd have to duplicate settle() code in _repay(). But if we change settle() in the future, we might overlook changing it here too.

hal909 (approved):

PR #553, opened 2021-03-30
Make truefi2 contracts claimable instead of ownable

MiksuJak:


contracts/truefi2/TrueFiPool2.sol
```diff
@@ -65,7 +65,7 @@ contract TrueFiPool2 is ITrueFiPool2, ERC20, Ownable {
         address __owner
     ) external override initializer {
         ERC20.__ERC20_initialize(concat("TrueFi ", _token.name()), concat("tf", _token.symbol()));
-        Ownable.initialize();
+        Claimable.initialize();
         transferOwnership(__owner);
```
vanruch, line 69: This will only set pending owner. And the problem is we can't change _owner as it's private. Do we want to add internal method to Claimable only for this specific case?

vanruch: If we leave it like this, we'll have to initialize a vote in for Governance to claim the ownership which doesn't sound convinient

hal909 (approved):

PR #556, opened 2021-03-30
Make ImplementationReference claimable instead of ownable

skalermo:


vanruch:
Do we want to use claimable pattern for rest of TrueFi2 contracts?

vanruch (approved):

vanruch (approved):

skalermo:
sure

hal909 (changes requested):
Let's go ahead with what we discussed in the meeting today. Use UpgradeableClaimable and rename the file to Claimable.sol. This is the claimable contract we will use in all future contracts.

PR #557, opened 2021-03-30
Add allowAll flag to PoolFactory

skalermo:


contracts/truefi2/PoolFactory.sol
```diff
@@ -63,7 +70,7 @@ contract PoolFactory is Ownable {
      * @param token Address of token to be checked in whitelist
      */
     modifier onlyAllowed(address token) {
-        require(isAllowed[token] == true, "PoolFactory: This token is not allowed to have a pool");
+        require(allowAll == true || isAllowed[token] == true, "PoolFactory: This token is not allowed to have a pool");
```
hal909, line 73: you can just change this to `allowAll && isAllowed[token]`

hal909 (changes requested):
just one fix here, then we're good

vanruch (approved):

PR #568, opened 2021-04-05
Check views before calling TrueLenderReclaimer.reclaimAll() to save gas

yuchenlintt:


contracts/truefi/keepers/TrueLenderReclaimer.sol
```diff
@@ -41,35 +41,70 @@ contract TrueLenderReclaimer {
     }
 
     /**
-     * @dev Reclaim funds from all settled loans in lender
+     * @dev Determine whether lender has fully repaid loans
+     * If called outside of a contract, this view should not cost gas
+     */
+    function hasSettleableLoans() public view returns (bool) {
```
hal909, line 47: since I merged the pull request settling loans that use the repay() function, let's change this contract to also check if a LoanToken has been settled but not reclaimed, and reclaim it if so.

yuchenlintt: If I'm not mistaken, I think that should be what this new version of TrueLenderReclaimer already does. The updated version of the autotask script basically calls something like this:
 
    if (await contract.hasSettleableLoans()) {
      await contract.settleAll();
    }
    if (await contract.hasReclaimableLoans()) {
      await contract.reclaimAll();
    }

The has{Settleable, Reclaimable}Loans() functions are views that let us shortcut if there's no work to be done. Settleable refers to any Withdrawn loan that's fully repaid, and Reclaimable refers to any Settled loan.

hal909 (approved):

PR #570, opened 2021-04-06
Set TrueLender in PoolFactory

skalermo:


hal909 (changes requested):
Is there a reason we need this? Couldn't we store the TrueLender2 address in this contract, rather than passing it in this function? This also opens up the risk of someone passing an invalid TrueLender into the public function, rather than having TrueLender set by owner.

Please change this by adding a function which sets TrueLender, and when you call initialize() pass in that stored address.

contracts/truefi2/PoolFactory.sol
```diff
@@ -79,11 +82,16 @@ contract PoolFactory is IPoolFactory, Claimable {
...
+        ImplementationReference _poolImplementationReference,
+        ERC20 _stakingToken,
+        ITrueLender2 _trueLender
+    ) external initializer {
         Claimable.initialize(msg.sender);
 
         stakingToken = _stakingToken;
         poolImplementationReference = _poolImplementationReference;
+        trueLender = _trueLender;
     }
```
hal909, line 96: Add a function setTrueLender(ITrueLender2 _trueLender2) onlyOwner

This will allow us to change the TrueLender if we ever need to update it.

hal909 (changes requested):
Add an onlyOwner function to set TrueLender

contracts/truefi2/PoolFactory.sol
vanruch: Do we want it? We don't expect TrueLender to be changed and these setters only add points where something can go wrong

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -63,13 +63,15 @@ contract TrueFiPool2 is ITrueFiPool2, ERC20, Claimable {
...
+        ITrueLender2 _lender,
         address __owner
     ) external override initializer {
         ERC20.__ERC20_initialize(concat("TrueFi ", _token.name()), concat("tf", _token.symbol()));
         Claimable.initialize(__owner);
 
         token = _token;
         stakingToken = _stakingToken;
+        lender = _lender;
     }
```
vanruch, line 76: Remove setLender method on line 79

hal909 (approved):

PR #572, opened 2021-04-06
Change rating system

MiksuJak:


contracts/truefi/TrueLender.sol
```diff
@@ -58,11 +58,11 @@ contract TrueLender is ITrueLender, Ownable {
     uint256 public minApy;
     uint256 public maxApy;
 
-    // How many votes in prediction market
-    uint256 public participationFactor;
```
hal909, line 62: we cannot change the order of storage variables by deleting this line.

The pattern we use for deprecating variables is to add _Depricated to the end of the variable name. Please add back any variables deleted here, and add new variables to the bottom of the storage section.

hal909: e.g. participationFactor_Depricated, riskAversion_Depricated

renaming storage variables is okay, changing the order or removing storage variables is not

contracts/truefi/TrueLender.sol
```diff
@@ -58,11 +58,11 @@ contract TrueLender is ITrueLender, Ownable {
...
+    // How many votes are needed for a loan to be approved
+    uint256 public minVotes;
 
-    // How much worse is it to lose $1 TUSD than it is to gain $1 TUSD
-    uint256 public riskAversion;
+    // Minimum ratio of yes votes to note votes for a loan to be approved
+    uint256 public minYesToNoRatio;
```
hal909, line 65: add a comment explaining the precision of this variable. We usually use basis points for ratios, so you can do 10000 = 100%. Then in the constructor set minYesToNoRatio to 8000 = 80%

contracts/truefi/TrueLender.sol
```diff
@@ -189,8 +189,8 @@ contract TrueLender is ITrueLender, Ownable {
 
         minApy = 1000;
         maxApy = 3000;
-        participationFactor = 10000;
-        riskAversion = 15000;
+        minVotes = 15 * (10**6) * (10**8);
+        minYesToNoRatio = 4;
```
hal909, line 193: set to 8000 = 80%

hal909: even though technically this will never be called, it's fine to have it in case we deploy a new Lender

contracts/truefi/TrueLender.sol
```diff
@@ -257,21 +257,21 @@ contract TrueLender is ITrueLender, Ownable {
...
     /**
-     * @dev Set new risk aversion factor. Only owner can change parameters.
-     * @param newRiskAversion New risk aversion factor
+     * @dev Set new yes to no votes ratio. Only owner can change parameters.
+     * @param newMinYesToNoRatio New yes to no votes ratio
      */
-    function setRiskAversion(uint256 newRiskAversion) external onlyOwner {
-        riskAversion = newRiskAversion;
-        emit RiskAversionChanged(newRiskAversion);
+    function setMinYesToNoRatio(uint256 newMinYesToNoRatio) external onlyOwner {
```
hal909, line 272: require newMinYesToNoRatio <= 10000

contracts/truefi/TrueLender.sol
hal909, line 65: rename this variable to minRatio and add a comment:
// minimum ratio of yes to total votes required to pass a loan
// basis precision: 10000 = 100%

hal909: Add a constant at the bottom of storage
uint256 constant BASIS_RATIO = 10000

contracts/truefi/TrueLender.sol
```diff
@@ -464,31 +464,20 @@ contract TrueLender is ITrueLender, Ownable {
...
      */
-    function loanIsCredible(
-        uint256 apy,
-        uint256 term,
-        uint256 yesVotes,
-        uint256 noVotes
-    ) public view returns (bool) {
-        return apy.mul(term).mul(yesVotes).div(365 days) >= noVotes.mul(riskAversion);
+    function loanIsCredible(uint256 yesVotes, uint256 noVotes) public view returns (bool) {
+        return yesVotes >= noVotes.mul(minYesToNoRatio);
```
hal909, line 481: change logic to something like:

```
uint256 total = yes.add(no);
return yes >= total.mul(minRatio).div(BASIS_PRECISION);
```

The function is supposed to return true if the amount of yes votes is at least the minRatio of yes / total

hal909 (changes requested):
- fix storage issues (add new variables to the bottom, don't delete variables
- rename minYesToNoRatio to minRatio &
- change precision of ratio to use basis 10000 = 100%
- add BASIS_PRECISION constant = 10000
- change logic to calculate if a loan is credible

contracts/truefi/TrueLender.sol
MiksuJak: Renaming is actually what I did. I figured that we need 2 variables of size uint256 and want to get rid of 2 of the same type at once.

hal909 (approved):

PR #575, opened 2021-04-07
Make ImplementationReference UpgradeableClaimable

MiksuJak:
Here ImplementationReference uses already existing UpgradeableClaimable instead of developing new Claimable contract. It uses initializer inside of constructor, but it still works like any other non-upgradeable contract would work.
Tests were copied from #556 
Note that it will have conflicts with #574 if we decide to merge it.


hal909 (changes requested):
Since we renamed Claimable to UpgradeableClaimable, this will need to use that new contract. Once that's changed we can merge

hal909 (approved):

PR #582, opened 2021-04-09
Inline 1Inch library

vanruch:


contracts/truefi2/strategies/CurveYearnStrategy.sol
```diff
@@ -23,7 +23,7 @@ interface IERC20WithDecimals is IERC20 {
  * Supports DAI, USDC, USDT and TUSD
  * Curve LP tokens are being deposited into Curve Gauge and CRV rewards can be sold on 1Inch exchange and transferred to the pool
  */
-contract CurveYearnStrategy is Claimable, ITrueStrategy {
+contract CurveYearnStrategy is UpgradeableClaimable, ITrueStrategy {
```
yuchenlintt, line 26: Since we're making this upgradeable, could you put the state variables in a warning comment block?

vanruch: It's done in Hal's PR, I only did it here to fix build after merging 2 parallel PRs broke it

yuchenlintt: Ah, I see. Carry on then, and thanks for the fix!

yuchenlintt (approved):

PR #583, opened 2021-04-12
Make Pool V2 distrustful towards strategies

skalermo:


hal909 (approved):

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -59,6 +59,11 @@ contract TrueFiPool2 is ITrueFiPool2, ERC20, Claimable {
     uint256 private strategyValueCache;
     uint256 private loansValueCache;
 
+    // tolerance difference (percents) between
+    // expected and actual transaction results
+    // when dealing with strategies
+    uint8 public toleratedError;
```
vanruch, line 65: Let's keep it as a constant, will save us some gas

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -474,4 +486,13 @@ contract TrueFiPool2 is ITrueFiPool2, ERC20, Claimable {
...
         return mintedAmount;
     }
+
+    /**
+     * @dev Decrease provided amount percentwise by error
+     * @param amount Amount to decrease
+     * @return Calculated value
+     */
+    function withToleratedError(uint256 amount) internal view returns (uint256) {
+        return amount.mul(10000 - toleratedError * 100).div(10000);
```
vanruch, line 496: Just make `toleratedError` constant and make it equal 98, we'll get something like
        `return amount.mul(TOLERATED_STRATEGY_ERROR).div(100);`

PR #584, opened 2021-04-12
Draft fix for mars deploy script

yuchenlintt:
I'm introducing a new yarn shortcut to deploy contracts using mars. I didn't realize mars was so powerful!

---

1. If everything works (TBD, I've yet to fix the existing script), I'm hoping we can use the following semantics:

    yarn deploy:truefi --network ropsten --dry-run

to simulate deploying on ropsten (maybe add this as a CircleCI check once it succeeds?),

    yarn deploy:truefi --network ropsten --verify --private-key {PRIVATE_KEY}

to actually deploy on ropsten with Etherscan verification, and

    yarn deploy:truefi --verify --private-key {PRIVATE_KEY}

to actually deploy on mainnet with Etherscan verification.

What do you think?

---

2. Is there a way to access the command line flags within a deploy() script? I'd like to conditionally deploy testTrustToken and use shorter time periods when we're deploying to ropsten, and deploy TrustToken with the full cooldown, etc. time periods when we deploy to mainnet.

---

3. I'd prefer contract deployment to be as explicit and verbose as possible, so I took the liberty of manually renaming all the variable names in deployments.json to lowercaseContractName style. Agree or disagree?

---

4. So far I've been able to reduce the deploy script to two 'used before its declaration' loops:
- liquidator <-> stkTruToken
- arbitraryDistributor <-> trueRatingAgency

I'm not familiar enough with these to know how to resolve the dependency graph. Any ideas?


vanruch:
Very cool 👍 
1. Running dry run on CI sounds like a great idea if it won't take very long (it shouldn't)
2. Yes, second parameter in `deploy` callback gives you deployment options
3. That's great, thanks
4. I think only thing you can do is passing empty callback as proxy initializer and then call `initialize` manually if `initialized` is set to false

PR #588, opened 2021-04-13
Usdc support for StkTruToken

MiksuJak:
Include any ERC20 feeToken (probably usdc) support for StkTruToken. Simply includes it in all calculations, claimable updates, transfers, balances etc.

Once all of origination fees in tfusd are paid, we can remove tfusd related methods (keeping claiming only), which would save a lot of gas.

### DIFF
--- a/contracts/governance/StkTruToken.sol
+++ b/contracts/governance/StkTruToken.sol
@@ -134,6 +134,8 @@ contract StkTruToken is VoteToken, StkClaimableContract, IPauseableContract, Ree
         updateClaimableRewards(tru, account);
         updateTotalRewards(tfusd);
         updateClaimableRewards(tfusd, account);
+        updateTotalRewards(feeToken);
+        updateClaimableRewards(feeToken, account);
         _;
     }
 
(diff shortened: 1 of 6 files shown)


contracts/governance/StkTruToken.sol
```diff
@@ -176,6 +187,15 @@ contract StkTruToken is VoteToken, StkClaimableContract, IPauseableContract, Ree
         initalized = true;
     }
 
+    /**
+     * @dev Set tfUSDC address
+     * @param _feeToken Address of tfUSDC to be set
+     */
+    function setFeeToken(IERC20 _feeToken) external onlyOwner {
```
vanruch, line 194: It might be unsafe to change feeToken after some fees were paid. Maybe check if totalFarmRewards for the token is 0?

test/governance/StkTruToken.test.ts
```diff
@@ -446,7 +514,7 @@ describe('StkTruToken', () => {
 
     it('gas cost', async () => {
       const tx = await (await stkToken.transfer(staker.address, amount.div(2), { gasLimit: 300000 })).wait()
-      expect(tx.gasUsed).to.be.lt(150000)
+      expect(tx.gasUsed).to.be.lt(200000)
```
vanruch, line 517: 😔

hal909 (approved):

PR #589, opened 2021-04-14
Swap loan fee for USDC

vanruch:


contracts/truefi2/TrueLender2.sol
```diff
@@ -253,16 +295,46 @@ contract TrueLender2 is ITrueLender2, Claimable {
...
         loanToken.redeem(loanToken.balanceOf(address(this)));
         uint256 balanceAfter = pool.token().balanceOf(address(this));
 
         // gets reclaimed amount and pays back to pool
         fundsReclaimed = balanceAfter.sub(balanceBefore);
-        pool.token().approve(address(pool), fundsReclaimed);
-        pool.repay(fundsReclaimed);
+        uint256 fee = _swapFee(pool.token(), loanToken, data);
+
+        pool.token().approve(address(pool), fundsReclaimed.sub(fee));
```
hal909, line 312: we need to deposit the fee into a feePool (tfUSDC), and pay the fee in that token. So after the swap deposit the token into the pool, transfer the pool tokens to the staking contract.

hal909 (changes requested):
We need to deposit the feeToken into a feePool (USDC) after performing the swap

hal909 (approved):

PR #594, opened 2021-04-15
Make collectFees in pool public

vanruch:


vanruch:
@hal909 do we need this? Or maybe we want to get rid of pool fees? Anyway, having onlyOwner claimFee would be uncomfortable.

yuchenlintt (commented):
The change looks good to me overall. I think this way is certainly much better than forcing the owner to sign off on every collectFees() transaction.

LGTM, but wait for Hal's official approval before merging.

---

Ancillary question: who do we expect `beneficiary` to be set to? Your unit tests set it to `owner`, but if that's always the case then there's no need to store a separate state variable to track it. If it's a distributor, then we could enforce that type instead having an arbitrary `address` as `beneficiary`.

hal909 (approved):

PR #596, opened 2021-04-15
TrueFiPool2 tests

skalermo:


test/truefi2/TrueFiPool2.test.ts
```diff
@@ -160,6 +158,57 @@ describe('TrueFiPool2', () => {
     })
   })
 
+  describe('liquidValue', () => {
+    const includeFee = (amount: BigNumber) => amount.mul(10000).div(9975)
```
yuchenlintt, line 162: Thanks for moving this inside the smallest scope it's needed in!

Could you rename `includeFee()` to be more explicit? It took me multiple readthroughs of your tests to realize this is linked with the pool's 0.25% joining fee. Something like `include025PercentJoiningFee()`, perhaps?

test/truefi2/TrueFiPool2.test.ts
```diff
@@ -220,13 +268,202 @@ describe('TrueFiPool2', () => {
...
+        LoanToken2Factory,
+        pool.address,
+        borrower.address,
+        lender.address,
+        AddressZero,
+        parseEth(1e6),
+        DAY * 365,
+        1000,
+      )
+      await rater.mock.getResults.returns(0, 0, parseTRU(15e6))
```
yuchenlintt, line 300: If I'm reading this right, this is 15e6 = 1e7 * 150% because we're time traveling 365/2 days (half the loan term) in this test.

Is this correct? If so, could you make the calculations (here and elsewhere) explicit? Being explicit makes the tests much more legible, if we're trying to puzzle out why it's failing months down the line!

test/truefi2/TrueFiPool2.test.ts
```diff
@@ -220,13 +268,202 @@ describe('TrueFiPool2', () => {
...
+      await lender.connect(borrower).fund(loan2.address)
+
+      const liquidValue = await pool.liquidValue()
+      const totalSupply = await pool.totalSupply()
+      const exitAmount = totalSupply.div(2)
+
+      await pool.exit(exitAmount)
+      expect(await tusd.balanceOf(owner.address)).to.equal(exitAmount.mul(liquidValue).div(totalSupply))
+      expectCloseTo(await loan1.balanceOf(owner.address), parseEth(55e4), 10)
+      expectCloseTo(await loan2.balanceOf(owner.address), parseEth(625e3), 10)
```
yuchenlintt, line 343: I'm not sure I get where the 55e4 and 625e3 come from. Could you show the calculation?

test/truefi2/TrueFiPool2.test.ts
```diff
@@ -220,13 +268,202 @@ describe('TrueFiPool2', () => {
...
+      await pool.join(amount)
+      await pool.switchStrategy(poolStrategy1.address)
+    })
+
+    it('burns pool tokens on exit', async () => {
+      const supply = await pool.totalSupply()
+      await pool.liquidExit(supply.div(2))
+      expect(await pool.totalSupply()).to.equal(supply.div(2))
+      await pool.liquidExit(supply.div(3))
+      expect(await pool.totalSupply()).to.equal(supply.sub(supply.mul(5).div(6)))
```
yuchenlintt, line 427: Couldn't supply - (supply * 5 / 6) be simplified to supply / 6?

yuchenlintt (changes requested):
Thanks for adding tests! Looks good overall, just a couple minor points to make the tests easier to read.

hal909 (approved):

PR #601, opened 2021-04-19
Use legacy TUSD pool with TrueFi2

vanruch:


hal909 (commented):
still reviewing this, I'm going to have Ben Cosman take a look at it because this is a very risky change, and will not be included in the audit.

hal909 (approved):

BenjaminCosman:
I see various 1inch stuff being removed in this PR. The version with 1inch was never actually deployed, right? As long as that's true, this all looks good to me. (But if the 1inch version *was* deployed, then since _lender2 is taking _1inchExchange's storage slot, it won't start out as 0x0 so this will break.)

PR #603, opened 2021-04-19
Add a public view of Initializable's initialized variable

yuchenlintt:
I think adding this view is reasonable so we can check initialization status, but perhaps we should be more cautious about making changes to library code.

Two possible concerns:
1. If OpenZeppelin updates this contract and we want to merge from upstream, would this present issues?
2. Would a change like this require that we redeploy impls that we'd rather not touch? May not be worth it in that case.

### DIFF
--- a/contracts/common/Initializable.sol
+++ b/contracts/common/Initializable.sol
@@ -63,6 +63,7 @@ contract Initializable {
 
     /**
      * @dev Return true if and only if the contract has been initialized
+     * @return whether the contract has been initialized
      */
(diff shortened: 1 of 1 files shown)


contracts/common/Initializable.sol
```diff
@@ -60,6 +61,13 @@ contract Initializable {
         return cs == 0;
     }
 
+    /**
+     * @dev Return true if and only if the contract has been initialized
```
hal909, line 65: add @return documentation

hal909 (approved):

PR #605, opened 2021-04-20
TrueFi vault for 180 day vested tokens

hal909:


hal909 (commented):
@yuchenlintt We need to add a function which allows the vault beneficiary to rate loans in the TrueRatingAgencyV2 contract

contracts/governance/TrueFiVault.sol
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @dev Rate NO on a loan by staking TRU
+     * @param id Loan ID
+     */
+    function rateLoanNo(address id) external onlyBeneficiary {
+        ratingAgency.no(id);
+    }
+
```
hal909, line 120: Add one more function here to remove rating (i.e. vote 0 yes and 0 no)

yuchenlintt: ITrueRatingAgencyV2 doesn't have that function visible, so I added it to the interface in PR #610. Will update it here as well once that's checked in.

hal909: done. this looks fine now as is

yuchenlintt:
> @yuchenlintt We need to add a function which allows the vault beneficiary to rate loans in the TrueRatingAgencyV2 contract

Done. This PR has been ready for review since two days ago, but the branch has been failing 4 Timelock tests of unknown cause.

BenjaminCosman:
If I was the beneficiary of one of these vaults and also owned some normal TRU of my own, I would want a way to delegate the vault's votes to me so I could vote from my own account in a single transaction. (I suppose I could instead delegate *my* votes to the *vault*, but that feels bizarre.) One solution would be to add delegation as an option, but perhaps a simpler solution (unless there's a reason someone would *not* want to delegate to themselves?) is to just automatically delegate on initialization, and then you could entirely remove castVote, resetCastLoanRatings, rateLoanYes, and rateLoanNo.

yuchenlintt:
Hi Ben!

Yeah I agree, {VoteToken delegation} sounds like a much better option than {OOP delegation pattern} composition. And certainly the only correct direction to delegate is from the vault to the beneficiary, otherwise:
1.  you could get confusing interactions if a beneficiary has two vaults, and
2. a malicious owner could grab all of a beneficiary's non-vault delegate votes by calling withdrawToOwner(). (Not that we plan to be evil, but we should still mitigate that risk.)

I prefer your simpler solution. It minimizes API surface and gives users one opinionated "correct" way to vote + rate.

I'll work on updating this PR to use VoteToken delegation and add some tests for it. Thanks!

BenjaminCosman:
What do you think of having _withdrawTo automatically call claimRewards? Mostly just for convenience since I can't think of why you'd withdraw everything *without* first claiming rewards. Plus otherwise, the only way to get *all* the rewards you've earned is actually *three* transactions (withdraw-claim-withdraw) - the rewards left behind from claim-withdraw is miniscule, but it's the principle of the matter :)

yuchenlintt:
Huh, that's a really good point. Then I think this current version of withdrawToOwner() might lead to rewards getting stuck, because claimRewards() is currently onlyBeneficiary.

Presumably we would only call withdrawToOwner() if the grant was made in error, the beneficiary lost their private keys, the beneficiary breached some legal agreement, etc. Could a malicious beneficiary do whatever bad thing makes us call withdrawToOwner(), and then just wait 180 days until expiry to claimRewards() and withdrawToBeneficiary()?

yuchenlintt:
I'm starting to think the best way to resolve this malicious beneficiary issue is to set `beneficiary = owner` inside withdrawToOwner() and replace _withdrawTo(recipient) with a private _withdrawToBeneficiary() that skips the expiry check.

We would then be able to have _withdrawToBeneficiary() call claimRewards() without removing claimRewards()'s onlyBeneficiary check.

I'll work on a commit to see if this pans out.

yuchenlintt (approved):
I'm not sure who this PR is blocked on: Hal, Ben, or me. But I think the change is at least ready to merge, although I'd like Ben's audit approval before launch.

Since this is technically Hal's PR (and hence he might be unable to approve), I'm giving approval now but letting one of you merge in case you had more to say.

BenjaminCosman (approved):
Oh, sorry if it was blocked on me; looks good to merge!

PR #608, opened 2021-04-20
Sushiswap rewarder

yuchenlintt:
Is this how we want to implement the Sushiswap Rewarder? Let me know if there's anything you'd like to change before I start writing tests for this. As I recall we mentioned a target of `_rewardMultiplier = 100` for now, so I'm planning to go with that in unit testing.


hal909 (commented):
we're still waiting on a response from the SUSHI team as to whether or not we can use a proxy. Once we get that, we can accordingly make this have a constructor or initializer.

hal909 (approved):

PR #609, opened 2021-04-20
Hardhat integration

MiksuJak:


contracts/truefi/distributors/ArbitraryDistributor.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.6.10;
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
 
-import {Ownable} from "../common/UpgradeableOwnable.sol";
+import {Ownable} from "../../common/UpgradeableOwnable.sol";
```
hal909, line 7: Are these just to make the compiler work? Looks like Waffle was smart enough to figure out where this file was.

scripts/deploy_mock_governance.ts
```diff
@@ -4,29 +4,29 @@
 import { ethers, providers } from 'ethers'
 
 import {
-  TestTrustTokenFactory,
-  MockTimeLockFactory,
-  GovernorAlphaFactory,
-  OwnedUpgradeabilityProxyFactory,
-} from '../build'
+  TestTrustToken__factory,
```
hal909, line 7: is there a reason we need to use this new double underscore? Not really an issue just curious

hal909 (approved):

scripts/deploy_mock_governance.ts
vanruch: We updated typechain and new version has namings like that to avoid conflicts like wi had with LoanToken/LoanTokenFactory

vanruch (changes requested):
block merge

vanruch (approved):

PR #611, opened 2021-04-23
Deploy TrueFi on Ropsten

vanruch:
Could not deploy on Ropsten because apparently gas block limit is 3M now which is not enough to deploy some of our contracts


yuchenlintt (changes requested):
Thanks for working on this (and in particular, fixing TrueFiPool deployment)!

Overall the PR looks good, though I think we should just distinguish between `mainnet` and `testnet` deployParams. That way all our testnets would be consistent with each other, and there's less to update when we want to change something in test.

My other big comment is to keep the ropsten addresses in deployments.json, and just append kovan addresses. I'm not actually sure how much work it'll be for product to switch to testing on kovan, but regardless let's not hide the existing info.

I think it's also a good idea to keep the `deploy` CircleCI on a clean testnet to make sure all deploy steps continue working. If we've started recording kovan deployments, maybe switch CircleCI to gorli? We can do this in a separate PR.

yuchenlintt:
After an offline chat with Hal, I'll take over this PR, make these changes myself, and try again with Ropsten. Sounds like the gas block limit might be gone now.

Have a good weekend!

yuchenlintt:
I merged in my changes from the mars-deployments branch and then successfully deployed everything to Ropsten!

The addresses in deployments.json now accurately reflect what the proxies point to in Ropsten. I'll update the spreadsheet later today.

yuchenlintt (approved):

yuchenlintt:
Thanks for fixing the Timelock CI issue!

In the future, could you merge in changes from dev after people have begun review? Your rebase mangled the change history of this PR.

PR #618, opened 2021-04-27
Update mainnet addresses to match spreadsheet

yuchenlintt:


deployments.json
```diff
@@ -124,17 +124,105 @@
...
+    },
+    "ratingAgencyV2Distributor": {
+      "txHash": "0xa95a7a398b323c81e9027235e2893537c42cdfbf7769e73f8af59386d556e813",
+      "address": "0x87d1616B9B3a0fD756EF4B4Abff29B30ab813f42"
+    },
+    "ratingAgencyV2Distributor_proxy": {
+      "txHash": "0xbc3ebd3bc5ee273dd84811e77c067e629d786effa78099303f71998bd986ba70",
+      "address": "0x6151570934470214592AA051c28805cF4744BCA7"
+    },
+    "timelock": {
```
hal909, line 211: remove timelock, timelock_proxy, govornorAlpha, and govornorAlpha_proxy.

these contracts are not going to be used, we can add them when we deploy new contracts

yuchenlintt: This PR just updates deployments.json to match the current spreadsheet, for any addresses that would be touched by `yarn deploy:truefi`. At the moment this is mostly meant to cut down on areas where we're publishing two statements that say different things, even though they should be the same. And it lets us confirm (in a dry run) which contracts will be updated if we were to run the script.

If we decide to not use these contract addresses in the future, then we can always delete them at that time. But I'm wary of introducing data discrepancies *now* based on *future* actions we plan to take. I think the right time to delete these addresses should be when we deploy new ones to replace them.

That all said, I'm not strongly attached to keeping these addresses, if you insist.

hal909: that's fine

hal909 (approved):

PR #623, opened 2021-04-28
SushiRewarder timelock

MiksuJak:
This PR introduces new Proxy for sushirewarder. There are 2 variables, that can only be changed with a certain delay. Implementation and of course delay. For both of them there are 2 types of functions that need to be called, in order to set new values.
 - initialize(SetDelay/UpgradeTo) sets new pending value and calculates timestamp, in which setter can be executed. This function can be only called by owner of the proxy.
 - execute(SetDelay/UpgradeTo) ensures that needed timestamp has passed and executes the setter. Since execute call will only set predeclared pending value and will only be able to set it once, it is not restricted for owner use only.

It does not use Timelock, beceause Timelock itself is initializable and could not be relied on that much by external users.

### DIFF
--- a/contracts/proxy/TimeLockedOwnedUpgradeabilityProxy.sol
+++ b/contracts/proxy/TimeLockedOwnedUpgradeabilityProxy.sol
@@ -169,7 +169,14 @@ contract TimeLockedOwnedUpgradeabilityProxy {
         assembly {
             sstore(position, _delay)
         }
+        emit DelayChanged(_delay);
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


hal909 (changes requested):
My fault for not making this clear - we want to use a Timelock contract to manage this farm, not create a proxy that delays upgrades. This contract *might* be useful in the future though, if this is a feature we want.

I went ahead and implemented the Sushiswap timelock in a new PR.

hal909 (approved):

PR #624, opened 2021-04-28
sushi farm with timelock

hal909:


contracts/truefi/SushiTimelock.sol
```diff
@@ -0,0 +1,133 @@
...
+
+        bytes memory callData;
+
+        if (bytes(signature).length == 0) {
+            callData = data;
+        } else {
+            callData = abi.encodePacked(bytes4(keccak256(bytes(signature))), data);
+        }
+
+        // solhint-disable-next-line avoid-call-value, avoid-low-level-calls
```
yuchenlintt, line 120:
```suggestion
        // YYY: // solium-disable-next-line security/no-call-value
        // solhint-disable-next-line avoid-call-value, avoid-low-level-calls
```

contracts/truefi/SushiTimelock.sol
```diff
@@ -0,0 +1,133 @@
...
+        (bool success, bytes memory returnData) = target.call.value(value)(callData);
+        require(success, "Timelock::executeTransaction: Transaction execution reverted.");
+
+        emit ExecuteTransaction(txHash, target, value, signature, data, eta);
+
+        return returnData;
+    }
+
+    function getBlockTimestamp() internal view returns (uint) {
+        // solhint-disable-next-line no-block-members
```
yuchenlintt, line 130:
```suggestion
        // YYY: // solium-disable-next-line security/no-block-members
        // solhint-disable-next-line no-block-members
```

scripts/deploy_sushiswap_farm.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const sushiFarmImpl = await (await new TruSushiswapRewarder__factory(wallet).deploy(txnArgs)).deployed()
+  console.log(`sushiFarmImpl: ${sushiFarmImpl.address}`)
+
+  const sushiFarm = await (await new OwnedUpgradeabilityProxy__factory(wallet).deploy(txnArgs)).deployed()
+  console.log(`sushiFarm: ${sushiFarm.address}`)
+
+  const timelock = await (await new SushiTimelock__factory(wallet).deploy(ownerAddress, 172800, txnArgs)).deployed()
+  console.log(`timelock: ${timelock.address}`)
+
```
yuchenlintt, line 31: Don't you also need an upgradeTo() call from the proxy to the impl?

scripts/deploy_sushiswap_farm.ts
```diff
@@ -0,0 +1,36 @@
...
+  const masterChefAddress = '0xc2EdaD668740f1aA35E4D8f227fB8E17dcA888Cd'
+  const truAddress = '0x4c19596f5aaff459fa38b0f7ed92f11ae6543784'
+
+  const sushiFarmImpl = await (await new TruSushiswapRewarder__factory(wallet).deploy(txnArgs)).deployed()
+  console.log(`sushiFarmImpl: ${sushiFarmImpl.address}`)
+
+  const sushiFarm = await (await new OwnedUpgradeabilityProxy__factory(wallet).deploy(txnArgs)).deployed()
+  console.log(`sushiFarm: ${sushiFarm.address}`)
+
+  const timelock = await (await new SushiTimelock__factory(wallet).deploy(ownerAddress, 172800, txnArgs)).deployed()
```
yuchenlintt, line 29: Where and how is SushiTimelock used to delay upgrades to the TruSushiswapRewarder? As far as I can tell they're just two unrelated contracts being deployed by the same script.

yuchenlintt (changes requested):
For audit clarity I think we should keep the XXX diffs and add our own YYY diffs to show what we've changed on top of Sushi.

Maybe I don't fully understand how Timelock interacts with the proxy, but from the script it just looks like we're deploying two unrelated contracts. Could you explain how this works?

Also, don't forget to upgradeTo the impl!

---

(If you want) an alternative to writing another deploy script like this is to add a couple lines to `deploy/truefi.ts`:

```
const timelock = contract(SushiTimelock, [ownerAddress, 172800])
const sushiFarm = proxy(contract(TruSushiswapRewarder), 'initialize', [rewardMultiplier, truAddress, masterChefAddress])
sushiFarm.transferOwnership(ownerAddress)
```

contracts/truefi/SushiTimelock.sol
```diff
@@ -0,0 +1,133 @@
...
+// 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Ctrl+f for XXX to see all the modifications.
```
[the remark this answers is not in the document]
hal909: no modifications to the code were made other than changing the version of solidity

scripts/deploy_sushiswap_farm.ts
hal909: we will set them up through the multisig

scripts/deploy_sushiswap_farm.ts
hal909: yes good catch :)

scripts/deploy_sushiswap_farm.ts
yuchenlintt: Gotcha. Looks good then!

yuchenlintt (approved):

PR #627, opened 2021-04-29
Introduce TrueMultiFarm

barrutko:


contracts/truefi2/TrueMultiFarm.sol
```diff
@@ -0,0 +1,376 @@
...
+
+    /**
+     * @dev Set shares for farms
+     * Example: setShares([DAI, USDC], [1, 2]) will ensure that 33.(3)% of rewards will go to DAI farm and rest to USDC farm
+     * If later setShares([DAI, TUSD], [2, 1]) will be called then shares of DAI will grow to 2, shares of USDC won't change and shares of TUSD will be 1
+     * So this will give 40% of rewards going to DAI farm, 40% to USDC and 20% to TUSD
+     * @param tokens Token addresses
+     * @param updatedShares share of the i-th token in the multifarm
+     */
+    function setShares(IERC20[] calldata tokens, uint256[] calldata updatedShares) public onlyOwner {
```
hal909, line 178: Would it be better to do something more similar to MasterChef (https://github.com/sushiswap/sushiswap/blob/master/contracts/MasterChef.sol)

In MasterChef, there is simply a total allocation points, and each farm has a certain number of points. Adding points to a farm increases the number of total allocation points, removing points from a farm decreases the total allocation points. Masterchef then has a total block reward which is split based on farm_points / total_points. This might simplify this system unless you think setting shares in a single call is more efficient.

hal909 (changes requested):
looks pretty good so far, I had one idea for how we track farm shares in this contract, let me know what you think

contracts/truefi2/TrueMultiFarm.sol
vanruch: Looks like Masterchef does it similarly to us https://github.com/sushiswap/sushiswap/blob/master/contracts/MasterChef.sol#L106
We only added an option of changing multiple allocation points or shares in our case. It doesn't complicate the code by much and can be a handy feature considering we have Governance and don't want to have lots of polls to rebalance the farm

barrutko:
> looks pretty good so far, I had one idea for how we track farm shares in this contract, let me know what you think
Here comes some an example
image

hal909 (approved):

hal909:
Feedback from ben:
#1
(https://github.com/trusttoken/audits/blob/master/TrueFi/truefi_2.0_audit_ben_cosman_2_17_2020.pdf) entitled "Differing units should be better documented" - I don't think it's reasonable that e.g. fields named totalClaimedRewards and totalRewards are just expected to be off from each other by a factor of 10^30 with no documentation of that fact

#2
If I'm reading this right, it looks like once the distributor is distributing rewards, if a farm ever has 0 tokens staked (or similarly for the meta level, if the entire multifarm has assigned 0 shares across its sub-farms), then any rewards accumulated to that farm during that period are just locked in the contract forever? perhaps this never happens in practice, since people join a farm early in anticipation of a scheduled start date? and similarly at the meta level you can just set shares to non-zero before distribution begins? But if you think there might actually be pockets of stuck TRU, then I would consider burning the extra, or alternatively if you want to award the accumulating TRU to whoever joins first then you could enclose the entirety of _updateTotalRewards in the if-statement that currently guards only part of it (and similarly for _updateCumulativeRewardPerShare)

hal909 (approved):

PR #630, opened 2021-04-30
Curve base rate oracle

skalermo:


contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,62 @@
...
+
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+import "../truefi/interface/ICurve.sol";
+
+// prettier-ignore
+contract CrvBaseRateOracle {
+    using SafeMath for uint256;
+    uint8 public constant BUFFER_SIZE = 7;
+
+    ICurve public curve;
```
vanruch, line 12: Don't mix storage variables with compile-time structs, move this below HistoricalRatesBuffer

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,62 @@
...
+    function calculateAverageRate() public view returns (uint256) {
+        // Average rate is calculated based on formula below:
+        //           v_n - v_1
+        // avgRate = ---------
+        //           t_n - t_1
+        //
+        // where v_n is the most recent rate, v_1 is the oldest rate,
+        // t_n and t_1 are their timestamps respectively
+        uint8 idxOfOldestRate = histBuffer.insertIndex;
+        uint256 v_n = histBuffer.baseRates[(idxOfOldestRate - 1) % BUFFER_SIZE];
```
vanruch, line 46: This will result in underflow if idxOfOldestRate is 0. And 255 % 7 = 3 instead of expected 7 so we need to make additional checks for this case

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    function rate() public view returns (uint256, uint256, uint256) {
+        require(histBuffer.timestamps[histBuffer.insertIndex] != 0, "CrvBaseRateOracle: histBuffer must be filled up");
+        uint256 curCrvBaseRate = curve.get_virtual_price();
+        uint256 immediateProfit = calculateAverageRate();
+        uint256 weeklyProfit = immediateProfit.mul(7 days).div(curCrvBaseRate);
+        uint256 monthlyProfit = immediateProfit.mul(30 days).div(curCrvBaseRate);
+        uint256 yearlyProfit = immediateProfit.mul(365 days).div(curCrvBaseRate);
+        return (weeklyProfit, monthlyProfit, yearlyProfit);
```
vanruch, line 60: 👍 I like returning 3 types of profits

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,62 @@
...
+    struct HistoricalRatesBuffer {
+        uint256[BUFFER_SIZE] baseRates;
+        uint256[BUFFER_SIZE] timestamps;
+        uint8 insertIndex;
+    }
+
+    HistoricalRatesBuffer public histBuffer;
+
+    constructor(ICurve _curve) public {
+        curve = _curve;
```
vanruch, line 23: Let's fill the buffer with current values on contract creation

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,62 @@
...
+        histBuffer.timestamps[iidx] = block.timestamp;
+        histBuffer.baseRates[iidx] = curve.get_virtual_price();
+        histBuffer.insertIndex = (iidx + 1) % BUFFER_SIZE;
+    }
+
+    function calculateAverageRate() public view returns (uint256) {
+        // Average rate is calculated based on formula below:
+        //           v_n - v_1
+        // avgRate = ---------
+        //           t_n - t_1
```
vanruch, line 41: Add explanation how this formula was derived

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    function updateRate() public {
+        uint8 iidx = histBuffer.insertIndex;
+        histBuffer.timestamps[iidx] = block.timestamp;
+        histBuffer.baseRates[iidx] = curve.get_virtual_price();
+        histBuffer.insertIndex = (iidx + 1) % BUFFER_SIZE;
+    }
+
+    function calculateAverageRate() public view returns (uint256) {
```
vanruch, line 37: Let's make `calculateAverageRate` accept `curCrvBaseRate` instead of latest measurement. t_n would then be block.timestamp and rates would be more dynamic

yuchenlintt (changes requested):

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,101 @@
...
+        // Average rate is calculated by taking
+        // the time-weighted average of the speeds
+        // of virtual price changes.
+        //
+        // let 'd_{i,j}' be the speed of virtual price change:
+        //           v_i - v_j
+        // d_{i,j} = --------- , i != j
+        //           t_i - t_j
```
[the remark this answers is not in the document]
skalermo: Yeah, I was having doubts about what I should take average of. When the formula got heavily simplified I was a bit surprised and somewhat satisfied, but I was just wrong. Your formula seems to be the right one.

skalermo:
@yuchenlintt Thanks for the review! I'll implement your suggestions as soon as I can.

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,101 @@
...
+        return (v_n.sub(v_1)).mul(100_00).div(t_n.sub(t_1));
+    }
+
+    function rate() public view returns (uint256, uint256, uint256) {
+        uint256 curCrvBaseRate = curve.get_virtual_price();
+        uint256 avgRate = calculateAverageRate();
+        uint256 weeklyProfit = avgRate.mul(7 days).div(curCrvBaseRate);
+        uint256 monthlyProfit = avgRate.mul(30 days).div(curCrvBaseRate);
+        uint256 yearlyProfit = avgRate.mul(365 days).div(curCrvBaseRate);
+        return (weeklyProfit, monthlyProfit, yearlyProfit);
```
BenjaminCosman, line 99: What's the purpose of having these 3 different return values if they're all just constant factors off from each other? it seems almost misleading, since I would expect just from hearing the interface that these would be calculated from different time resolutions, e.g. maybe Curve has been low-yield over the past year as a whole but high-yield over the past month.

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,110 @@
...
+    uint256 public cooldownTime;
+
+    uint8 public constant BUFFER_SIZE = 7;
+
+    /**
+     * @dev Throws if cooldown is on when updating the historical buffer
+     */
+    modifier offCooldown() {
+        uint256 lastUpdated = histBuffer.timestamps[histBuffer.insertIndex.add(BUFFER_SIZE).sub(1) % BUFFER_SIZE];
+        require(now >= lastUpdated.add(cooldownTime), "CrvBaseRateOracle: Buffer on cooldown");
```
BenjaminCosman, line 35: `now` has been deprecated in newer versions of Solidity, so I'd suggest following their lead and using `block.timestamp` instead.

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,110 @@
...
+        _;
+    }
+
+    constructor(ICurve _curve) public {
+        curve = _curve;
+        cooldownTime = 1 days;
```
BenjaminCosman, line 41: I feel like this makes more sense as an _argument_ to the constructor rather than a hardcoded constant? works either way though

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,110 @@
...
+     * Formula is given below:
+     *
+     *           sum_{i=1}^{n - 1} v_i * (t_i - t_{i-1})
+     * avgRate = ---------------------------------------
+     *                      (t_{n-1} - t_0)
+     *
+     * where v_i, t_i are values of the prices and their respective timestamps
+     * stored in the historical buffer. Index n-1 corresponds to the most
+     * recent values and index 0 to the oldest ones.
+     * @return Average rate in percentage with precision
```
BenjaminCosman, line 85: Reader will not know how to interpret "in percentage with precision". Would it be accurate to just say "in basis points"?

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,110 @@
...
+     */
+    function calculateAverageRate() public view returns (uint256) {
+        uint8 iidx = histBuffer.insertIndex;
+        uint256 sum;
+        for (uint8 i = 1; i < BUFFER_SIZE; i++) {
+            uint8 idx = uint8(iidx.add(i) % BUFFER_SIZE);
+            uint8 prevIdx = uint8(iidx.add(i).sub(1) % BUFFER_SIZE);
+            uint256 dt = histBuffer.timestamps[idx].sub(histBuffer.timestamps[prevIdx]);
+            sum = sum.add(histBuffer.baseRates[idx].mul(dt));
```
BenjaminCosman, line 94: So we're basically taking the price _during_ each interval to be equal to the price at the _end_ of that interval. Would it be more accurate to average the interval's start and end instead? so `histBuffer.baseRates[idx].add(histBuffer.baseRates[prevIdx]).mul(dt).div(2)`

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * where v_i, t_i are values of the prices and their respective timestamps
+     * stored in the historical buffer. Index n-1 corresponds to the most
+     * recent values and index 0 to the oldest ones.
+     * @return Average rate in percentage with precision
+     */
+    function calculateAverageRate() public view returns (uint256) {
+        uint8 iidx = histBuffer.insertIndex;
+        uint256 sum;
+        for (uint8 i = 1; i < BUFFER_SIZE; i++) {
```
BenjaminCosman, line 90: Currently we average only using the historical buffer, which notably ignores the most current available price. Would it be better to include now's price as well? So effectively a virtual entry at the end of the buffer with timestamp `block.timestamp` and price `curve.get_virtual_price()`, averaged in just like any other entry.

BenjaminCosman (changes requested):
See earlier comments. Also, I don't understand what `weeklyMonthlyYearlyProfit` is really for; a comment on that would be good.

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,149 @@
...
+        require(
+            bufferSizeNeeded <= bufferSize(),
+            "CrvBaseRateOracle: Needed buffer size cannot exceed size limit"
+        );
+        uint16 iidx = histBuffer.insertIndex;
+        uint256 sum;
+        uint256 totalTime;
+        for (uint16 i = 1; i < bufferSizeNeeded; i++) {
+            uint16 prevIdx = uint16(iidx.add(bufferSize()).sub(i).sub(1) % bufferSize());
+            if (histBuffer.timestamps[prevIdx] == 0)
```
vanruch, line 106: Always use {} in ifs/fors etc, we don't want to repeat [Apple's](https://blog.codecentric.de/en/2014/02/curly-braces/) mistakes

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,149 @@
...
+                break;
+            uint16 idx = uint16(iidx.add(bufferSize()).sub(i) % bufferSize());
```
vanruch, line 108: Is it same as `(prevIdx + 1) % bufferSize`?

test/truefi2/CrvBaseRateOracle.test.ts
```diff
@@ -0,0 +1,251 @@
...
+      await updateRateRightAfterCooldown()
+      const curTimestamp = await getCurrentTimestamp()
+      const [baseRates, timestamps, insertIndex] = await crvBaseRateOracle.getHistBuffer()
+      expect(baseRates[0]).to.eq(100)
+      expect(timestamps[0]).to.eq(curTimestamp)
+      expect(insertIndex).to.eq(1)
+    })
+  })
+
+  describe('calculateAverageRate', () => {
```
vanruch, line 128: Could you write in every comment what the expected value is roughly equal to to see if it make sense?

vanruch (changes requested):

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,157 @@
...
+            uint16 idx = uint16(prevIdx.add(1) % bufferSize());
+            uint256 dt = histBuffer.timestamps[idx].sub(histBuffer.timestamps[prevIdx]);
+            sum = sum.add(histBuffer.baseRates[idx].add(histBuffer.baseRates[prevIdx]).mul(dt));
+            totalTime = totalTime.add(dt);
+        }
+        uint256 curCrvBaseRate = curve.get_virtual_price();
+        uint256 curTimestamp = block.timestamp;
+        uint16 idx = uint16(iidx.add(bufferSize()).sub(1) % bufferSize());
+        sum = sum.add(curCrvBaseRate.add(histBuffer.baseRates[idx]).mul(curTimestamp.sub(histBuffer.timestamps[idx])));
+        totalTime = totalTime.add(curTimestamp.sub(histBuffer.timestamps[idx]));
```
BenjaminCosman, line 121: `curTimestamp.sub(histBuffer.timestamps[idx])` is long enough that it should be factored out into a variable (`dt` again?)

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,157 @@
...
+        uint256 curCrvBaseRate = curve.get_virtual_price();
+        return ((int256(curCrvBaseRate.mul(10000)) - avgRate) * 10000) / avgRate;
+    }
+
+    /**
+     * @dev Profit in 30 days based on average rate
+     * from collected data from last 30 days
+     * @return Calculated monthly profit
+     */
+    function monthlyProfit() public view returns (int256) {
```
BenjaminCosman, line 141: These three functions are almost identical; factor out the identical parts into an internal helper function.

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,157 @@
...
+        totalTime = totalTime.add(curTimestamp.sub(histBuffer.timestamps[idx]));
+        return sum.mul(10000).div(2).div(totalTime);
+    }
+
+    /**
+     * @dev Profit in 7 days based on average rate
+     * from collected data from last 7 days
+     * @return Calculated weekly profit
+     */
+    function weeklyProfit() public view returns (int256) {
```
BenjaminCosman, line 130: I still don't understand what these three functions are really for - why would someone want to know the difference between the current price and the average price over the last week? As far as I can tell, none of the other contracts call this function.

BenjaminCosman (changes requested):

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,158 @@
...
+     * Index n-1 corresponds to the most recent values and index 0 to the oldest ones.
+     * Notice that whether we are going to use the whole buffer or not
+     * depends on value of timeToCover parameter.
+     * @param timeToCover For how much time average should be calculated
+     * @return Average rate in basis points
+     */
+    function calculateAverageRate(uint256 timeToCover) public view returns (uint256) {
+        require(1 days <= timeToCover && timeToCover <= 365 days, "CrvBaseRateOracle: Expected amount of time in range 1 to 365 days");
+        // estimate how much buffer we need to use
+        uint256 bufferSizeNeeded = timeToCover.div(cooldownTime);
```
BenjaminCosman, line 102: In practice, I would guess that `updateRate` will be called noticeably later each day, which means that the actual buffer size needed may end up smaller than this value. Perhaps remove this variable entirely, and instead break out of the loop either right before or right after the point when `totalTime > timeToCover`?

BenjaminCosman (changes requested):

BenjaminCosman:
(Note that one of my comments is on a 'resolved' issue - I don't know how to mark it as not resolved.)

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,158 @@
...
+    constructor(ICurve _curve, uint256 _cooldownTime) public {
+        curve = _curve;
+        cooldownTime = _cooldownTime;
+
+        totalsBuffer.lastRate = curve.get_virtual_price();
+        totalsBuffer.timestamps[0] = block.timestamp;
+    }
+
+    function bufferSize() public virtual pure returns (uint16) {
+        return 365;
```
vanruch, line 49:
```suggestion
        return MAX_BUFFER_SIZE;
```

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -101,57 +101,58 @@ contract CrvBaseRateOracle {
...
-     * from collected data from last 7 days
-     * @return Calculated weekly profit
+     * @dev What percent of average rate is difference of current rate and average rate.
+     * Based on average rate from collected data in the nearest past.
+     * @param time Determines from how far in the past collected data should be used.
+     * @return Calculated percentage of estimated apy.
      */
-    function weeklyProfit() public view returns (int256) {
-        int256 avgRate = int256(calculateAverageRate(7 days));
+    function apy(uint256 time) internal view returns (int256) {
```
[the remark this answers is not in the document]
vanruch: Agree that this should be APR

contracts/truefi2/CrvBaseRateOracle.sol
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @dev Update the totalsBuffer:
+     * Gets current virtual price from Curve and writes down
+     * new total running value.
+     * If the buffer is filled overwrites the oldest value
+     * with a new one and updates its timestamp.
+     */
+    function updateBuffer() public offCooldown {
```
vanruch, line 74: Maybe call it simply `update()`?

yuchenlintt (changes requested):
My apologies for ignoring this PR for so long. We put this on the backburner, and then I lost track of the long history of changes. I just looked this over again from scratch and think the new streaming sum approach looks fantastic. It's a really elegant solution to the major issue we had with calculating the entire integral from scratch each time.

I want to push back against two of @BenjaminCosman's proposals, and I hope I can make a compelling case for these changes:
1. I think we should calculate a left Riemann sum instead of the trapezoid rule.
2. I don't think we should include the current spot rate in `calculateAverageRate()`.

---

1. Left Riemann sum vs trapezoid rule

I know trapezoid rule is theoretically more accurate since it doesn't lag the way a left Riemann sum does, but I don't think the accuracy benefit outweighs the complexity cost for us:
- SafeMath is not fun to read for a code reviewer or auditor, so even a relatively simple formula like `runningTotals[i+1] = runningTotals[i] + (rate + lastRate) * dt / 2` can be more cumbersome and error-prone to review than `runningTotals[i+1] = runningTotals[i] + rate * dt`
- we have to track an additional `lastRate` state variable and make sure it's initialized, updated, and used correctly
- every update requires more gas from a state read + write and additional arithmetic

We only care about accuracy insofar as borrowers might take advantage of our rates when they're rising (or lenders take advantage of our rates when they're falling). But when we have the granularity of a day, this accuracy shouldn't matter that much over a time-span of a month or more (our target use case). Both sums will converge to the same value, within our error tolerance.

---

2. Not including current spot rate

Two reasons:
a) accuracy vs complexity

Same as above for left Riemann sum vs trapezoid rule. We're duplicating logic, increasing gas spend (for the user this time!), and complicating what should be a simple `(runningTotals[lastIndex] - runningTotals[startIndex]) / totalTime` calculation by adding an external function call.

b) gaming rates

With a cooldown, we prevent gaming rates (e.g., by calling a hundred times in a single hour when rates are especially high or low). This leaves a tiny window in which an attacker could game rates: between each day's cooldown expiry and our keeper bot's function call, which can be on the exact minute if we really cared about it. However, letting users call the spot rate during the average rate calculation would let a malicious caller snipe a transient single-block aberration to screw up the average (well, probably more like 1/30th of it). We get more stability by having less control over when the rate is called.

hal909:
stalled

barrutko:
We're probably gonna take a different approach.

PR #638, opened 2021-05-11
Phase 2 audit requested changes

MiksuJak:
**This PR addresses listed CertiK audit pre-report issues. Each commit represents a single issue with its ID, in order to ease picking changes if necessary.**

1. GAA-01 - not obligatory, but increases code quality by a bit.
2. STT-01 - good catch. Since we already use `SafeERC20` in other contracts, it might be good to use it everywhere for consistency.
3. STT-02 - explicitly setting variables visibility specifiers.
4. STT-03 [SKIPPED] - declaration of a new variable and return would add some extra code. This is a fairly simple function and it should remain compact.
5. STT-04 [SKIPPED] - STT-01 already solves the problem
6. LT2-01 - for consistency all constants should be in upper snake case notation.
7. LT2-02 - comment might have been a little misleading. Dividing by `(365 days / precision)` does not equal dividing by `365 days` and then `precision.`
8. LT2-03 - setting `apy` precision as constant, increases project consistency and code readability.
9. LT2-04 [SKIPPED] - `interest` function calculates loan interest that will be paid after the term passes. However with `value` method, we want to calculate loan’s value at a given time (meaning calculate interest only for a time that has passed).
10. PFY-01 - ensuring that `trueLender` in `PoolFactory` cannot be set to 0 address.
11. TFP-01 [SKIPPED] - might need future investigation if a hard fork of ethereum influences `tx.origin` in an undesirable way. For now it remains a good method to prevent join-exit _flash loan_ attacks.
12. TFP-02 - minor gas saving by allocating removing allocation if check has not passed.
13. TFP-03 - similarly to LT2-03 but for `TrueFiPool2,` increase project consistency and code readability by adding frequently use constants.
14. TFP-04 - similarly to PFY-01, ensuring that beneficiary address cannot be set to 0.
15. TFP-05 [SKIPPED][TO BE CLARIFIED]
16. TFP-06 [SKIPPED] - there might be a case when we want to set oracle's address to 0.
17. TL2-01 [SKIPPED] - present functionality might actually come handy. Currently setting `maxLoans` for less than previously, does not break anything. In fact it ensures that once running loans number decreases, new limit would apply.
18. TL2-02 [OUTDATED] - this issue was resolved while the audit was being conducted.
19. TL2-03 - explicitly setting variables visibility specifiers.
20. TL2-04 - `loans` not changed being a compact function. `redeemAndRepay` and `swapFee` changed to have explicitly named `return`. 
21. CYS-01 - explicitly setting constants visibility specifiers.

### DIFF
--- a/contracts/governance/GovernorAlpha.sol
+++ b/contracts/governance/GovernorAlpha.sol
@@ -358,7 +358,7 @@ contract GovernorAlpha is UpgradeableClaimable {
         require(state(proposalId) == ProposalState.Active, "GovernorAlpha::_castVote: voting is closed");
         Proposal storage proposal = proposals[proposalId];
         Receipt storage receipt = proposal.receipts[voter];
-        require(receipt.hasVoted == false, "GovernorAlpha::_castVote: voter already voted");
+        require(!receipt.hasVoted, "GovernorAlpha::_castVote: voter already voted");
         uint96 votes = countVotes(voter, proposal.startBlock);
 
         if (support) {
--- a/contracts/governance/StkTruToken.sol
+++ b/contracts/governance/StkTruToken.sol
@@ -24,8 +24,8 @@ import {IPauseableContract} from "../governance/interface/IPauseableContract.sol
 contract StkTruToken is VoteToken, StkClaimableContract, IPauseableContract, ReentrancyGuard {
     using SafeMath for uint256;
     using SafeERC20 for IERC20;
-    uint256 constant PRECISION = 1e30;
-    uint256 constant MIN_DISTRIBUTED_AMOUNT = 100e8;
+    uint256 public constant PRECISION = 1e30;
+    uint256 public constant MIN_DISTRIBUTED_AMOUNT = 100e8;
 
     struct FarmRewards {
         // track overall cumulative rewards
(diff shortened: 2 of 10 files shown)


vanruch (approved):

hal909 (approved):

yuchenlintt:
This PR looks great to me! I'm approving this but will let Hal merge.

Thanks for organizing the commits to exactly map to the prelim report issues by ID. I checked against each issue and verified that your commits fix each one (except where you documented exceptions or need for clarification).

Some discussion follows:

STT-03 and TL2-04 `loans`: I generally agree with CertiK's opinion that named return values reduce readability. But that's just my personal style preference, and in this case I think matching the style of the rest of the codebase (i.e., using named return values for compact functions) should take precedence.

LT2-04: Maybe CertiK's confusion could be cleared up with some more explicit naming? I'm not sure what would be best here, but some options are:
- rename `value(_balance)` to `presentValue(_amount)` or `currentValue(_amount)`
- rename `interest(amount)` to `fullValue(_amount)` or `termValue(_amount)`. I believe `interest` typically doesn't include the principal, which could be confusing.
- rename the line 268 `interest` variable to `passedInterest` or `currentInterest`
- factor the duplicated interest calculation out to a single private function `_interest(_amount, passedTime)` that both functions call
We can discuss and follow up on this in a separate PR.

TFP-06: If we want to allow setting `oracle` to `address(0)`, then this should be documented at `oracle`'s declaration. Note that line 535 calls `oracle.truToToken()` without first checking whether `oracle` is 0.

PR #655, opened 2021-05-19
Introduce Pauser contract

MiksuJak:
This is actually really similar to our already developed Governance contract.
It is meant to allow for pausing owned by Timelock contracts, when such a pause request accumulates at least 50M TRU/stkTRU votes.

### DIFF
--- a/contracts/governance/Pauser.sol
+++ b/contracts/governance/Pauser.sol
@@ -22,6 +22,9 @@ contract Pauser is UpgradeableClaimable {
     // @notice The address of the TrustToken Protocol Timelock
     ITimelock public timelock;
 
+    // @notice The address of the TrustToken Protocol Governor
+    IOwnedUpgradeabilityProxy public governor;
+
     // @notice The address of the TrustToken governance token
     IVoteToken public trustToken;
 
@@ -118,12 +121,14 @@ contract Pauser is UpgradeableClaimable {
      */
     function initialize(
         ITimelock _timelock,
+        IOwnedUpgradeabilityProxy _governor,
         IVoteToken _trustToken,
         IVoteToken _stkTRU,
         uint256 _votingPeriod
     ) external {
         UpgradeableClaimable.initialize(msg.sender);
         timelock = _timelock;
+        governor = _governor;
         trustToken = _trustToken;
         stkTRU = _stkTRU;
         votingPeriod = _votingPeriod;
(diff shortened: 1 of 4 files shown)


contracts/governance/Pauser.sol
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        uint256 startBlock = block.number;
+        uint256 endTime = add256(block.timestamp, votingPeriod);
+
+        requestCount++;
+        PauseRequest memory newRequest = PauseRequest({
+            id: requestCount,
+            requester: msg.sender,
+            eta: EXECUTION_PERIOD,
```
BenjaminCosman, line 189: If every request is going to have the same "eta", then I would remove it from this struct and just use the constant `EXECUTION_PERIOD` in the place eta is currently used; it will make it easier to understand what's going on (and also save some gas on storage).  (also "eta" is an inconsistent name for this concept, since in GovernorAlpha it represented a timestamp while here it represents a time duration).

contracts/governance/Pauser.sol
```diff
@@ -0,0 +1,316 @@
...
+
+    // @notice The latest request for each requester
+    mapping(address => uint256) public latestRequestIds;
+
+    // ======= STORAGE DECLARATION END ============
+
+    // @notice The name of this contract
+    string public constant name = "TrueFi Pauser";
+
+    uint256 public constant EXECUTION_PERIOD = 1 days;
```
BenjaminCosman, line 48: This constant is important; add a comment on what it means

contracts/governance/Pauser.sol
```diff
@@ -0,0 +1,316 @@
...
+        uint256 eta;
+        // @notice the ordered list of target addresses of contracts to be paused
+        address[] targets;
+        // @notice The ordered list of functions to be called
+        // different types of proxies might require different types of pause functions
+        PausingMethod[] methods;
+        // @notice The block number at which voting begins: holders must delegate their votes prior to this block
+        uint256 startBlock;
+        // @notice The timestamp at which voting ends: votes must be cast prior to this timestamp
+        uint256 endTime;
```
BenjaminCosman, line 65: I notice that here we use an end *time* while in the very similar GovernorAlpha we use an end *block*. I'm not sure which is better, but we should probably make up our minds and pick one?

BenjaminCosman (changes requested):

hal909 (approved):

PR #654, opened 2021-05-19
TrueFi 3.0

hal909:

PR #659, opened 2021-05-20
Merge 20210519 master into dev

yuchenlintt:
In yesterday's PR #654 we rebased a copy of `dev` on top of `master` and then merged this `dev-merge` branch into `master`. The end result was a `master` that incorporated all of the changes in `dev` on top of the minor fixes to `master` (README, dependabot, etc.) that had accumulated over the past few months.

However, since rebasing creates a new commit hash for each original commit hash, we could no longer merge or rebase `master` or `dev` into each other after this rebase. If we tried, there would be over a hundred duplicate commits with different hashes. Our commit and PR history, which references specific commit hashes, would have become mangled.

(This is largely my bad. I had assured Hal this approach could work, but forgot the consequences of rebasing an actively worked-on branch.)

---

One possible way to solve this would be to abandon the now outdated `dev` and branch a new `dev` off of `master`. But this would require cherry-picking commits that have been added to `dev` since the `master` push, as well as fixing all the orphaned feature branches that came off the current `dev` and its commits.

So I don't think keeping the new `master` commits and abandoning the current `dev` is a good idea.

---

With this PR I'm proposing a new fix to the issue:
1. Rewind `master` to before we merged PR #654 using a force push. For reference, this would currently look like branch `20210519-master`.
2. Merge `20210519-master` into a new branch `master-to-dev` and resolve any merge conflicts. (This PR.)
3. Merge `master-to-dev` into `dev`, which should be a fast-forward. (This PR.)
4. Do any last-day changes to `dev` we want to make.
5. When we're ready, merge `dev` into `master`, which should be a fast-forward.

This would ensure that `master` and `dev` become completely synced, in a safe way that only uses revertible merges (apart from the `master` rewind to `20210519-master`, which we'd do to scrub the bad rebase hashes from master's history).

How does this proposal sound?


yuchenlintt:
I'm closing for now since this isn't a priority. Will reopen (or consider other approaches) once we have more bandwidth next week.

PR #661, opened 2021-05-20
truefi 3.0 deploy scripts & flattened files

hal909:


yuchenlintt (approved):
If we'd like to migrate all deployments to Mars, I propose we merge this PR exactly as is to ensure this deployment is in the commit record, and then submit a follow-up PR to delete `scripts/deploy_truefi2.ts`.

PR #666, opened 2021-05-21
Add mars script for strategy deployment

vanruch:
Tested setting strategy, flushing it and selling CRV on mainnet fork and it works with USDC pool

### DIFF
--- a/deploy/strategy.ts
+++ b/deploy/strategy.ts
@@ -1,0 +1,21 @@
+import { contract, createProxy, deploy, runIf } from 'ethereum-mars'
+import { CrvPriceOracle, CurveYearnStrategy, OwnedUpgradeabilityProxy, PoolFactory } from '../build/artifacts'
+
+const USDC = '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
+const GAUGE = '0xfa712ee4788c042e2b7bb55e6cb8ec569c4530c1'
+const CURVE_POOL = '0xbbc81d23ea2c3ec7e56d39296f0cbb648873a5d3'
+const MINTER = '0xd061d61a4d941c39e5453435b6345dc261c2fce0'
+const ONE_INCH = '0x11111112542d85b3ef69ae05771c2dccff4faa26'
+
+deploy({}, () => {
+  const proxy = createProxy(OwnedUpgradeabilityProxy)
+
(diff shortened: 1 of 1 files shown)


yuchenlintt:
Thanks Ivan!

Have you already deployed this to mainnet?
- If so, let's add the deployed address to `deployments.json` in this PR and merge it exactly as is so we have it in the commit record. I'll follow up with a separate review suggestions PR.
- If not, I have a proposed refactor and two minor style edits:
  1. Let's move this code under `deploy/truefi2.ts`. I'll be working on some changes today to fix the deploy script after yesterday's hacks, so we can add this once those changes are merged.
  2. To make it easier to separate impl deployments from proxy upgrades, let's split `proxy(contract(CurveYearnStrategy), () => {})` into separate lines: `const curveYearnStrategy_impl` and `proxy(curveYearnStrategy_impl, () => {})`. Thanks for separating the `initialize()` call into its own `runIf()`-guarded line!
  3. Let's adhere to a naming convention in deploy scripts. My proposal:
    - a generic contract: `const contractName = contract(ContractName)`
    - an impl behind a proxy: `const contractName_impl = contract(ContractName)`
    - a proxy: `const contractName = proxy(contractName_impl, () => {})`
    - a specialization of a contract with different versions: `const usdc_ContractName = contract('usdc_ContractName', ContractName)`

Thanks again! Let me know what you think.

hal909 (approved):

vanruch:
@yuchenlintt No, I didn't deploy it yet. Tbh, I'd rather not create variables holding implementations to avoid accidentally sending transactions to it, but I don't hold any strong opinion on this topic

yuchenlintt:
> @yuchenlintt No, I didn't deploy it yet. Tbh, I'd rather not create variables holding implementations to avoid accidentally sending transactions to it, but I don't hold any strong opinion on this topic

Yeah, I agree that it's not great to create variables that point to implementations. Ideally we would only hold references to contracts that we're allowed to interact with directly. But I also suspect this is a tradeoff we have to make. If we separate impls from their proxies, then it becomes possible to use Mars for mainnet deploys where we don't have script ownership of the multisig address. And it cleanly splits impl deployment from upgrades/initialization, which in practice is how we handle protocol upgrades.

Anyway, now that this is merged let's deploy!!! Hack now, clean up + refactor later :)

PR #675, opened 2021-05-25
Override LoanToken2 decimals with underlying token decimals

skalermo:


contracts/truefi2/LoanToken2.sol
```diff
@@ -137,11 +147,12 @@ contract LoanToken2 is ILoanToken2, ERC20 {
         uint256 _amount,
         uint256 _term,
         uint256 _apy
-    ) public ERC20("TrueFi Loan Token", "LOAN") {
+    ) public {
         require(_lender != address(0), "LoanToken2: Lender is not set");
 
         pool = _pool;
         token = _pool.token();
+        ERC20.__ERC20_initialize(concat("TrueFi ", token.name()), concat("tf", token.symbol()));
```
MiksuJak, line 155: Does it not create some confusion?  In this case loan tokens would have same names and symbols as their corresponding pools.

contracts/truefi/SushiTimelock.sol
```diff
@@ -125,7 +125,7 @@ contract SushiTimelock {
 
         // YYY: // solium-disable-next-line security/no-call-value
         // solhint-disable-next-line avoid-call-value, avoid-low-level-calls
-        (bool success, bytes memory returnData) = target.call.value(value)(callData);
+        (bool success, bytes memory returnData) = target.call{value:value}(callData);
```
yuchenlintt, line 128: This is to fix the `yarn build` warning, right? Let's move this change to a separate PR since it's unrelated to LoanToken2s.

Also since this contract is copied verbatim from Sushiswap without unit testing, let's add a comment to show the diff, e.g.:
```
// YYY: (bool success, bytes memory returnData) = target.call.value(value)(callData);
```
This helps auditors determine what we've changed and what's original.

And finally, a style nit: `target.call{value: value}`.

Thanks!

contracts/truefi2/LoanToken2.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.6.10;
 
-import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {ERC20} from "../common/UpgradeableERC20.sol";
```
yuchenlintt, line 4: Let's move this line down to the local imports section, alphabetically first.

contracts/truefi2/LoanToken2.sol
```diff
@@ -119,6 +119,16 @@ contract LoanToken2 is ILoanToken2, ERC20 {
      */
     event Liquidated(Status status);
 
+    /**
+     * @dev Helper function to concatenate two strings
+     * @param a First part of string to concat
+     * @param b Second part of string to concat
+     * @return Concatenated string of `a` and `b`
+     */
+    function concat(string memory a, string memory b) internal pure returns (string memory) {
```
yuchenlintt, line 128: This concat function doesn't appear to be used. Could you remove it?

yuchenlintt (changes requested):
Thank you for the fix! Looks great overall and thanks for the fast turnaround.

Let's move the build warning fix to another PR, and address some minor comments here. After that this should be good to merge.

hal909 (changes requested):
Implement Yuchen's requested changes and we can check this in

skalermo:
> Thank you for the fix! Looks great overall and thanks for the fast turnaround.
> 
> Let's move the build warning fix to another PR, and address some minor comments here. After that this should be good to merge.

Moved the fix to #678.

yuchenlintt (approved):

yuchenlintt:
Thanks for the fix! Looks good to me.

hal909 (approved):

PR #676, opened 2021-05-25
Deploy strategy and oracle

MiksuJak:


yuchenlintt (approved):
Since this change represents what has been deployed on mainnet, we should merge it as is to preserve it in the commit record. However we generally shouldn't mix code changes and deploys in the same PR. Code changes need review (and ideally, audit), and deploys need to reflect exactly what exists in the codebase.

I'll approve + merge this PR now and follow up with a review.

yuchenlintt (commented):
Actually this looks great as is now that I look at it. No further comments!

PR #680, opened 2021-05-27
Add linear continuous vesting for Vault

vanruch:


contracts/governance/TrueFiVault.sol
```diff
@@ -61,6 +66,17 @@ contract TrueFiVault {
...
 
+    function withdrawable(IERC20 token) public view returns (uint256) {
+        if (beneficiary == owner) {
+            return token.balanceOf(address(this));
+        }
+        uint256 timePassed = block.timestamp.sub(expiry.sub(DURATION));
+        if (timePassed > DURATION) {
+            timePassed = DURATION;
+        }
+        return token.balanceOf(address(this)).add(withdrawn[token]).mul(timePassed).div(DURATION).sub(withdrawn[token]);
```
BenjaminCosman, line 77: This system doesn't correctly enforce linear vesting: for example (ignoring staking rewards for a moment), assume someone started with 100 TRU in this vault. Then after half a year they could withdraw 50 TRU, tell the vault to stake the remaining 50, and then withdraw 25 stkTRU, effectively withdrawing 75% of the principal after only 50% of the time.

BenjaminCosman: Separately, what is the desired behavior as regards staking rewards? e.g. if someone starts with 100 TRU and after half a year has say 130 total thanks to staking rewards, is the desired behavior that they can withdraw 65: half their principal as well as half the rewards?

BenjaminCosman (changes requested):

contracts/governance/TrueFiVault.sol
vanruch: Regarding withdrawing stkTRU - maybe we should just ban withdrawing stkTRU until the vault is opened? 
I think it's fine to allow withdrawing staking rewards

BenjaminCosman:
I don't think that's an acceptable solution, since this means that someone has to unstake before withdrawing, and initiate cooldown two weeks before unstaking, so that's two extra txs and significant planning ahead just to claim what should be theirs.

Potential alternate solution: you could allow withdrawing of both TRU and stkTRU, but do the accounting entirely denominated in TRU - any time they withdraw stkTRU, you compute how much TRU that was worth (in the same way the stkTRU contract does).

contracts/governance/TrueFiVault.sol
```diff
@@ -17,34 +17,40 @@ import {StkTruToken} from "./StkTruToken.sol";
...
     using SafeMath for uint256;
 
+    uint256 public DURATION = 365 days;
+
     address public owner;
     address public beneficiary;
     uint256 public expiry;
+    mapping(IERC20 => uint256) public withdrawn;
+
     IERC20 public tru;
```
hal909, line 30: Can we hard code the contract addresses here instead of initializing?

BenjaminCosman:
> There's no point really to get stkTRU from vault - they are staked anyway. So if someone wants to get their locked TRU, they'll have to initiate cooldown and then unstake them anyway.

stkTRU is tradable, so one doesn't necessarily need to do a cooldown before selling it. (I don't think there's a liquid market for it yet, but there might well be within the year+ we're talking about here, plus people might have other ways to trade it.)

contracts/governance/TrueFiVault.sol
```diff
@@ -17,34 +18,40 @@ import {StkTruToken} from "./StkTruToken.sol";
...
+    ) external initializer {
         owner = msg.sender;
         beneficiary = _beneficiary;
-        expiry = block.timestamp.add(_duration);
+        expiry = block.timestamp.add(DURATION);
         tru = _tru;
         stkTru = _stkTru;
 
-        require(tru.transferFrom(owner, address(this), _amount), "TrueFiVault: insufficient owner balance");
         stkTru.delegate(beneficiary);
```
BenjaminCosman, line 47: Perhaps call delegate on `tru` as well?

BenjaminCosman (changes requested):
I notice that `claimRewards` only claims pending TRU; there should be a way to claim fees (USDC) as well.

hal909 (approved):

PR #685, opened 2021-06-04
Fix Rating Agency decimal calculations

MiksuJak:


vanruch (approved):

contracts/truefi2/mocks/MockUsdc.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.6.10;
+
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract MockUsdc is ERC20 {
```
yuchenlintt, line 6: We already have a TestUSDCToken deployed to Ropsten in `contracts/trusttoken/mocks/`. Can we reuse that code here, or do you think this mock should be maintained separately?

contracts/truefi/TrueRatingAgencyV2.sol
```diff
@@ -379,15 +380,13 @@ contract TrueRatingAgencyV2 is ITrueRatingAgencyV2, Ownable {
...
-
+            uint256 interest = ILoanToken2(id).profit();
             // calculate reward
             // prettier-ignore
             uint256 totalReward = toTRU(
-                interest
-                    .mul(distributor.remaining())
-                    .mul(rewardMultiplier)
-                    .div(distributor.amount())
+                interest.mul(distributor.remaining()).mul(rewardMultiplier).mul(10**18).div(distributor.amount()).div(
```
yuchenlintt, line 387: Just double-checking for sanity here: is it okay that we can't call `decimals()` on LoanToken?

I think the original LoanToken imports ERC20 directly from OpenZeppelin, whereas LoanToken2 imports ERC20 from our UpgradeableERC20. I don't believe OZ's ERC20 has `decimals()` implemented. If we upgrade this TrueRatingAgencyV2 contract, wouldn't we be unable to interact with existing (or new legacy) LoanTokens?

contracts/truefi2/mocks/MockUsdc.sol
MiksuJak: Sure I can use that one. I didn't really realise it is there

contracts/truefi/TrueRatingAgencyV2.sol
MiksuJak: OZ's ERC20 actually does have decimals implemented and they it returns 18 as default.  Actually if You look at the tests, they run on both old LoanToken and the new LoanToken2 and they seem to pass just fine. Then again, Loans that are in review state right now are created by new LoanFactory2 so they are LoanTokens2 as well. I will not resolve this conversation if any further discussion is to appear. Anyway, I'm pretty confident that it works with both LoanToken and LoanToken2

contracts/truefi2/deprecated/LoanToken2Deprecated.sol
```diff
@@ -0,0 +1,471 @@
+// SPDX-License-Identifier: MIT
```
vanruch, line 1: Can we have this inherit from LoanToken2 and override decimals?

contracts/truefi/TrueRatingAgencyV2.sol
yuchenlintt: My bad, you're right! I just confirmed that OZ's ERC20 returns 18 by default. You also raise a good point that new loans should all be LoanToken2s because they come from LoanFactory2. Thanks for looking into this.

contracts/truefi2/deprecated/LoanToken2Deprecated.sol
MiksuJak: I wanted it to match exactly on chain version but tbh it might just require to write some integration tests

PR #690, opened 2021-06-09
vault deploy script

hal909:


contracts/governance/TrueFiVault.sol
```diff
@@ -38,6 +36,7 @@ contract TrueFiVault is Initializable {
 
     function initialize(
         address _beneficiary,
+        address _finalOwner,
```
vanruch, line 39: would be unfortunate if we provide both wrong beneficiary and final owner, but ok, we'll just have to double check

scripts/truefi_vault_deploy.ts
```diff
@@ -5,22 +5,24 @@ import { ethers, providers } from 'ethers'
...
   TrueFiVault__factory,
-} from 'contracts'
+} from '../build'
 
 // inputs
-const beneficiary = ''
-const amount = '10000000000'
-const start = ''
-const txnArgs = { gasLimit: 1_000_000, gasPrice: 60_000_000_000 }
+const beneficiary = '0x58f5F0684C381fCFC203D77B2BbA468eBb29B098' // blocktower
```
vanruch, line 13: let's not leave investors' names in the comments

vanruch (approved):

contracts/governance/TrueFiVault.sol
hal909: yes, we will just double check. If we make a mistake we have to upgrade the smart contract, this is intentional

yuchenlintt: It looks like `owner` is no longer used anywhere in this contract. Can we remove it?

If we want to support future `owner` authorization, perhaps we should inherit UpgradeableClaimable instead of Initializable. That seems to be our canonical way of expressing contract ownership.

PR #694, opened 2021-06-09
script to set credit scores

hal909:


yuchenlintt (approved):
If this is the deploy script that was run, let's merge it as is. I've reinitiated the failing integration tests, so once those pass you can merge.

Medium term, can we set up a new 1-of-n gnosis safe for admin actions unrelated to contract upgrades? If I understand our current process correctly, our credit team manually updates a spreadsheet, and then Hal reads the score from the spreadsheet and updates the contract state. If our threat model considers our credit team/spreadsheet to be untrustworthy, then our current process is still vulnerable even if Hal is the one updating score. If our threat model trusts our credit team/spreadsheet, then we ought to grant them direct setScore() privileges instead of bottlenecking through Hal.

I believe this won't even require code changes, just an UpgradeableClaimable transferOwnership() call. If we keep *proxy* ownership under our ownership multisig but grant *impl* ownership to a credit team multisig, then our credit team would be able to call setScore. Thoughts on this proposal?

PR #705, opened 2021-06-11
Fix slither detectors

skalermo:
For now I just made slither ignore all found issues except those related to reentrancy attack.

A few thoughts on the matter:
- I think we could safely exclude [incorrect-equality](https://github.com/crytic/slither/wiki/Detector-Documentation#dangerous-strict-equalities) and [boolean-cst](https://github.com/crytic/slither/wiki/Detector-Documentation#misuse-of-a-boolean-constant) from used slither detectors. 
- Maybe we should consider using `throw` or `return` to handle potential failure of functions like [approve](https://github.com/trusttoken/smart-contracts/blob/3a0225c203839f939bc9d2247906a599731c5b93/contracts/truefi/TrueLender.sol#L191), [queueTransaction](https://github.com/trusttoken/smart-contracts/blob/3a0225c203839f939bc9d2247906a599731c5b93/contracts/governance/GovernorAlpha.sol#L243), [executeTransaction](https://github.com/trusttoken/smart-contracts/blob/3a0225c203839f939bc9d2247906a599731c5b93/contracts/governance/GovernorAlpha.sol#L257) etc.


hal909 (changes requested):
Let's fix the re-entrancy Issue.

I feel like we shouldn't remove incorrect inequalities or boolean constants unless there is a very compelling reason to do so. It would be best to make our code as slither-compliant as possible Unless @vanruch or @barrutko  thinks otherwise

For throw/catch, I agree we should use them - this will let us have better handling of failed transactions I think. We need to make sure we get the correct error message though (for example, if a transaction fails, we don't want to display "Governor Execution Failed" but rather something like "SafeMath: subtraction overflow"

hal909 (approved):

PR #707, opened 2021-06-11
Get LoanTokens from the pool

vanruch:


contracts/truefi/TrueFiPool.sol
```diff
@@ -75,6 +75,8 @@ contract TrueFiPool is ITrueFiPool, IPauseableContract, ERC20, ReentrancyGuard,
 
     ITrueLender2 public _lender2;
 
+    address public safu;
```
barrutko, line 78: Why we need safu pointer here if we are not about to use it?

contracts/truefi2/PoolFactory.sol
```diff
@@ -120,7 +130,7 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
         pool[token] = address(proxy);
         isPool[address(proxy)] = true;
 
-        ITrueFiPool2(address(proxy)).initialize(ERC20(token), liquidationToken, trueLender2, ONE_INCH_ADDRESS, this.owner());
+        ITrueFiPool2(address(proxy)).initialize(ERC20(token), liquidationToken, trueLender2, ONE_INCH_ADDRESS, safu, this.owner());
```
barrutko, line 133: If someone creates a new pool before we set safu in this factory, safu will be set to 0. Maybe let's require it not to be 0?

contracts/truefi/TrueFiPool.sol
vanruch: We should add liquidate function same as to Pool2

contracts/truefi2/PoolFactory.sol
vanruch: We have a setter for this case, it's fine

contracts/truefi/TrueFiPool.sol
vanruch: Didn't have time to do this yet

contracts/truefi2/PoolFactory.sol
barrutko: This is fine 🔥🐶🔥

contracts/truefi2/SAFU.sol
```diff
@@ -25,6 +25,8 @@ contract SAFU is UpgradeableClaimable {
         require(loanFactory.isLoanToken(address(loan)), "SAFU: Unknown loan");
         require(loan.status() == ILoanToken2.Status.Defaulted, "SAFU: Loan is not defaulted");
         ITrueFiPool2 pool = ITrueFiPool2(loan.pool());
-        IERC20(pool.token()).safeTransfer(address(pool), loan.debt());
+        pool.liquidate(loan);
+        uint256 lostByPool = loan.debt().mul(loan.balanceOf(address(this))).div(loan.totalSupply());
```
barrutko, line 29: `poolLoss` ?

contracts/truefi2/SAFU.sol
```diff
@@ -25,6 +25,8 @@ contract SAFU is UpgradeableClaimable {
...
+        IERC20(pool.token()).safeTransfer(address(pool), lostByPool);
```
MiksuJak, line 30: I think pool should be a little more distrustful towards SAFU there. Maybe we could approve pool for `lostByPool` amount of tokens and then `pool.liquidate(loan)` would handle `transferFrom`?

MiksuJak, line 30: `pool.liquidate` will transfer loan tokens to the SAFU contract. Maybe we could approve pool by SAFU for `lostByPool` amount of token, then `pool.liquidate` would make a `transferFrom` and if it passes the pool would transfer loan tokens to SAFU. This way pools will be a little more distrustful towards already centralized SAFU.

contracts/truefi2/TrueLender2.sol
```diff
@@ -399,6 +399,16 @@ contract TrueLender2 is ITrueLender2, UpgradeableClaimable {
         _distribute(recipient, numerator, denominator, msg.sender);
     }
 
+    /**
+     * @dev Allow pool to transfer all LoanTokens to the SAFU in case of liquidation
+     * @param loan LoanToken address
+     * @param recipient expected to be SAFU address
+     */
+    function transferAllLoanTokens(ILoanToken2 loan, address recipient) public override {
```
MiksuJak, line 407: Should this be public? It seems to me, that it could be called at any time, by anyone

contracts/truefi2/PoolFactory.sol
MiksuJak: I think safu should stay out of initializer and should be set by setter. Not all of the pools will be guarded by the assurance fund

contracts/truefi2/TrueLender2.sol
vanruch: there's a require below

vanruch: should be changed to external btw

MiksuJak: Yes there is, but require only checks if pool is correct, does not check if the loan has defaulted

contracts/truefi2/SAFU.sol
vanruch: not a bad idea, but we would still need to pass compensation amount to the pool. What if it's 0? This change will not give us any additional protection, i think

MiksuJak (approved):

PR #711, opened 2021-06-15
Ignore frequently failing integration tests on CI

skalermo:


hal909 (changes requested):
Do we really want this? What if we change code in a way that breaks one of these critical tests?

hal909 (changes requested):
I'm going to close this unless anyone has an objection

yuchenlintt:
Thanks Roman for bringing this up and proposing a solution!

I agree that flaky tests are a problem for dev workflow, although I don't think this PR is the best way to fix them so I'm closing it.

I've been finding myself mentally ignoring + manually rerunning tests when I see `sell TRU on 1inch`, etc. It's becoming a bit of a "boy who cried wolf" problem. If I don't trust these test results anyway, then will I pay attention when there's a real failure? And should we halt our development every time there's market volatility?

I don't know of any perfect fixes. Some compromise options:
1. Keep running integration tests on every PR, but make passing optional. Then we'll slowly start ignoring failures, and real regressions might emerge.
2. Refactor the tests so they don't call the actual 1inch API. But then these aren't real integration tests any more, are they?
3. Decouple integration tests from individual PRs. Looks like CircleCI supports [scheduling workflows](https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow), e.g., we could run integration tests on `dev` nightly at 0000 UTC.
4. Find a way to temporarily quarantine flaky tests. Is there a way where we could halt them in CircleCI, and then they automatically start running again after a day or two?

I think (3) is the least bad option. Let me know if you have any other ideas though.

PR #714, opened 2021-06-18
Remove build dependency from lint and slither job

skalermo:


yuchenlintt (approved):
Thanks for the change. Hopefully this helps our CI run a little more efficiently. lint + slither weren't our real bottlenecks, but at least we'll reduce congestion a little when starting the post-build tests.

PR #716, opened 2021-06-21
Allow single file contract verification with TrueFi logo

vanruch:
There's a strange problem with singlefile verification - the bytecode of flat contracts is different from the compiled ones.


yuchenlintt (approved):
Thanks for the change! Looks like this is mostly just a version bump, so I'll approve + merge.

The bytecode difference is worrisome. I wonder whether it's the same reason our `yarn flatten` output sometimes fails to verify on Etherscan. Might be a good idea to diff the bytecode from normal compiles vs `yarn flatten` vs mars flattenSingleFile.

vanruch:
@yuchenlintt yarn flatten and mars flattener would be same flatteners and they both may produce different bytecode. The problem is that the compiler doesn't flatten the file, it produces the [standard JSON input](https://docs.soliditylang.org/en/v0.7.5/using-the-compiler.html#compiler-input-and-output-json-description) just like default mars verification.

PR #719, opened 2021-06-21
Make TrueFiVault UpgradeableClaimable

vanruch:
Important: we must not use this as implementation for previously deployed vaults as it will break storage. I've decided not to create multiple versions because it's unlikely we will want to change old vaults' implementation and I wanted to avoid having Vault and Vault2 in the repo


yuchenlintt (approved):
Looks good! I agree, it's very unlikely we will ever upgrade any vault that's already been granted, and I'd prefer to have less clutter/code duplication in the repo.

PR #720, opened 2021-06-21
Feature/reward borrower

jpalmer-trust:


hal909 (changes requested):
We need to create a modifier onlyLender which restricts who can call this function. If it is just external, anyone can set the RewardRate and anyone can call distribute() to steal TRU from the contract.

See: https://docs.soliditylang.org/en/v0.6.10/structure-of-a-contract.html?highlight=modifier#function-modifiers

We also want to add an initialize() function which sets the initial value of the lender address. Initialize can set the uint256 rewardRate to 100 and address lender to msg.sender.

contracts/truefi2/BorrowerRewardDistributor.sol
```diff
@@ -0,0 +1,47 @@
...
+
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+
+/**
+ * @title BorrowerRewardDistributor
+ * @dev Reward borrowers for taking out loans in TRU
+ */
+
```
hal909, line 12: remove extra line here

contracts/truefi2/BorrowerRewardDistributor.sol
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @dev Reward rate is set as 1% by default. 2 decimals
+     * so the actual reward can be calculated as: loan * rewardRate / 10000
+     */
+    uint256 public rewardRate = 100;
+
+    /** @notice setting TRU address
+     */
+    constructor(IERC20 _rewardCurrency) public {
```
hal909, line 30: add a function initialize(IERC20 _rewardCurrency, address _lender) instead of a constructor. Make initialize set:
rewardCurrency  = _rewardCurrency;
lender = _lender;
rewardRate = 100;

contracts/truefi2/BorrowerRewardDistributor.sol
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+contract BorrowerRewardDistributor {
+    using SafeMath for uint256;
+    using SafeERC20 for IERC20;
+
+    /**
+     * @notice Will be TRU
+     */
+    IERC20 public rewardCurrency;
```
hal909, line 20: Add variable:
address lender;

contracts/truefi2/BorrowerRewardDistributor.sol
```diff
@@ -0,0 +1,47 @@
...
+     */
+    uint256 public rewardRate = 100;
+
+    /** @notice setting TRU address
+     */
+    constructor(IERC20 _rewardCurrency) public {
+        rewardCurrency = _rewardCurrency;
+    }
+
+    function setRewardRate(uint256 _rewardRate) external {
```
hal909, line 34: add documentation

contracts/truefi2/BorrowerRewardDistributor.sol
```diff
@@ -0,0 +1,47 @@
...
+        require(_rewardRate < 10000, "Reward rate should be less than 10000(100%)");
+        rewardRate = _rewardRate;
+    }
+
+    function distribute(address borrower, uint256 loanAmount) external {
```
hal909, line 39: add documentation

yuchenlintt:
My apologies, I should've reposted my comments here to prevent duplicate review work from Hal.

My PMed comments below:

---

some comments:
- let's move `BorrowerRewardDistributor` under a `truefi2/distributors` subdir (see `truefi/`), and rename it to `BorrowerDistributor` (all our distributors are for rewards, so no need to make it more verbose)
- make it `UpgradeableClaimable` so we can authorize contract param changes using an `owner` address and initialize the contract safely
- replace the constructor with an `initialize()` function. this lets us deploy it behind an `OwnedUpgradeabilityProxy` (must be deployed with empty constructor), which means we can upgrade implementations while keeping the same contract address
- since upgrades can mangle contract state, add a comment block for state vars (see `LinearTrueDistributor` for an example)
- we'll have to convert between loan decimals (18 for TUSD, 6 for USDC + USDT -- should be determined by `LoanToken2.decimals()`) and TRU decimals (8 -- we'd have to make it an `IERC20WithDecimals` to call `decimals()` ) to calculate rewards properly
- i'm not sure we need to require `rewardRate` to be less than 10_000. depending on the TRU/loan exchange rate, we could conceivably grant > 100% TRU as reward
- `safeTransfer` already does a balance check and reverts on insufficient funds, so we won't need to check it ourselves. i'm not sure yet whether we want to revert on failure, but if we don't need to we can always add the check back in the future

contracts/truefi2/distributors/BorrowerDistributor.sol
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * @notice Only lenders can adjust the reward rate and distribute the reward
+     */
+    address public lender;
+
+    /**
+     * @dev Reward rate is set as 1% by default. 2 decimals
+     * so the actual reward can be calculated as: loan * rewardRate / 10000
+     */
+    uint256 public rewardRate = 100;
```
hal909, line 37: set this value in the initialize() function

contracts/truefi2/distributors/BorrowerDistributor.sol
```diff
@@ -0,0 +1,64 @@
...
+
```
hal909, line 38: Add a line below indicating storage declaration is over:

// ======= STORAGE DECLARATION END ============

contracts/truefi2/distributors/BorrowerDistributor.sol
```diff
@@ -0,0 +1,64 @@
...
+    /** @notice setting reward rate
+     */
+    function setRewardRate(uint256 _rewardRate) external onlyLender {
+        rewardRate = _rewardRate;
+    }
+
+    /** @notice Distribute reward
+     */
+    function distribute(address borrower, uint256 loanAmount) external onlyLender {
+        uint256 rewardAmount = loanAmount.mul(rewardRate).div(10000);
```
hal909, line 61: Store 10000 as a constant: BASIS_RATIO

contracts/truefi2/distributors/BorrowerDistributor.sol
```diff
@@ -0,0 +1,64 @@
...
+    /** @notice setting TRU address
+     */
+    function initialize(IERC20 _rewardCurrency, address _lender) external initializer {
+        rewardCurrency = _rewardCurrency;
+        lender = _lender;
+    }
+
+    /** @notice setting reward rate
+     */
+    function setRewardRate(uint256 _rewardRate) external onlyLender {
```
hal909, line 54: Make this onlyOwner instead of onlyLender

contracts/truefi2/distributors/BorrowerDistributor.sol
```diff
@@ -0,0 +1,64 @@
...
+        rewardRate = _rewardRate;
+    }
+
+    /** @notice Distribute reward
+     */
+    function distribute(address borrower, uint256 loanAmount) external onlyLender {
```
yuchenlintt, line 60: We also need the loan decimals, otherwise we aren't properly scaling rewards for different loan currencies. I would pass in an ILoanToken2 and set `rewardAmount = loan.amount() * rewardCurrency.decimals() / loan.decimals() * rewardRate / BASIS_RATIO` (translated to SafeMath, of course). Maybe split this calculation into two lines for legibility?

You'll need to import `IERC20WithDecimals` from `contracts/truefi2/interface/' instead of OpenZeppelin's IERC20.

contracts/truefi2/distributors/BorrowerDistributor.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @notice Will be TRU
+     */
+    IERC20 public rewardCurrency;
+
+    /**
+     * @notice Only lenders can adjust the reward rate and distribute the reward
+     */
+    address public lender;
```
yuchenlintt, line 31: Let's import ITrueLender2 and use that instead of an arbitrary address.

contracts/truefi2/distributors/BorrowerDistributor.sol
yuchenlintt: Also if we're passing in a LoanToken2 then that should already include the borrower address `loan.borrower()`. So no need to pass it in as a separate arg.

yuchenlintt (changes requested):
Thanks for the changes so far!

Let me know if you run into any roadblocks or want to discuss our comments.

yuchenlintt (approved):

yuchenlintt:
@jpalmer-trust I think you can go ahead and add some unit tests for this.

We'll work on creating an interface + integrating this into TrueLender2 in separate PRs.

marekkirejczyk:
Looks stalled.

hal909:
stalled

PR #721, opened 2021-06-22
Amnesic prompt for deploy private key

yuchenlintt:
I got tired of clearing my shell history after every deploy, so I looked up some basic shell hygiene for secrets.

Running a mars deploy script works just like before, e.g.,
```
$ yarn deploy:truefi --network ropsten --dry-run
```

Now, instead of setting the PRIVATE_KEY environment variable manually, the script prompts for the secret without echoing it to the command line output or recording it in history. Moreover once the script exits, its environment variables should all be cleared.

This lets us deploy with Mars *while screensharing* without leaking secrets.

WARNING The PRIVATE_KEY env var still gets leaked to the process table while the script is running. You can run `ps -E` to see all environment variables. No easy way I can see around this; we might have to route signatures through a wallet like Geth.


vanruch (approved):
That's pretty cool

PR #728, opened 2021-06-24
Allow tether flushing

MiksuJak:


yuchenlintt (changes requested):
Wow, that's a surprisingly simple fix! Kudos to you, Ivan and Mikolaj, for finding a solution.

I also spent several hours trying to figure out what broke, but I'm still not sure I understand it. Why does safeApprove() work while approve() doesn't? As I understand it, approve() doesn't even revert when it fails, it just returns false. Yet our failed transactions are reverting on the approve() call.

For the PR, just a couple style notes and you're good to go.

yuchenlintt (approved):

PR #730, opened 2021-06-25
Fix tether loan funding

yuchenlintt:
USDT's approve() function doesn't comply with the ERC20 spec since it reverts on failure and doesn't return a boolean for success/failure.

TrueLender2 currently calls the low-level ERC20 approve() instead of SafeERC20's safeApprove(), so it fails due to ABI incompatibility with USDT.

---

This PR adds a failing integration test to demonstrate the bug, and then fixes the test by globally replacing ERC20 approve() calls with SafeERC20 safeApprove() calls.

### DIFF
--- a/contracts/governance/TrueFiVault.sol
+++ b/contracts/governance/TrueFiVault.sol
@@ -119,7 +119,7 @@ contract TrueFiVault is UpgradeableClaimable {
      * @param amount Amount of TRU to stake
      */
     function stake(uint256 amount) external onlyBeneficiary {
-        tru.approve(address(stkTru), amount);
+        tru.safeApprove(address(stkTru), amount);
         stkTru.stake(amount);
     }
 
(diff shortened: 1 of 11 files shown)


yuchenlintt:
Success! I figured out what was causing this PR's new integration test to fail: the test itself had a fatal bug. The bug is now fixed and the test passes in commit 1dc5002.

To confirm the original TrueLender2 would have failed the test, I checked out commit a16c120 in detached HEAD mode, cherry-picked 1dc5002, and rebuilt + reran my test. Failure as desired.

---

What went wrong: it turned out to be another decimals bug, this time in the integration test.

I had blindly copied over test code from the other TrueLender2 integration test. The other test called `parseEth(10000)` (i.e., 10_000 TUSD) to determine the TUSD amount (10_000_000_000_000_000_000_000). TUSD happens to match `parseEth`'s 18 decimals, whereas USDT uses 6 decimals. So when joining the lending pool my arbitrary USDT holder address sadly did not have enough funds.

I then mistook the `revert SafeERC20: low-level call failed` error to mean there must be something wrong with the `safeApprove()` call in TrueLender2's `fund()`, and went on a wild-goose chase.

vanruch (approved):

PR #742, opened 2021-07-01
Use SafeERC20

MiksuJak:


yuchenlintt (approved):
Thanks for doing this! Looks great to me.

The possible slither reentrancy looks like a preexisting issue that got uncovered. I think it's fine to merge this PR and then separately investigate + fix that. Could you file an Asana task under Ready for development?

PR #743, opened 2021-07-02
Allow creating private pools

MiksuJak:


contracts/truefi2/PoolFactory.sol
```diff
@@ -50,18 +54,33 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
...
+     */
+    event PrivatePoolCreated(address borrower, address token, address pool);
+
     /**
      * @dev Event to show that token is now allowed/disallowed to have a pool created
      * @param token Address of token
      * @param status New status of allowance
      */
-    event AllowedStatusChanged(address token, bool status);
+    event TokenAllowedStatusChanged(address token, bool status);
```
vanruch, line 70: I think we should never change event names. We have them to have a clear history of what was happening with the contract and changing it could be confusing. On the other hand it sucks to have AllowedStatusChanged and BorrowerAllowedStatus changed. Maybe we could use the whitelist term for borrowers instead of Allowance?

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -103,9 +103,10 @@ contract TrueFiPool2 is ITrueFiPool2, IPauseableContract, ERC20, UpgradeableClai
         ERC20 _token,
         ITrueLender2 _lender,
         ISAFU _safu,
-        address __owner
+        address __owner,
+        string memory name
```
vanruch, line 107: I think argument name "name" is misleading. I understand that the pool will be called something like "TrueFi Alameda USDC". I have no idea what it should be called. Maybe we need to have separate initializers for public and private pools? This one will stay same and private initializer will take full name and symbol?

yuchenlintt (changes requested):
Looks like this is the right way to implement pools for single borrowers.

I'm not sure I like the idea of calling it `private` though. `private` is an overloaded term that can refer to code visibility (in CS), protection from surveillance (in human rights), or business vs state sectors (in government).

Any suggestions for a better name? I'm considering: `singleBorrowerPool`, `personalPool`, `monopsonyPool`.

contracts/truefi2/PoolFactory.sol
```diff
@@ -58,10 +70,17 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
     event AllowedStatusChanged(address token, bool status);
 
     /**
-     * @dev Event to show that allowAll status has been changed
-     * @param status New status of allowAll
+     * @dev Event to show that borrower is now allowed/disallowed to have a private pool
+     * @param borrower Address of borrower
+     * @param status New status of allowance
      */
-    event AllowAllStatusChanged(bool status);
```
barrutko, line 64: Not sure if we want to rename the event.

contracts/truefi2/PoolFactory.sol
```diff
@@ -122,33 +150,62 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
...
      * @param token Address of token which the pool will correspond to.
      */
-    function createPool(address token) external onlyAllowed(token) onlyNotExistingPools(token) {
+    function createPool(address token) external onlyAllowedTokens(token) onlyNotExistingPools(token) {
         OwnedProxyWithReference proxy = new OwnedProxyWithReference(this.owner(), address(poolImplementationReference));
         pool[token] = address(proxy);
         isPool[address(proxy)] = true;
 
-        ITrueFiPool2(address(proxy)).initialize(ERC20(token), trueLender2, safu, this.owner());
+        ITrueFiPool2(address(proxy)).initialize(ERC20(token), trueLender2, safu, this.owner(), "");
```
barrutko, line 158: Why empty string?

barrutko (commented):
Let's keep old naming unchanged, only add new stuff.

barrutko (commented):
Token allowance, borrowers whitelist

yuchenlintt:
Thanks, I think this looks almost ready.

Two remaining questions:
1. Can we come up with a better name than private? I've polled the team in #truefi to see if they have better ideas.
2. Should we restrict single borrower pool creation? Currently we let the public create a pool for any {whitelisted borrower} x {allowed token} combination. Is there a situation where we want to let Alice have a USDC pool but don't want anyone to give her a USDT pool? I'm not sure there is... but if there is such a situation then here are some other options to consider:
  - drop the borrower whitelist and make single borrower pool creation onlyOwner
  - keep the whitelist, but only allow borrowers to create their own single borrower pools

MiksuJak:
> Thanks, I think this looks almost ready.
> 
> Two remaining questions:
> 
> 1. Can we come up with a better name than private? I've polled the team in #truefi to see if they have better ideas.
> 2. Should we restrict single borrower pool creation? Currently we let the public create a pool for any {whitelisted borrower} x {allowed token} combination. Is there a situation where we want to let Alice have a USDC pool but don't want anyone to give her a USDT pool? I'm not sure there is... but if there is such a situation then here are some other options to consider:
> 
> * drop the borrower whitelist and make single borrower pool creation onlyOwner
> * keep the whitelist, but only allow borrowers to create their own single borrower pools

I personally belive, that letting the protocol go free is generally a good practice. As long as creation of a new pool does not hurt us in any way, we're fine I guess. That might create new opportunities for us to pursue. If someone creates a new pool and users willingly join it, maybe it is the right move to start supporting such pool that has a hight interest. We still can decide not to support the pool, do not create a farm and do not whitelist it in liquidator and it would kind of just stay there on it's own. 

That's just my opinion. Waiting for some feedback from the rest of truefi team :)

MiksuJak:
By the way, we probably need to change liquidator, so it whitelists pools instead of tokens. I think this PR is big enough. If it goes through we should do it in a new one.

yuchenlintt:
Agreed, let's keep this PR focused on pool creation.

1. My poll came back with 11 unanimous votes for `singleBorrowerPool`, so let's go with that.
2. I completely agree with you. Decentralization is better, creating pools is harmless, and it would be a strong sign of lending interest if lenders are depositing into a pool with no farm.

contracts/truefi2/PoolFactory.sol
```diff
@@ -132,16 +160,49 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
...
+    /**
+     * @dev Create a new private pool behind proxy. Update new pool's implementation.
+     * Transfer ownership of created pool to Factory owner.
+     * @param token Address of token which the pool will correspond to.
+     */
+    function createSingleBorrowerPool(
+        address token,
+        string memory borrowerName,
+        string memory borrowerSymbol
+    ) external onlyAllowedTokens(token) onlyWhitelistedBorrowers(msg.sender) {
```
hal909, line 172: I'm not sure if we want whitelisted borrowers to be able to make a pool for every token. I will get clarification on this from the team. @yuchenlintt

hal909 (changes requested):
Just need clarification on whether whitelisted borrowers should be able to create multiple pools. Once I get this, I can approve

contracts/truefi2/PoolFactory.sol
```diff
@@ -41,6 +41,10 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
 
     ISAFU public safu;
 
+    // @dev Mapping of borrowers to mapping of ERC20 token's addresses to its private pools
```
yuchenlintt, line 44:
```suggestion
    // @dev Mapping of borrowers to mapping of ERC20 token's addresses to its single borrower pool
```

contracts/truefi2/PoolFactory.sol
```diff
@@ -50,13 +54,28 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
      */
     event PoolCreated(address token, address pool);
 
+    /**
+     * @dev Event to show creation of the new private pool
```
yuchenlintt, line 58:
```suggestion
     * @dev Event to show creation of the new single borrower pool
```

contracts/truefi2/PoolFactory.sol
```diff
@@ -50,13 +54,28 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
...
+
     /**
      * @dev Event to show that token is now allowed/disallowed to have a pool created
      * @param token Address of token
      * @param status New status of allowance
      */
     event AllowedStatusChanged(address token, bool status);
 
+    /**
+     * @dev Event to show that borrower is now allowed/disallowed to have a private pool
```
yuchenlintt, line 73:
```suggestion
     * @dev Event to show that borrower is now allowed/disallowed to have a single borrower pool
```

contracts/truefi2/PoolFactory.sol
```diff
@@ -132,16 +160,49 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
         emit PoolCreated(token, address(proxy));
     }
 
+    /**
+     * @dev Create a new private pool behind proxy. Update new pool's implementation.
```
yuchenlintt, line 164:
```suggestion
     * @dev Create a new single borrower pool behind proxy. Update new pool's implementation.
```

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -115,6 +115,28 @@ contract TrueFiPool2 is ITrueFiPool2, IPauseableContract, ERC20, UpgradeableClai
         safu = _safu;
     }
 
+    /**
+     * @dev Initializer for custom pools
+     */
+    function customInitialize(
```
yuchenlintt, line 121:
```suggestion
    function initializeSingleBorrower(
```

contracts/truefi2/interface/ITrueFiPool2.sol
```diff
@@ -15,6 +15,15 @@ interface ITrueFiPool2 is IERC20 {
         address __owner
     ) external;
 
+    function customInitialize(
```
yuchenlintt, line 18:
```suggestion
    function initializeSingleBorrower(
```

yuchenlintt (changes requested):

yuchenlintt (approved):
I'm changing my review to an Approve since the PR itself looks good to me now. Still waiting on a higher-up decision on whether we want to launch this, but there's no need to block on me if Hal gets that thumbs up.

hal909 (approved):

PR #745, opened 2021-07-05
Ignore reentrancy detector in claim function

MiksuJak:
Slither was warning us, that a low level call is performed before writing some state variables. It happens because of _SafeERC20_ `_callOptionalReturn` function. It is used for the _safeTransfer_, _safeApprove_ etc. _Calldata_ is explicitly encoded in those functions, therefore it cannot be manipulated.

My proposition is to simply ignore the warning. It would be really nice, if someone else could have a quick look at it, to double check whether im correct.

### DIFF
--- a/contracts/truefi/TrueRatingAgencyV2.sol
+++ b/contracts/truefi/TrueRatingAgencyV2.sol
@@ -379,7 +379,6 @@ contract TrueRatingAgencyV2 is ITrueRatingAgencyV2, Ownable {
      * R = Total Reward = (interest * chi * rewardFactor)
      * @param id Loan ID
      */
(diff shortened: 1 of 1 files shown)


vanruch (approved):

yuchenlintt:
Thanks for digging into the weeds! I don't see any major problems with this as is, although I'm wary of disabling reentrancy checks for the whole function. If we introduce an actual reentrancy risk to claim() later on, slither wouldn't be able to detect it.

I took a shot at inlining calculateTotalRewards() in PR #746. This makes the code a little harder to read but it is another option for fixing slither. Which do you think looks better?

PR #756, opened 2021-07-09
Add utilization function

skalermo:


contracts/truefi2/TrueFiPool2.sol
```diff
@@ -605,6 +605,14 @@ contract TrueFiPool2 is ITrueFiPool2, IPauseableContract, ERC20, UpgradeableClai
         return token.balanceOf(address(this)).sub(claimableFees);
     }
 
+    /**
+     * @dev Utilization of the pool
+     * @return Utilization in basis points
+     */
+    function utilization() public view returns (uint256) {
+        return poolValue().sub(liquidValue()).mul(BASIS_PRECISION).div(poolValue());
```
vanruch, line 613: poolValue calculation is expensive, call it only once

yuchenlintt:
The spec wording is phrased in terms of utilization because that's how the credit team modeled it, but our utilization adjustment is actually calculated using 1 - utilization. I think it would be simpler to make a liquidRatio() function instead of this utilization(), otherwise we'd be spending some extra gas to calculate 1 - (poolValue - liquidValue) / poolValue == liquidValue / poolValue

yuchenlintt (changes requested):
The spec wording is phrased in terms of utilization because that's how the credit team modeled it, but our utilization adjustment is actually calculated using 1 - utilization. I think it would be simpler to make a liquidRatio() function instead of this utilization(), otherwise we'd be spending some extra gas to calculate 1 - (poolValue - liquidValue) / poolValue == liquidValue / poolValue

vanruch:
@yuchenlintt Are you sure? When liquidValue = poolValue, utilization is 0% -> rate = 0, when liquidValue = 0 -> utilization = 100%, rate is high. Looks like it should be `(poolValue - liquidValue) / poolValue`

yuchenlintt:
Yes, I believe I'm correct.

The interest rate adjustment for utilization should asymptotically approach infinity (but capped at 500%) as utilization approaches 100% (discouraging further borrowing and encouraging more lending), and it should approach 0% as utilization approaches 0% (encouraging further borrowing). This gets added to the base rate.

yuchenlintt (approved):
Thanks for adding `liquidRatio()`!

I don't believe `utilization()` will be used anywhere else, and if it isn't used then I'd prefer to remove it. But this isn't a strong opinion so I'm approving and letting you decide whether to keep it.

PR #759, opened 2021-07-09
Group borrowers by their credit scores and calculate interest

vanruch:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -13,12 +13,31 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+    struct SavedInterest {
+        uint256 total;
+        uint256 perShare;
+    }
+
+    struct CreditScoreBucket {
+        uint16 borrowersCount;
+        uint128 timestamp;
+        uint256 rate;
+        uint256 cumulativeInterestPerShare; // How much interest was gathered by 1 wei times 10^27
```
barrutko, line 25: Not sure what that comment means. If there was accumulated 100 USDC-wei of interest, this variable will store a value of 100*10^27, right?

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -66,16 +86,41 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
 
-    function creditScoreAdjustmentRate(address borrower) public view returns (uint256) {
-        uint8 score = creditOracle.getScore(borrower);
+    function updateCreditScore(ITrueFiPool2 pool, address borrower) external {
+        uint8 oldScore = creditScore[pool][borrower];
+        uint8 newScore = creditOracle.getScore(borrower);
+        if (oldScore == newScore) {
+            return;
+        }
+        require(newScore > 0, "TrueCreditAgency: Cannot set credit score to 0");
```
barrutko, line 95: Why?

contracts/truefi2/TrueCreditAgency.sol
vanruch: If for each wei you have borrowed, you have to pay 0.1 wei of interest, it will store 0.1e27.

barrutko: 👌

contracts/truefi2/TrueCreditAgency.sol
vanruch: Division by 0 protection, we shouldn't give score 0 to any existing borrower

test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -150,6 +152,21 @@ describe('TrueCreditAgency', () => {
...
+
+    it('correctly handles the case when credit score is changing', async () => {
+      await creditAgency.connect(borrower).borrow(tusdPool.address, 1000)
+      expect(await creditAgency.creditScore(tusdPool.address, borrower.address)).to.equal(255)
+      expect((await creditAgency.buckets(tusdPool.address, 255)).totalBorrowed).to.equal(1000)
+      expect((await creditAgency.buckets(tusdPool.address, 255)).borrowersCount).to.equal(1)
+
+      await creditOracle.setScore(borrower.address, 200)
+      await creditAgency.connect(borrower).borrow(tusdPool.address, 1000)
+      expect(await creditAgency.creditScore(tusdPool.address, borrower.address)).to.equal(200)
```
barrutko, line 164: Should `creditScore` be "per pool"?

test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -187,4 +205,82 @@ describe('TrueCreditAgency', () => {
...
+    beforeEach(async () => {
+      await creditAgency.allowBorrower(borrower.address, true)
+      await creditAgency.allowBorrower(owner.address, true)
+      await creditAgency.setRiskPremium(1000)
+      await creditOracle.setScore(owner.address, 255)
+    })
+
+    it('interest for single borrower and stable rate', async () => {
+      await creditAgency.connect(borrower).borrow(tusdPool.address, 1000)
+      await timeTravel(YEAR)
```
barrutko, line 219: Should we do a year-long timeTravel, while we expect the interest to be paid off monthly?

test/truefi2/TrueCreditAgency.test.ts
vanruch: It's not perfect, but we have to update credit score separately. We probably want to change this later and iterate over all pools during updates, but for now it's a workaround.

test/truefi2/TrueCreditAgency.test.ts
vanruch: I don't see any connection between these things. Year-long time travel makes tests much easier to understand as we don't have to divide by 12

contracts/truefi2/TrueCreditAgency.sol
barrutko: What if we dewhitelist borrower, he's removed from borrowers list and his score goes to zero? IMO sounds like something that might happen.

test/truefi2/TrueCreditAgency.test.ts
barrutko: Maybe `mapping(ITrueFiPool2 => mapping(address => uint8)) public creditScore;` can be just `mapping(address => uint8) public creditScore;`? And we we will just always omit the pool part?
Instead of always typing `creditScore[pool][user]`, we'll just type `creditScore[user]`? Or will it cause severe problems with rebucketing?

contracts/truefi2/TrueCreditAgency.sol
vanruch: We should just set it to 1

test/truefi2/TrueCreditAgency.test.ts
vanruch: We don't know if score was updated for the pool so we don't know which bucket we should look for

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: Maybe we can use saturating operators that return max uint256 interest for borrowers with credit score 0. It will be cut off by the 500% interest cap anyway.

Protection from divide by zero errors should happen as close as possible to the source, to limit the spread of extraneous error-checking. I think there's no need to separately handle borrowers with 0 credit score, as far as that's feasible.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -66,16 +86,41 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+    function interest(ITrueFiPool2 pool, address borrower) external view returns (uint256) {
+        CreditScoreBucket storage bucket = buckets[pool][creditScore[pool][borrower]];
+        return _interest(pool, bucket, borrower);
+    }
+
     function borrow(ITrueFiPool2 pool, uint256 amount) external onlyAllowedBorrowers {
         require(isPoolAllowed[pool], "TrueCreditAgency: The pool is not whitelisted for borrowing");
+        uint8 oldScore = creditScore[pool][msg.sender];
+        uint8 newScore = creditOracle.getScore(msg.sender);
+        require(newScore > 0, "TrueCreditAgency: Cannot borrow without a credit score");
```
yuchenlintt, line 120: There should be no need to manually restrict this as long as the adjusted credit limit is 0 and interest rate is max uint256

yuchenlintt (commented):
This looks really promising! I think I'll need to review this PR over a couple iterations. Updating individual borrower rates via credit scores is likely one of the core challenge of lines of credit, so it's worth getting this right.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -48,6 +67,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
 
     function setRiskPremium(uint256 newRate) external onlyOwner {
         riskPremium = newRate;
+        // TODO keep separate premiums per pool, pass pool as an argument here
```
barrutko, line 70: TODO in this PR or in some next?

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -66,16 +86,41 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+
+        _rebucket(pool, borrower, oldScore, newScore, borrowed[pool][borrower]);
+    }
+
+    function creditScoreAdjustmentRate(ITrueFiPool2 pool, address borrower) public view returns (uint256) {
+        return _creditScoreAdjustmentRate(creditScore[pool][borrower]);
+    }
+
+    function _creditScoreAdjustmentRate(uint8 score) internal view returns (uint256) {
         if (score == 0) {
```
barrutko, line 105: How can score be 0?

contracts/truefi2/TrueCreditAgency.sol
vanruch: some next

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -85,4 +130,80 @@ contract TrueCreditAgency is UpgradeableClaimable {
         pool.token().safeApprove(address(pool), amount);
         pool.repay(amount);
     }
+
+    function poke(ITrueFiPool2 pool) public {
+        for (uint16 i = 1; i <= MAX_CREDIT_SCORE; i++) {
+            CreditScoreBucket storage bucket = buckets[pool][i];
+            if (bucket.borrowersCount == 0) {
+                continue;
```
barrutko, line 138: Won't that be super costly?

barrutko: Maybe we somehow store min and max credit score that we currently have in use and restrict iteration to that range?

test/truefi2/TrueCreditAgency.test.ts
barrutko: Isn't score the same for every pool?

contracts/truefi2/TrueCreditAgency.sol
vanruch: Giving max uint256 resulted in overflow when we added premium rate a bit further. Caped it with 500%, should not cause problems now

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: We could also use a sorted array of buckets, with max size 256. It would only contain the scores that are already set, with a worst case that behaves like this.

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: I believe Roshan wanted a single premium, with no distinction between different pools. I would hold off on implementing separate premiums until/unless the feature is requested.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -103,9 +102,9 @@ contract TrueCreditAgency is UpgradeableClaimable {
 
     function _creditScoreAdjustmentRate(uint8 score) internal view returns (uint256) {
         if (score == 0) {
-            return type(uint256).max;
```
yuchenlintt, line 72: The interest rate cap is owner-settable, so we should extract this into a state variable. Not the core part of this PR, so we can add a TODO to handle it later.

And to make sure we're on the same page: we want to cap the _overall_ interest rate at 500%. It shouldn't matter how the intermediate calculations propagate values that are over the cap, but I think the code would be cleaner if we made a saturatingAdd() math library function that capped the individual `+`s in `baseRate + riskPremium + creditAdjustment + utilizationAdjustment` at type(uint256).max, and then applied the 500% cap only once at the end.

contracts/truefi2/TrueCreditAgency.sol
vanruch: I used a bit set to mark used buckets in #bucket-2 branch and looks like this could improve the gas cost a lot (6 times in a simple case)

yuchenlintt: That's awesome! Yeah LoC is looking very gas-hungry. I think we'll have to do some profiling after all the pieces are connected to see if there are optimizations like this that we've left on the table.

contracts/truefi2/TrueCreditAgency.sol
vanruch: I think it would be just as clean if we had this maxRateCap and neither component of the rate calculation would not surpass it. This will remove the necessity to take care about overflows when adding anything to INT_MAX

yuchenlintt: Yeah that makes sense. It's worth it to avoid having to do overflow checks for arithmetic near INT_MAX.

barrutko (approved):

MiksuJak (approved):

PR #761, opened 2021-07-09
Make TrueCreditLine farm-like token 

MiksuJak:


contracts/truefi2/TrueCreditLine.sol
```diff
@@ -1,35 +1,116 @@
...
+        uint256 totalNewRewards = newTotalInterestRewards.sub(totalInterestRewards);
+        // return claimable reward for this account
+        // prettier-ignore
+        return claimableRewards[account].add(
+            balanceOf(account).mul(cumulativeTotalRewards.add(totalNewRewards).sub(previousCumulatedRewards[account])).div(totalSupply())
+        );
+    }
+
+    function transfer(address recipient, uint256 amount) public virtual override update(msg.sender) update(recipient) returns (bool) {
+        super.transfer(recipient, amount);
```
vanruch, line 94: Instead of overriding both `transfer` and `transferFrom` you can just override `_ transfer`. Btw you forgot to return true

contracts/truefi2/TrueCreditLine.sol
```diff
@@ -1,38 +1,141 @@
...
+            balanceOf(account).mul(cumulativeTotalRewards.add(totalNewRewards).sub(previousCumulatedRewards[account])).div(totalSupply())
+        );
+    }
+
+    function _transfer(
+        address sender,
+        address recipient,
+        uint256 amount
+    ) internal virtual override update(sender) update(recipient) {
+        super._transfer(sender, recipient, amount);
```
vanruch, line 105: Shouldn't we call _claim for both sender and recipient before doing transfer?

test/truefi2/TrueCreditLine.test.ts
```diff
@@ -13,29 +14,302 @@ use(solidity)
...
+      // owner
+      expect(await creditLine.claimable(owner.address)).to.eq(quarterReward.mul(2))
+      expect(await creditLine.claimableRewards(owner.address)).to.eq(0)
+      expect(await creditLine.previousCumulatedRewards(owner.address)).to.eq(0)
+    })
+
+    it('complex transfer scenario', async () => {
+      // owner and holder transfer credit line tokens between each other
+      await creditLine.mint(holder.address, parseEth(500))
+      await creditLine.mint(owner.address, parseEth(500))
```
vanruch, line 296: Make their initial shares 400-600, just to be more sure

contracts/truefi2/TrueCreditLine.sol
```diff
@@ -1,38 +1,141 @@
...
     using SafeMath for uint256;
     using SafeERC20 for ERC20;
 
     address public borrower;
-    address public pool;
+    ITrueFiPool2 public pool;
+
+    uint256 public principalDebt;
+
+    uint256 unpaidAccumulatedInterest;
```
MiksuJak, line 20: delete

barrutko (approved):

PR #764, opened 2021-07-13
Remove Liquidator contract

skalermo:


deploy/truefi.ts
```diff
@@ -125,18 +122,12 @@ deploy({}, (deployer, config) => {
...
-    liquidator.initialize(trueFiPool, stkTruToken, trustToken, truPriceOracle, loanFactory)
-  })
   runIf(stkTruToken_LinearTrueDistributor.isInitialized().not(), () => {
     stkTruToken_LinearTrueDistributor.initialize(deployParams[NETWORK].DISTRIBUTION_START, deployParams[NETWORK].DISTRIBUTION_DURATION, deployParams[NETWORK].STAKE_DISTRIBUTION_AMOUNT, trustToken)
   })
   runIf(stkTruToken_LinearTrueDistributor.farm().equals(stkTruToken).not(), () => {
     stkTruToken_LinearTrueDistributor.setFarm(stkTruToken)
   })
-  runIf(stkTruToken.initalized().not(), () => {
-    stkTruToken.initialize(trustToken, trueFiPool, trueFiPool, stkTruToken_LinearTrueDistributor, liquidator)
```
yuchenlintt, line 138: We should pass in an `AddressZero` here instead of getting rid of stkTruToken initialization.

yuchenlintt (commented):
Let's also remove the ropsten and mainnet `liquidator` and `liquidator_proxy` entries from `deployments.json`.

deploy/truefi.ts
yuchenlintt: Better yet, we can go remove Liquidator functions from StkTruToken:
- drop the `_liquidator` argument in `initialize()`
- set `liquidator` to `address(0)` in `initialize()`
- remove `setLiquidator()`, `LiquidatorChanged`, etc.

skalermo: If we set liquidator to address(0), we will not be able to call withdraw function because of the onlyLiquidator modifier. As I can see withdraw function is quite important and we do want to keep it.

MiksuJak: liquidator is still needed for liquidations to happen. If we remove it, then no contract will be able to withdraw tru from stkToken. We should get rid of deploy truefi1 script, so it does not bother us during repo cleaning. After all, it would not be needed anymore.

yuchenlintt: Whoa there, deleting `deploy/truefi.ts` is a bit extreme (and outside the scope of a PR that's _just_ supposed to remove Liquidator).

There may be several unused contracts in the truefi deploy script that we can slice off, but the script still serves two critical purposes:
1. Maintenance + upgrades of contracts that we still *do* support, like `StkTruToken`, `TrueFiPool` (until we launch a new tfTUSD pool and have no more loans or funds in the legacy pool), `TrueRatingAgencyV2`, the sushiswap farm, and our future governance contracts.
2. It forces us to pay attention to unintended consequences when repo cleaning. We were only able to catch the fact that Liquidator is still used in `StkTruToken` because this was recorded in this deploy script.

Let's revert the `deploy/truefi.ts` deletion and keep this PR focused on Liquidator removal. Instead of setting liquidator to address(0), is Liquidator2 compatible?

hal909 (approved):

PR #766, opened 2021-07-14
Remove minMintAmount argument from TrueFiPool flush

yuchenlintt:
This aligns TrueFiPool's slippage protections with TrueFiPool2 + CurveYearnStrategy, and makes automating CRV flushes with a keeper much simpler.

Compare `TrueFiPool.flush()` against `TrueFiPool2.flush()`:
https://github.com/trusttoken/smart-contracts/blob/c94ad4e2cc5aa2f4c3050d79eb75585065e287ba/contracts/truefi2/TrueFiPool2.sol#L488

Compare `TrueFiPool._curvePoolDeposit()` against `CurveYearnStrategy.deposit()`:
https://github.com/trusttoken/smart-contracts/blob/c94ad4e2cc5aa2f4c3050d79eb75585065e287ba/contracts/truefi2/strategies/CurveYearnStrategy.sol#L107

### DIFF
--- a/contracts/truefi/TrueFiPool.sol
+++ b/contracts/truefi/TrueFiPool.sol
@@ -539,30 +539,30 @@ contract TrueFiPool is ITrueFiPool, IPauseableContract, ERC20, ReentrancyGuard, 
      * @dev Deposit idle funds into curve.fi pool and stake in gauge
      * Called by owner to help manage funds in pool and save on gas for deposits
      * @param currencyAmount Amount of funds to deposit into curve
-     * @param minMintAmount Minimum amount to mint
      */
-    function flush(uint256 currencyAmount, uint256 minMintAmount) external {
+    function flush(uint256 currencyAmount) external {
         require(currencyAmount <= currencyBalance(), "TrueFiPool: Insufficient currency balance");
 
-        // add TUSD to curve
-        _flush(currencyAmount, minMintAmount);
-
-        // stake yCurve tokens in gauge
-        uint256 yBalance = _curvePool.token().balanceOf(address(this));
-        _curvePool.token().safeApprove(address(_curveGauge), yBalance);
-        _curveGauge.deposit(yBalance);
-
+        uint256 expectedMinYTokenValue = yTokenValue().add(conservativePriceEstimation(currencyAmount));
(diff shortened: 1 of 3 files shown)


vanruch:
The integration test is failing for curve deposits now. Maybe the minAmount is too big?

yuchenlintt:
> The integration test is failing for curve deposits now. Maybe the minAmount is too big?

Yeah it failed 5 times in a row, which would imply an unacceptably high failure rate for us to use in practice.

I can't really do anything about the minMintAmount though. This PR was only intended to match the slippage behavior of TrueFiPool2, which also sets a 99.9% minMintAmount. I think we'll need @hal909's approval to lower that number.

vanruch (changes requested):
So the problem is that you are putting the Curve LP tokens into the Gauge in `_curvePoolDeposit` and as a result the amount isn't changing and it fails on `exchangeProtector` modifier

yuchenlintt:
Ah, that makes sense. I removed the `exchangeProtector` modifier so this implementation matches with `CurveYearnStrategy's.deposit()`. Running locally I got the integration test to pass.

vanruch:
But now don't have any control if Curve really returns us the expected amount. Why not separate deposit part and gauge part and have guard only on the prior like we had before?

yuchenlintt:
That's a fair point if we consider `_curvePoolDeposit()` in isolation, although this function is only called by `flush()`, which checks that `yTokenValue` doesn't slip too much.

I think there are two different approaches you and I are taking for this refactor, both reasonable:
1. your suggestion: make the minimum changes necessary to TrueFiPool
2. this PR: minimize the code divergence between TrueFiPool and {TrueFiPool2 + CurveYearnStrategy}

I would prefer (1) as well if we didn't have to maintain both TrueFiPool2 and TrueFiPool (which is hopefully only in the short term). However I think (2) is a little better in our current situation since it would help us backport any future changes we might need from TrueFiPool2 to TrueFiPool. Code duplication isn't ideal, but it's a compromise.

Let me know if you'd still prefer (1)! I'm happy to refactor this if you still think that's the best approach.

vanruch (approved):
Missed the yValue check, this PR looks good

hal909 (approved):

PR #769, opened 2021-07-16
Use bitmap to check which buckets are used

vanruch:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -50,6 +50,8 @@ contract TrueCreditAgency is UpgradeableClaimable {
 
     ITrueFiCreditOracle public creditOracle;
 
+    uint256 public usedBuckets;
```
yuchenlintt, line 53: Let's add a descriptive comment explaining this is a bitmap and why our uint8 credit scores make it okay to store the bitmap in a uint256. Otherwise readers could get confused and think this is a count.

Probably also include Bitmap in the name, e.g., `usedBucketsBitmap`. Though if you prefer this naming then I wouldn't mind.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -129,11 +131,14 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
-        for (uint16 i = 1; i <= MAX_CREDIT_SCORE; i++) {
-            CreditScoreBucket storage bucket = buckets[pool][i];
-            if (bucket.borrowersCount == 0) {
+        uint256 bitMap = usedBuckets;
+        for (uint16 i = 0; i <= MAX_CREDIT_SCORE; (i++, bitMap >>= 1)) {
+            if (bitMap & 1 == 0) {
                 continue;
             }
+
+            CreditScoreBucket storage bucket = buckets[pool][i];
```
yuchenlintt, line 140: Something to A/B test in a separate PR: can we look up `buckets[pool]` only once? Otherwise we'd call it once for every used bucket in this loop.

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: If you have profiling results, this might also be a good place to document how much gas a bitmap empirically saves compared to a direct mapping.

yuchenlintt (changes requested):
This looks amazing and very clever! I especially like how it uses our uint8 credit scores to exactly fit inside a uint256.

If you recorded some profiling results, I think it would be great for future debuggers + auditors if you could document how much gas we save by switching like this to a bitmap.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -129,11 +131,14 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+
             uint256 timeNow = block.timestamp;
```
yuchenlintt, line 142: Same as the line above (in a separate PR). The block timestamp can probably be taken out of the inner loop.

contracts/truefi2/TrueCreditAgency.sol
vanruch: Surprisingly it reduced the gas consumption just by a tiny bit (383741 vs 385374) for a large amount of buckets. Solidity must be doing some internal optimization. But still a good catch, will take it out of loop

vanruch: Surprisingly it reduced the gas consumption just by a tiny bit (383741 vs 385374) for a large amount of buckets (20). Solidity must be doing some internal optimization. But still a good catch, will take it out of loop

yuchenlintt (approved):
Looks great, thanks!

PR #772, opened 2021-07-19
Add borrow limit view function

vanruch:


yuchenlintt (changes requested):
Thanks, the overall structure of this looks great!

It looks like handling non-integer exponents is somewhat challenging and pulls in a fixed point library.

I'm wondering if we could simplify implementation and gas by using reduced fractions, e.g., instead of 75_00/100_00, we'd just store the numerator and denominator 3/4. Integer exponentiation is easy. Could we calculate integer roots more cleanly, or would it be more work + gas than using an existing pow() library?

yuchenlintt:
I did some back-of-the-envelope math for calculating integer roots using [Newton's method](https://en.wikipedia.org/wiki/Newton%27s_method). This might be worth considering as an alternative to ABDKMath64x64, though it adds complexity in implementing our own library.

---

Goal: Find an integer approximation to the `k`th root `x` of an integer `N`.

Algorithm: Let `f(x) = x^k - N`, so `f(x) == 0` iff `x == kth root of N`. Then the sequence
```
x0 = N/k
            f(xn)    (k-1) * xn^k + N
xn+1 = xn - ------ = ----------------
            f'(xn)     k * xn^(k-1)
```
converges very quickly to the `k`th root of `N`. We could probably run 4 iterations and be more precise than 1 in 10_000.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -115,6 +161,48 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+                borrowed[pools[i]][borrower].mul(resultPrecision).div(
+                    uint256(10)**(ITrueFiPool2WithDecimals(address(pools[i])).decimals())
+                )
+            );
+        }
+        return borrowSum;
+    }
+
+    function borrowLimitAdjustment(uint256 score) public view returns (uint256) {
+        return ((ABDKMath64x64.fromUInt(score) / MAX_CREDIT_SCORE).pow(borrowLimitConfig.limitAdjustmentPower) * 10000).toUInt();
```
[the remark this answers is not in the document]
yuchenlintt: If we care a lot about gas expenditure in this function (profile first to see if this is necessary!), another optimization we could try is to cache MAX_CREDIT_SCORE.pow(limitAdjustmentPower) and divide score.pow(limitAdjustmentPower) by the cached value each time.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -6,10 +6,16 @@ import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
 import {ERC20, IERC20, SafeMath} from "../common/UpgradeableERC20.sol";
 import {UpgradeableClaimable} from "../common/UpgradeableClaimable.sol";
 import {ITrueFiCreditOracle} from "./interface/ITrueFiCreditOracle.sol";
+import {ABDKMath64x64} from "./libraries/ABDKMath64x64.sol";
+
+interface ITrueFiPool2WithDecimals is ITrueFiPool2 {
```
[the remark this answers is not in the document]
vanruch: Sadly not, because solidity doesn't know if it should override ERC20 or ITrueFiPool2With

contracts/truefi2/TrueCreditAgency.sol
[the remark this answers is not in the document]
vanruch: 1. Yes, it's ok. We'll need to use div function from the library if we wanted to divide on `ABDKMath64x64` by another 64x64 number
2. `pow()` expects second argument to be in basis point precision
3. We would still need to calculate `pow` once so this shouldn't optimize the function by a lot

vanruch:
An idea with Newton's method is cool but we would probably have to do this on fixed-point decimals and I doubt this will save us a lot of gas. But if we have time to experiment, could be totally worth it

contracts/truefi2/TrueCreditAgency.sol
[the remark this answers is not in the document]
yuchenlintt: Ah darn, diamond inheritance. This looks good for now.

contracts/truefi2/TrueCreditAgency.sol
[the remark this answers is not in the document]
yuchenlintt: 1. Gotcha. `ABDKMath64x64` is fixed point arithmetic so normal arithmetic operations on it still work (just need to scale by the binary point and keep track of the sign).
2. Oh, it looks like that's something we added to `ABDKMath64x64`.
3. That's a good point. No need to optimize early and make the code illegible with no real gain.

contracts/truefi2/libraries/ABDKMath64x64.sol
```diff
@@ -0,0 +1,210 @@
+// SPDX-License-Identifier: BSD-4-Clause
```
yuchenlintt, line 1: I was confused by your mention that `pow()` expects basis point precision. Just realized today that we introduced our own modifications to `ABDKMath64x64`. To prevent similar auditor confusion, could you rename this file + contract to clarify that we're not using the verbatim original? Something like `TrueFiFixed64x64` (but at your discretion).

We still need to preserve the license + attribution, but I think we should prepend a docstring that:
1. explains how to find which functions we've added and removed (e.g., `Modifications made by TrueFi can be found by searching for lines that begin with a 'TFTFTF' comment.`)
2. links to the `ABDKMath64x64` github version that we based this on.

contracts/truefi2/libraries/ABDKMath64x64.sol
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /**
+     * Calculate x * y rounding down.  Revert on overflow.
+     *
+     * @param x signed 64.64-bit fixed point number
+     * @param y signed 64.64-bit fixed point number
+     * @return signed 64.64-bit fixed point number
+     */
+    function mul(int128 x, int128 y) internal pure returns (int128) {
```
yuchenlintt, line 190: The original library implements `mul()` before `log_2()` and `exp_2()`. Could we move this above as well for consistency?

contracts/truefi2/libraries/ABDKMath64x64.sol
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * @dev calculate x^y using the fact that
+     * x^y = (2^log2(x))^y = 2^(y * log2(x))
+     *
+     * @param x is 64.64-bit fixed point number
+     * @param y is basis-point precision (10000 = 100%)
+     * @return signed 64.64-bit fixed point number
+     */
+    function pow(int128 x, uint256 y) internal pure returns (int128) {
```
yuchenlintt, line 204: 1. Let's make both arguments 64.64-bit fixed point numbers. We can handle the `fromUint(y) / 10000` conversion outside of this library, which would make this library agnostic to our contract basis precision.
2. We should also rename the function to something like `powf()`, which disambiguates it from ABDK's original `pow()` function that we removed. I'm not sure `powf()` is the best name since the `f` usually means float, so if you have better suggestions I'm happy to hear them.
3. Let's move this above `log_2()` and below `mul()` for consistency with ABDK.
4. Prepend a comment line to explain TrueFi's modifications on top of ABDK, something like `TFTFTF Add a new pow implementation that takes 64.64-bit fixed point numbers for both arguments.`

contracts/truefi2/mocks/PowTest.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
```
yuchenlintt, line 1: Let's remove this wrapper contract and test directly against the library. The pow unit test should still be legible even if it has some `toUint()` and `fromUint()` conversion calls.

yuchenlintt (changes requested):
Great! Let's move forward with fixed point 64.64 for now. If we run into performance issues then we can test out alternate options like Newton's method.

All I have left is some feedback for our imported ABDK library, otherwise I think this PR should be ready to submit.

contracts/truefi2/libraries/ABDKMath64x64.sol
vanruch: 1. 👍 
2. We can call it `powi()` to keep in line with ABDK's naming
3. 👍 
4. 👍

contracts/truefi2/mocks/PowTest.sol
vanruch: We can't deploy the library and therefore can't unit test it. I'll keep the contract as simple as possible

contracts/truefi2/libraries/ABDKMath64x64.sol
yuchenlintt: 2.
I think the `i` in `muli()`, etc. stands for signed integer, which indicates the second arg:
https://github.com/abdk-consulting/abdk-libraries-solidity/blob/d8817cb600381319992d7caa038bf4faceb1097f/ABDKMath64x64.sol#L163

You can also see that `mulu()` multiplies by an unsigned integer second arg:
https://github.com/abdk-consulting/abdk-libraries-solidity/blob/d8817cb600381319992d7caa038bf4faceb1097f/ABDKMath64x64.sol#L201

Their convention for no suffix is to take an int128 [fixed64x64] as second arg:
https://github.com/abdk-consulting/abdk-libraries-solidity/blob/d8817cb600381319992d7caa038bf4faceb1097f/ABDKMath64x64.sol#L147

Unfortunately ABDK doesn't follow this convention for their `pow()`, which instead takes a uint256 second arg and should probably be named `powi()` instead.

contracts/truefi2/mocks/PowTest.sol
yuchenlintt: Oops, I missed that it was a library. Makes sense!

contracts/truefi2/libraries/ABDKMath64x64.sol
vanruch: Our implementation expects signed integers although it will fail for negative `y`. Let's keep `powi()`. Or name it totally differently, e.g. `decimalPow()`

yuchenlintt: I didn't mean to derail the whole PR for a single function name. Sorry about this :)

The [`powX()` for type `X`] naming convention is an often confusing legacy of ANSI C, so I agree that we might do better just expanding the name. I like `decimalPow()`, although our current implementation (as of dd54f45) is more of a `2^(y * log2(x))`, which doesn't directly deal with decimals (vs binary fixed point).

Let's settle on `fixed64x64Pow()`, which is a little verbose but should be completely unambiguous.

yuchenlintt (approved):

PR #775, opened 2021-07-20
Introduce monthly repayments periods

MiksuJak:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -14,6 +14,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
 
     uint8 constant MAX_CREDIT_SCORE = 255;
     uint256 constant MAX_RATE_CAP = 50000;
+    uint256 constant REPAYMENT_PERIOD = 31 days;
```
barrutko, line 17: Not sure if it wouldn't be nice if this was configurable.

MiksuJak: right

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -148,6 +150,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
         require(isPoolAllowed[pool], "TrueCreditAgency: The pool is not whitelisted for borrowing");
         uint8 oldScore = creditScore[pool][msg.sender];
         uint8 newScore = creditOracle.getScore(msg.sender);
+        nextRepaymentTerm[pool][msg.sender] = block.timestamp.add(REPAYMENT_PERIOD);
```
barrutko, line 153: No good. If someone has some debt already, he can push the moment of repayment forward, without repaying, but with just borrowing more. It cannot work like that.

barrutko (changes requested):
There needs to be a repayment if the debt is already non-zero in order to push liquidation moment.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -39,11 +39,14 @@ contract TrueCreditAgency is UpgradeableClaimable {
 
     mapping(ITrueFiPool2 => mapping(address => uint8)) public creditScore;
     mapping(ITrueFiPool2 => mapping(address => uint256)) public borrowed;
+    mapping(ITrueFiPool2 => mapping(address => uint256)) public nextRepaymentTerm;
```
barrutko, line 42: Not sure if the word `term` fits well here. Maybe `date` or `timestamp`? Not sure tbh.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -148,14 +159,19 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
 
-        _rebucket(pool, msg.sender, oldScore, newScore, borrowed[pool][msg.sender].add(amount));
+        _rebucket(pool, msg.sender, oldScore, newScore, currentDebt.add(amount));
 
         pool.borrow(amount);
         pool.token().safeTransfer(msg.sender, amount);
     }
 
     function repay(ITrueFiPool2 pool, uint256 amount) external {
+        nextRepaymentTimestamp[pool][msg.sender] = block.timestamp.add(repaymentPeriod);
```
yuchenlintt, line 174: If the borrower repays with `amount` less than their interest owed or even 0, this would let them reset their repayment timestamp. We should only reset if `amount >= interest(pool, borrower)`.

yuchenlintt: Borrowers who might not be accustomed to block-by-block interest accrual could get surprised if they send an `amount` a couple cents short of their interest and end up defaulting. In separate PRs we should consider adding `repayInterest(pool)` or `repayInFull(pool)` functions to help them meet our repay schedules.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -39,11 +39,14 @@ contract TrueCreditAgency is UpgradeableClaimable {
 
     mapping(ITrueFiPool2 => mapping(address => uint8)) public creditScore;
     mapping(ITrueFiPool2 => mapping(address => uint256)) public borrowed;
+    mapping(ITrueFiPool2 => mapping(address => uint256)) public nextRepaymentTimestamp;
```
yuchenlintt, line 42: I'm sorry to make you rename this yet again, but we should clarify that this is for interest repayments. Let's call this `nextInterestRepaymentTimestamp`, or shorten it if you'd like (`nextInterestRepayTime`?). Update elsewhere as well.

We'll also need to track biannual full repayments, but that can be done in another PR after this one.

yuchenlintt (changes requested):
Thanks for the PR!

yuchenlintt (approved):

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -40,11 +40,14 @@ contract TrueCreditAgency is UpgradeableClaimable {
     mapping(ITrueFiPool2 => mapping(address => uint8)) public creditScore;
     mapping(ITrueFiPool2 => mapping(address => uint256)) public borrowed;
     mapping(ITrueFiPool2 => mapping(address => uint256)) public totalPaidInterest;
+    mapping(ITrueFiPool2 => mapping(address => uint256)) public nextInterestRepayTime;
 
     mapping(ITrueFiPool2 => bool) public isPoolAllowed;
 
     mapping(address => bool) public isBorrowerAllowed;
 
+    uint256 public repaymentPeriod;
```
yuchenlintt, line 49: Let's update this name with the result of this discussion thread: https://github.com/trusttoken/smart-contracts/pull/780#discussion_r675704796

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -271,6 +286,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
     function _payInterestWithoutTransfer(ITrueFiPool2 pool) internal returns (uint256) {
         uint256 accruedInterest = interest(pool, msg.sender);
         totalPaidInterest[pool][msg.sender] = totalPaidInterest[pool][msg.sender].add(accruedInterest);
+        nextInterestRepayTime[pool][msg.sender] = block.timestamp.add(repaymentPeriod);
```
yuchenlintt, line 289: The version of `_payInterestWithoutTransfer()` from the base branch (#778) allows partial interest repayment of a custom amount, which would be incompatible with this version since a borrower could reset their period with less than a full interest payment.

Why don't we punt this decision (where in the code we reset the interest repay time) to that PR? This PR would just remain focused on setting the interest repayment period.

You can delete this line for now and we can figure out where it best belongs in PR #778.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -86,16 +91,22 @@ contract TrueCreditAgency is UpgradeableClaimable {
         creditAdjustmentCoefficient = 1000;
         utilizationAdjustmentCoefficient = 50;
         utilizationAdjustmentPower = 2;
+        repaymentPeriod = 31 days;
+    }
+
+    modifier onlyAllowedBorrowers() {
```
yuchenlintt, line 97: When I tried to auto-resolve merge conflicts with the base branch, Github duplicated this function. We probably want to remove this modifier change from this branch.

yuchenlintt (changes requested):
My mistake, I approved without realizing this base branch was moved from dev to repay-mechanisms. I force-pushed to undo my bad merge conflict resolution.

skalermo (approved):

yuchenlintt (approved):

barrutko (approved):

PR #778, opened 2021-07-21
Separate repayment mechanisms in Credit Agency

MiksuJak:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -155,10 +159,22 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+        emit InterestPaid(pool, msg.sender, accruedInterest);
+    }
+
+    function repayPrincipal(ITrueFiPool2 pool, uint256 amount) external {
+        uint256 currentDebt = borrowed[pool][msg.sender];
+        require(currentDebt >= amount, "TrueCreditAgency: Cannot repay more than principal debt");
         pool.token().safeTransferFrom(msg.sender, address(this), amount);
         pool.token().safeApprove(address(pool), amount);
         pool.repay(amount);
+        borrowed[pool][msg.sender] = currentDebt.sub(amount);
```
vanruch, line 176: You'll have to `poke` here

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -155,10 +159,22 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
-    function repay(ITrueFiPool2 pool, uint256 amount) external {
+    function payInterest(ITrueFiPool2 pool) external {
+        uint256 accruedInterest = interest(pool, msg.sender);
+        pool.token().safeTransferFrom(msg.sender, address(this), accruedInterest);
+        pool.token().safeApprove(address(pool), accruedInterest);
+        pool.repay(accruedInterest);
+        emit InterestPaid(pool, msg.sender, accruedInterest);
+    }
+
+    function repayPrincipal(ITrueFiPool2 pool, uint256 amount) external {
```
vanruch, line 170: Should we cover interest first before decreasing principal borrow amount? I'd argue that we need a method to pay interest and keep normal repay.

contracts/truefi2/TrueCreditAgency.sol
vanruch: And move state-changing lines before calling transfer and repay on pool

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: I agree. We shouldn't have a `repayPrincipal()` function because any payment must always cover interest first before it touches principal. The `payInterest()` function is for borrower convenience so they know they've fully covered interest for a given month.

We can change `repayPrincipal()` to a `repayInFull()` that completely settles this line of credit to 0 principal, 0 interest.

MiksuJak: @yuchennil As far as I know we also want to have an option to decrease principal debt, without repaying it in full. I just pushed new changes that should fix the problem.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -155,10 +160,23 @@ contract TrueCreditAgency is UpgradeableClaimable {
         pool.token().safeTransfer(msg.sender, amount);
     }
 
-    function repay(ITrueFiPool2 pool, uint256 amount) external {
-        pool.token().safeTransferFrom(msg.sender, address(this), amount);
-        pool.token().safeApprove(address(pool), amount);
-        pool.repay(amount);
+    function payInterest(ITrueFiPool2 pool) public {
```
vanruch, line 163: make it external

test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -261,17 +261,86 @@ describe('TrueCreditAgency', () => {
...
+
+    it('increases paidTotalInterest', async () => {
+      await creditAgency.connect(borrower).payInterest(tusdPool.address)
+      expect(await creditAgency.paidTotalInterest(tusdPool.address, borrower.address)).to.be.closeTo(BigNumber.from(100), 2)
+    })
+
+    it('pays close to nothing on second call', async () => {
+      await creditAgency.connect(borrower).payInterest(tusdPool.address)
+      await creditAgency.connect(borrower).payInterest(tusdPool.address)
+      expect(await creditAgency.paidTotalInterest(tusdPool.address, borrower.address)).to.be.closeTo(BigNumber.from(100), 2)
```
vanruch, line 289: This test doesn't say much without the context of the previous one. You can either add same expect after first payment or extract the amount from the event

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: Right, having the option to reduce principal without full repayment is important, but we can do that with the normal `repay(amount)` function. We should have:
- a one-param `repay(amount)` that can pay arbitrary amounts. This covers interest first and then pays down principal with the remaining amount (if any).
- a zero-param `payInterest()` that clears only interest. This is a convenience function to help borrowers make their monthly interest payments without leaving a couple cents unpaid.
- a zero-param `repayInFull()` that clears both interest and principal. This is a convenience function to help borrowers make their biannual full repayment without leaving a couple cents unpaid.

We can probably implement the two convenience functions by just delegating to the first one. `payInterest()` would be something like`repay(interest())` and `repayInFull()` would be `repay(interest() + principal())`.

In no cases should a borrower be allowed to reduce principal unless interest is exactly 0 during this block. Hope this helps!

vanruch: @yuchenlintt Do you reckon borrowers might want to have a bigger control over the principal debt? I mean, would they prefer calling repay(x) that will cover interest first and then principal amount (which means it will be hard to get exact princ. debt) or call  repayPrincipal(x) that will charge them `x + interest` tokens so the borrowed amount is more clear?

yuchenlintt: You make a good point. Given how our credit limit works I could see how borrowers might like an easy way to pay down to a specific principal, rather than hope their specified repay amount is enough.

I'm worried about semantics though. A function name like `repayPrincipal()` _feels_ like it just pays the principal without touching interest. That's especially the case if we also have a `repayInterest()` function that only pays interest without touching principal.

If we implement this, I think it might make sense to specify a target principal amount as the argument rather than an amount of principal to repay. For example, if (limit, principal, interest) == ($10, $9.5, $1) then repayToTargetPrincipal($9) would transfer $1.5 and leave ($10, $9, $0).

yuchenlintt (dismissed):
I think this is coming along really well and very close to ready.

Three proposed changes:
1. delegate payInterest() to call repay(interest())
2. fix the _payInterestWithoutTransfer() call without a corresponding transfer
3. handle repay period logic in repay(), and handle interest/principal-specific logic in its respective `_payXWithoutTransfer()` function

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -268,10 +273,35 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+        _payInterestWithoutTransfer(pool, accruedInterest);
+        _repay(pool, accruedInterest);
+    }
+
+    function repay(ITrueFiPool2 pool, uint256 amount) public {
+        uint256 currentDebt = borrowed[pool][msg.sender];
+        uint256 accruedInterest = interest(pool, msg.sender);
+        if (amount < accruedInterest) {
+            _payInterestWithoutTransfer(pool, amount);
+            return;
```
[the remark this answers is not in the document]
MiksuJak: > If you return here, then you haven't transferred any tokens yet.

How did I miss it 😓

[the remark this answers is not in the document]
MiksuJak: The reason I created `_payInterestWithoutTransfer ` is to not write the same lines of code twice. In case of principal repayment it happens only once in this contract.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -268,10 +273,34 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
+        uint256 currentDebt = borrowed[pool][msg.sender];
+        uint256 accruedInterest = interest(pool, msg.sender);
+
+        if (amount <= accruedInterest) {
+            _payInterestWithoutTransfer(pool, amount);
+        } else {
+            _payInterestWithoutTransfer(pool, accruedInterest);
+
+            uint256 repaidPrincipal = amount.sub(accruedInterest);
+            require(currentDebt.add(accruedInterest) >= amount, "TrueCreditAgency: Cannot repay more than debt");
```
barrutko, line 290: ☝️

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -268,10 +273,34 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
-        pool.repay(amount);
+    function payInterest(ITrueFiPool2 pool) external {
+        repay(pool, interest(pool, msg.sender));
+    }
+
+    function repay(ITrueFiPool2 pool, uint256 amount) public {
+        uint256 currentDebt = borrowed[pool][msg.sender];
+        uint256 accruedInterest = interest(pool, msg.sender);
+
+        if (amount <= accruedInterest) {
```
barrutko, line 284: <

barrutko (approved):

PR #779, opened 2021-07-22
Add rate calculation function

skalermo:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -176,7 +188,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
             bucket.cumulativeInterestPerShare = bucket.cumulativeInterestPerShare.add(
                 bucket.rate.mul(1e23).mul(timeNow.sub(bucket.timestamp)).div(365 days)
             );
-            bucket.rate = _creditScoreAdjustmentRate(uint8(i)).add(riskPremium);
+            bucket.rate = _currentRate(partialRate, _creditScoreAdjustmentRate(uint8(i)));
```
yuchenlintt, line 191: Looks like we call `_currentRate()` with the same arguments as `currentRate()`. Can we deduplicate and just call `currentRate()` from here instead?

That should also let us change `currentRate()` to a one-liner:
min(riskPremium + utilizationAdjustment + creditScoreAdjustment, MAX_RATE_CAP).

yuchenlintt (changes requested):

contracts/truefi2/TrueCreditAgency.sol
skalermo: Thank you for pointing this out. Actually I was having doubts about the need of `_currentRate()` function. 

I really wanted to save some gas by utilizing precalculated sum of `riskPremium` and `utilizationAdjustmentRate()` stored as a local variable `partialRate`. Then my thoughts were that it would be nice to wrap adding and capping of `partialRate` and `creditScoreAdjustmentRate()` in a function which I named `_currentRate()`.

I agree this approach somewhat hurts readability. Maybe some comments and docstrings could improve it.

yuchenlintt: I see your intent now.

If we haven't profiled and found a significant performance difference, I would prefer to keep the simpler implementation for code clarity. If you find a performance difference, document it so we know why we're opting for more convoluted code.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -165,6 +176,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
         uint256 bitMap = usedBucketsBitmap;
         CreditScoreBucket[256] storage creditScoreBuckets = buckets[pool];
         uint256 timeNow = block.timestamp;
+        uint256 partialRate = utilizationAdjustmentRate(pool).add(riskPremium);
```
yuchenlintt, line 179: If we decide to keep this (pending this comment: https://github.com/trusttoken/smart-contracts/pull/779/files#r675697168), swap the order to `riskPremium + utilizationAdjustmentRate` to match `currentRate()`'s order. I think it also reads more intuitively left-to-right.

yuchenlintt (dismissed):

vanruch:
@yuchenlintt could you check the PR again?

contracts/truefi2/TrueCreditAgency.sol
skalermo: Profiling showed gas use reduction by roughly 200_000 for 20 used buckets.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -274,6 +274,21 @@ contract TrueCreditAgency is UpgradeableClaimable {
         return saturatingSub(poolBorrowMax, totalBorrowed(borrower, poolDecimals));
     }
 
+    function currentRate(ITrueFiPool2 pool, address borrower) external view returns (uint256) {
+        return _currentRate(riskPremium.add(utilizationAdjustmentRate(pool)), creditScoreAdjustmentRate(pool, borrower));
+    }
+
+    /**
+     * @dev Helper function used by poke() to save gas by calculating partial terms of the total rate.
+     * For 20 used buckets gas use reduction is roughly 200_000.
```
MiksuJak, line 283: I don't think it is a necessary comment

MiksuJak (approved):

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: Thanks for profiling this. That's a pretty nice improvement!

PR #780, opened 2021-07-22
Yearly repayments

MiksuJak:


yuchenlintt (changes requested):

barrutko:
Is the solution flash-loan-proof?

hal909 (dismissed):
- 365 days after the borrower is whitelisted, they have 3 days to pay down their full balance
- After 365 days the whitelist status is removed and a borrower needs to be re-whitelisted in order to borrow again

yuchenlintt:
> * 365 days after the borrower is whitelisted, they have 3 days to pay down their full balance
> * After 365 days the whitelist status is removed and a borrower needs to be re-whitelisted in order to borrow again

Idea: we could remove the `isBorrowerAllowed` boolean and just use a `borrowerAllowedUntilTime` timestamp to track borrower allowed status. Borrowers who don't have an LOC would have timestamp 0, those who are currently allowed have a timestamp in the future, and those with a timestamp in the past can be defaulted if their principal + interest are nonzero. The only way to change `borrowerAllowedUntilTime` would be to manually call `allowBorrower(address, timestamp)`.

MiksuJak:
> > * 365 days after the borrower is whitelisted, they have 3 days to pay down their full balance
> > * After 365 days the whitelist status is removed and a borrower needs to be re-whitelisted in order to borrow again
> 
> Idea: we could remove the `isBorrowerAllowed` boolean and just use a `borrowerAllowedUntilTime` timestamp to track borrower allowed status. Borrowers who don't have an LOC would have timestamp 0, those who are currently allowed have a timestamp in the future, and those with a timestamp in the past can be defaulted if their principal + interest are nonzero. The only way to change `borrowerAllowedUntilTime` would be to manually call `allowBorrower(address, timestamp)`.

That's a great idea! What I like about it the most, is that we turn borrowers whitelist and repayment term into 1 variable. I'll implement it right away.

yuchenlintt (approved):

MiksuJak:
> * 365 days after the borrower is whitelisted, they have 3 days to pay down their full balance
> * After 365 days the whitelist status is removed and a borrower needs to be re-whitelisted in order to borrow again

I think we should include grace period in liquidations PR. For now we just handle borrowing allowances and grace period will be used only to tell if a LoC is in default or not.

PR #781, opened 2021-07-26
Restrict borrowers from borrowing over their limit

skalermo:


test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -385,6 +387,16 @@ describe('TrueCreditAgency', () => {
         .to.be.revertedWith('TrueCreditAgency: The pool is not whitelisted for borrowing')
     })
 
+    it('cannot borrow over the borrow limit', async () => {
+      await creditAgency.allowBorrower(borrower2.address, true)
+      await creditOracle.setScore(borrower2.address, 191)
+      await creditOracle.setMaxBorrowerLimit(borrower2.address, parseEth(100))
+
+      expect(await creditAgency.borrowLimit(tusdPool.address, borrower2.address)).to.eq(parseEth(80.51))
+      await expect(creditAgency.connect(borrower2).borrow(tusdPool.address, parseEth(80.51).add(1)))
```
barrutko, line 396: Would be nice to add a test case, where at first borrower borrows i.e. `parseEth(75)` and it works, and then attempts to borrow more - i.e. `parseEth(5.51).add(1)` and it fails.

test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -256,11 +259,11 @@ describe('TrueCreditAgency', () => {
     beforeEach(async () => {
       pool1 = await deployMockContract(owner, TrueFiPool2Json.abi)
       pool2 = await deployMockContract(owner, TrueFiPool2Json.abi)
-      await pool1.mock.poolValue.returns(parseEth(10))
+      await pool1.mock.poolValue.returns(parseEth(10_000))
```
MiksuJak, line 262: I really do not like using mock contracts in test suites, especially those that are developed by us. I know, that it was not introduced with this PR. I probably missed it in one of the old ones. Consider changing it here. It does not look to be too big of a change for me, but if any complications may occur, we can still do it in a new PR.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -172,15 +172,21 @@ contract TrueCreditAgency is UpgradeableClaimable {
...
-        uint8 newScore = creditOracle.getScore(borrower);
+        (uint8 oldScore, uint8 newScore) = _updateCreditScore(pool, borrower);
         if (oldScore == newScore) {
             return;
         }
 
         _rebucket(pool, borrower, oldScore, newScore, borrowed[pool][borrower]);
     }
 
+    function _updateCreditScore(ITrueFiPool2 pool, address borrower) internal returns (uint8, uint8) {
```
MiksuJak, line 183: 👍

test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -67,6 +68,8 @@ describe('TrueCreditAgency', () => {
     await creditAgency.allowPool(tusdPool.address, true)
 
     await creditOracle.setScore(borrower.address, 255)
+    await creditOracle.setMaxBorrowerLimit(owner.address, parseEth(100_000_000))
```
MiksuJak, line 71: Is this ever used?

skalermo: It is, in tests we `borrow()` couple times on behalf of the `owner`.

MiksuJak (approved):

yuchenlintt (approved):

PR #783, opened 2021-07-27
Aave base rate oracle

skalermo:


contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /**
+     * @dev Average apy is calculated by taking
+     * the time-weighted average of the aave deposit apys.
+     * Essentially formula given below is used:
+     *
+     *           (v + v_{n-1}) / 2 * (t - t_{n-1}) + sum_{i=1}^{n - 1} (v_i + v_{i-1}) / 2 * (t_i - t_{i-1})
+     * avgAPY = -------------------------------------------------------------------------------------------
+     *                                                  (t - t_0)
```
yuchenlintt, line 104: LRS: Update comment to sum_{i=1}^n v_i * (t_i - t_{i-1}) on top and t_n - t_0 below.

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,161 @@
...
+     */
+    function calculateAverageAPY(uint256 timeToCover) public view returns (uint256) {
+        require(
+            1 days <= timeToCover && timeToCover <= 365 days,
+            "AaveBaseRateOracle: Expected amount of time in range 1 to 365 days"
+        );
+        // estimate how much buffer we need to use
+        uint16 bufferSizeNeeded = uint16(timeToCover.div(cooldownTime));
+        require(bufferSizeNeeded <= bufferSize(), "AaveBaseRateOracle: Needed buffer size cannot exceed size limit");
+        uint16 lidx = totalsBuffer.latestIndex;
```
yuchenlintt, line 127: Update name to _currIndex

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,161 @@
...
+        uint16 startIndex = (lidx + bufferSize() - bufferSizeNeeded + 1) % bufferSize();
+        if (totalsBuffer.timestamps[startIndex] == 0) {
+            startIndex = 0;
+        }
+        uint256 runningTotalForTimeToCover = totalsBuffer.runningTotals[lidx].sub(totalsBuffer.runningTotals[startIndex]);
+        uint256 curValue = getAaveVariableBorrowAPY(asset);
+        uint256 curTimestamp = block.timestamp;
+        uint256 dt = curTimestamp.sub(totalsBuffer.timestamps[lidx]);
+        runningTotalForTimeToCover = runningTotalForTimeToCover.add(curValue.add(totalsBuffer.lastValue).mul(dt).div(2));
```
yuchenlintt, line 136: SPOT: Let's delete these lines so the average rate doesn't depend on a current spot rate. We'll sacrifice higher accuracy so that the average rate will be harder to game.

yuchenlintt (changes requested):
Let's simplify this oracle following my comments here: https://github.com/trusttoken/smart-contracts/pull/630#pullrequestreview-711987567

1. LRS: Switch from trapezoid rule back to a left Riemann sum
2. SPOT: Drop current spot rate from the average rate calculation

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,161 @@
...
+    ) public {
+        aavePool = _aavePool;
+        cooldownTime = _cooldownTime;
+        asset = _asset;
+
+        totalsBuffer.lastValue = getAaveVariableBorrowAPY(asset);
+        totalsBuffer.timestamps[0] = block.timestamp;
+    }
+
+    function bufferSize() public virtual pure returns (uint16) {
```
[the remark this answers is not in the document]
skalermo: Overriding this function allows us to set different `BUFFER_SIZE` constant in mock contracts. It helps us to setup the contract with smaller buffer size in tests. I'd rather not delete it unless you insist.

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,161 @@
...
+    uint256 private constant BASIS_PRECISION = 10000;
+
+    // A cyclic buffer structure for storing running total (cumulative sum)
+    // values and their respective timestamps.
+    // latestIndex points to the previously inserted value.
+    struct RunningTotalsBuffer {
+        uint256[MAX_BUFFER_SIZE] runningTotals;
+        uint256[MAX_BUFFER_SIZE] timestamps;
+        uint16 latestIndex;
+        uint256 lastValue;
```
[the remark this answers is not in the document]
skalermo: LRS: The funny thing is we still can't delete `lastValue`. The way we calculate average apy is by using running totals which require current apy and its time weight which we can't have yet in case of left Riemann sum. So either we save current apy and wait for next update's timestamp to get time weight or perhaps we could go with right Riemann sum.

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,161 @@
...
+     * new total running value.
+     * If the buffer is filled overwrites the oldest value
+     * with a new one and updates its timestamp.
+     */
+    function update() public offCooldown {
+        uint16 lidx = totalsBuffer.latestIndex;
+        uint16 nextIndex = (lidx + 1) % bufferSize();
+        uint256 apy = getAaveVariableBorrowAPY(asset);
+        uint256 dt = block.timestamp.sub(totalsBuffer.timestamps[lidx]);
+        totalsBuffer.runningTotals[nextIndex] = totalsBuffer.runningTotals[lidx].add(apy.add(totalsBuffer.lastValue).mul(dt).div(2));
```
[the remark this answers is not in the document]
skalermo: Isn't it right Riemann sum?

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,163 @@
...
+        returns (
+            uint256[BUFFER_SIZE] memory,
+            uint256[BUFFER_SIZE] memory,
+            uint16
+        )
+    {
+        return (totalsBuffer.runningTotals, totalsBuffer.timestamps, totalsBuffer.currIndex);
+    }
+
+    function getAaveVariableBorrowAPY(address _asset) public view returns (uint256) {
```
MiksuJak, line 74: Since oracle works for one asset only, why do we pass it here? Cannot we just use the contract variable? Also, cannot we make it internal? (We can still make a test version where we override it and make it public. Also it would be really nice to have this function tested on mainnet fork.)

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,163 @@
...
+     * depends on if it is filled up and the value of timeToCover parameter.
+     * @param timeToCover For how much time average should be calculated.
+     * @return Average apy.
+     */
+    function calculateAverageAPY(uint256 timeToCover) public view returns (uint256) {
+        require(
+            1 days <= timeToCover && timeToCover <= 365 days,
+            "AaveBaseRateOracle: Expected amount of time in range 1 to 365 days"
+        );
+        // estimate how much buffer we need to use
```
MiksuJak, line 126: In my opinion, variables are named well enough so we don't need this comment :)

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,163 @@
...
+        uint16 bufferSizeNeeded = uint16(timeToCover.div(cooldownTime));
```
MiksuJak, line 127: I'm a little concerned about what may happen, if `update` gets called with a one day delay, because of any random reason. Wouldn't we then need one more, update than we actually have stored? And wouldn't then the starting point be 0 (if we haven't updated the whole buffer yet)?

I think I would feel much more secure, if we passed `numberOfUpdates` instead of `timeToCover` to this function.

skalermo: Thanks, I like this idea of using `numberOfUpdates` instead of `timeToCover`. I'm going to implement it.

contracts/truefi2/AaveBaseRateOracle.sol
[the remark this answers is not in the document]
barrutko: In mocks, we can override methods utilizing `bufferSize()` and use some new constant added to the mock.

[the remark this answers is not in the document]
yuchenlintt: Sounds good, let's keep this then.

contracts/truefi2/AaveBaseRateOracle.sol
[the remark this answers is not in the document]
yuchenlintt: Sorry, my bad in mistaking how this works. Let's delete lastValue and go with a right Riemann sum.

contracts/truefi2/AaveBaseRateOracle.sol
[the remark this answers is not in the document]
yuchenlintt: Yeah, you're right. That's an embarassing mistake from my end :)

contracts/truefi2/AaveBaseRateOracle.sol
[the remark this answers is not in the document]
skalermo: Starting to work on that right away.

test/truefi2/AaveBaseRateOracle.test.ts
```diff
@@ -0,0 +1,268 @@
...
+        await updateBufferRightAfterCooldown(aaveBaseRateOracle)
+        // Aave variable borrow apys (percents): 1.0, 2.0, 3.0 probed with 1 day interval
+        // Expected value is (1.0 + 2.0 + 3.0) / 3 = 2.0
+        expect(await aaveBaseRateOracle.calculateAverageAPY(4)).to.be.closeTo(BN(2_0000), 1)
+      })
+    })
+
+    describe('getWeeklyAPY', () => {
+      describe('returns correct value if', () => {
+        it('prices grows for last 3 days', async () => {
```
yuchenlintt, line 194: Please fix weekly tests to calculate over 7 days instead of 6. Setting outlier values on the edge cases (e.g., 100 or 10000 for day 0 or 1) would help catch off by one errors.

yuchenlintt (changes requested):
Thanks for the changes! The code looks much simpler, I like it.

I read through the test cases. They looked pretty good until I noticed that your `getWeeklyAPY()` tests average rates over 6 days instead of 7. So it looks like we have an off-by-one error (see comment below for a proposed fix).

contracts/truefi2/AaveBaseRateOracle.sol
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * @param numberOfValues How many values of totalsBuffer should be involved in calculations.
+     * @return Average apy.
+     */
+    function calculateAverageAPY(uint16 numberOfValues) public view returns (uint256) {
+        require(numberOfValues > 1, "AaveBaseRateOracle: Number of values should be greater than 1");
+        require(numberOfValues <= bufferSize(), "AaveBaseRateOracle: Number of values is limited by buffer size");
+
+        uint16 _currIndex = totalsBuffer.currIndex;
+        uint16 startIndex = (_currIndex + bufferSize() - numberOfValues + 1) % bufferSize();
```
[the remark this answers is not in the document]
skalermo: Thank you for your notes! I had similar thoughts in mind but wasn't sure if we want to have extra field in the buffer.

yuchenlintt (approved):
@skalermo Thanks for working on this time-averaging oracle over the past few months. This has been a large (and understandably frustrating) effort, but I think the code has improved tremendously and the result achieves our goals admirably.

I'm approving but will let someone else look this over a final time before merging.

MiksuJak:
We should integrate the oracle with already merged spot base rate oracle, before merging (#785 ). Fortunately it would not change any logic in this contract, just the way we get the rate to store.

MiksuJak (changes requested):
Integrate with #785 spot base rate oracle

skalermo:
Functionality introduced in this PR was splitted into separate PRs #785 and #816. Closing in favor of those two.

test/truefi2/AaveBaseRateOracle.test.ts
```diff
@@ -0,0 +1,288 @@
...
+      it('spot apy value has no impact on average apy', async () => {
+        await mockAaveLendingPool.mock.getReserveData.returns(...mockReserveData(1))
+        await updateBufferRightAfterCooldown(aaveBaseRateOracle)
+        await mockAaveLendingPool.mock.getReserveData.returns(...mockReserveData(2))
+        await updateBufferRightAfterCooldown(aaveBaseRateOracle)
+        await mockAaveLendingPool.mock.getReserveData.returns(...mockReserveData(3))
+        await updateBufferRightAfterCooldown(aaveBaseRateOracle)
+        // Aave variable borrow apys (percents): 1.0, 2.0, 3.0 probed with 1 day interval
+        // Expected value is (1.0 + 2.0 + 3.0) / 3 = 2.0
+        expect(await aaveBaseRateOracle.calculateAverageAPY(3)).to.eq(2_0000)
```
MiksuJak, line 207: 20_000

MiksuJak, line 207: 2e4

PR #784, opened 2021-07-29
Feature/reward borrower

jpalmer-trust:


test/truefi2/BorrowerDistributor.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      expect(await borrowerDistributor.rewardRate()).to.equal(REWARD_RATE * 2)
+    })
+  })
+
+  describe('distribute', () => {
+    it('must be called by lender', async () => {
+      await expect(borrowerDistributor.distribute(loan.address)).to.be.revertedWith('Only lender can call this')
+    })
+  })
+})
```
yuchenlintt, line 78: Let's also add tests to make sure:
1) the reward amount is correct, and
2) this distributor works with tokens that have different decimals (e.g., TestUSDCToken).

yuchenlintt (changes requested):
Thanks! Two changes:
1) nit: let's reorder some of the code.
2) Let's add some tests to make sure the reward amounts are correct.

PR #785, opened 2021-07-29
Inroduce spot value base rate oracle

MiksuJak:


yuchenlintt:
I think this approach shows promise for diversifying our base rate over multiple protocols (Aave, Compound, etc.) and removing protocol-specific code from #783, which should ideally handle only time-averaging.

We do still need #783's time-weighted average. Time-weighting smooths out marketwide movements over a reasonable window (e.g., 30 days) so borrowers can't lock in a low 90 day fixed rate during an hour when all protocols have dropped their spot rates simultaneously.

We can rename this contract something like `SpotBaseRateOracle` and the other one in #783 `TimeAveragedBaseRateOracle`. Then TABRO would call SBRO each time it needs a spot rate. If we do this, let's merge the current version of #783 first before renaming/refactoring.

Q: what happens if one of the weight sources starts consistently failing/reverting? Is there a way to make this fault-tolerant and average the remaining sources?

barrutko (commented):
Let's remove weights for now and make it possible to pass an asset as an argument.

barrutko (approved):

PR #793, opened 2021-08-02
Audit SAFU L4: Declare liquidate() + tokenBalance() in ISAFU

yuchenlintt:


MiksuJak:
I guess we have to make a decision, which may affect all other interfaces. Until now, our contracts' interfaces simply included declarations of functions, that are used externally in other contracts. Those 2 are not used anywhere. If we decide, that every non internal function should be included in the interface, then we need to change all of the other interfaces.

PR #794, opened 2021-08-02
Audit SAFU L5: Emit event in swap()

yuchenlintt:


MiksuJak:
We should write some tests for `swap()` function.

MiksuJak:
Also since this PR modifies `swap()`, let's make it external instead of public.

hal909 (approved):

vanruch (approved):

PR #800, opened 2021-08-03
Correctly calculate interest on rebucket

MiksuJak:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -366,7 +369,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
             usedBucketsBitmap &= ~(uint256(1) << bucketNumber);
         }
         bucket.totalBorrowed = bucket.totalBorrowed.sub(borrowed[pool][borrower]);
-        totalBorrowerInterest = _interest(pool, bucket, borrower);
+        totalBorrowerInterest = _interest(pool, bucket, borrower).add(totalPaidInterest[pool][borrower]);
```
yuchenlintt, line 372: The `_interest()` implementation subtracts `totalPaidInterest[pool][borrower]`, so if we want to track total interest for a borrower I think we should factor out a `_totalBorrowerInterest()` function instead. Both `_interest()` and `_takeOutOfBucket()` would call that.

yuchenlintt (changes requested):

contracts/truefi2/TrueCreditAgency.sol
barrutko: Shouldn't tests catch that? ☝️😅

vanruch (approved):

PR #805, opened 2021-08-04
Add grace period after repayment due date, add borrower status

vanruch:


test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -453,6 +471,13 @@ describe('TrueCreditAgency', () => {
         .to.be.revertedWith('TrueCreditAgency: Borrower has credit score below minimum')
     })
 
+    it('fails is borrower has missed the repay time', async () => {
```
skalermo, line 474:
```suggestion
    it('fails if borrower has missed the repay time', async () => {
```

test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -512,6 +537,36 @@ describe('TrueCreditAgency', () => {
...
   })
 
+  describe('Status', () => {
+    enum Status {Ineligible, OnHold, Eligible}
+
+    beforeEach(async () => {
+      await creditAgency.allowBorrower(borrower.address, YEAR)
+    })
+
+    it('returns Eligible when account has not any missed payments', async () => {
```
skalermo, line 547: Maybe better 'has no missed payments'.

vanruch:
```suggestion
    it('returns Eligible when account has no missed payments', async () => {
```

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -291,7 +308,7 @@ contract TrueCreditAgency is UpgradeableClaimable {
         return _interest(pool, bucket, borrower);
     }
 
-    function borrow(ITrueFiPool2 pool, uint256 amount) external onlyAllowedBorrowers {
+    function borrow(ITrueFiPool2 pool, uint256 amount) external onlyAllowedBorrowers onlyEligible(pool) {
```
barrutko, line 311: Not sure if we need to move it to modifier? Why not put it with other `require`s?

barrutko (approved):

contracts/truefi2/TrueCreditAgency.sol
vanruch: agree, the revert message is too context dependant on borrow

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -356,6 +369,19 @@ contract TrueCreditAgency is UpgradeableClaimable {
         return borrowed[pool][borrower].add(interest(pool, borrower));
     }
 
+    function status(ITrueFiPool2 pool, address borrower) public view returns (Status) {
+        if (nextInterestRepayTime[pool][borrower] == 0) {
+            return Status.Eligible;
+        }
+        if (nextInterestRepayTime[pool][borrower].add(gracePeriod) < block.timestamp) {
```
hal909, line 376: This should return onHold

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -356,6 +369,16 @@ contract TrueCreditAgency is UpgradeableClaimable {
         return borrowed[pool][borrower].add(interest(pool, borrower));
     }
 
+    function status(ITrueFiPool2 pool, address borrower) public view returns (Status) {
+        if (nextInterestRepayTime[pool][borrower] == 0) {
+            return Status.Eligible;
```
yuchenlintt, line 374: Shouldn't this be Status.Ineligible? And the condition should be nextInterestRepayTime + gracePeriod < block.timestamp, otherwise we aren't even using the new gracePeriod variable.

Let's make a follow-up PR that fixes these and adds tests to catch it.

yuchenlintt: I just reviewed this line again.

I see now that we don't initialize `nextInterestRepayTime` when a borrower is allowed. This means it defaults to 0, which creates an exception to our otherwise simple rule of comparing `block.timestamp` to `nextInterestRepayTime`. For simplicity, I think the correct resolution to this would be to initialize nextInterestRepayTime to either `type(uint256).max` or `block.timestamp + 30 days` when we call allowBorrower. This goes along with the `borrowerAllowedUntilTime` that we set when allowing a borrower.

vanruch: We can't initialize it with `block.timestamp + 30 days` because it would mean that if a borrower hasn't borrowed anything in 30 days after whitelisting, they cannot borrow anymore. I'm not sure about INT_MAX as well because when we will check grace periods, this may lead to overflows

vanruch: We can simply assume that if nextInterestRepayTime is 0 that means that the borrower has not borrowed anything yet and therefore was not late with payments. But this gave me an idea of another problem: if someone repays everything and more than a month passes, they will be unable to take a loan unless they do repay of 0 before that.

yuchenlintt: Yeah, I also thought of the 30 day expiry and would prefer INT_MAX for the same reason.

How about INT_MAX - 1_000 days then? That timestamp is still way after the heat death of the universe, and I doubt we would ever update gracePeriod to be almost 3 years. Or we could compare `block.timestamp - gracePeriod` against `nextInterestRepayTime` to avoid overflow concerns in the latter.

yuchenlintt: Your comment reminds me: we should reset `nextInterestRepayTime` (to either 0 or INT_MAX, depending on our decision in this thread) if a borrower repays in full and ends up with a `currentDebt` of 0. Otherwise the borrower would need to call `repay(pool, 0)` every 30 days.

I'll add a task for this.

vanruch: So the problem with initialising repay time is that we track different due times for each pool. And if we could iterate by each pool when whitelisting the borrower, we cannot iterate by each borrower in case we will ever add a new pool. Looks like this change will introduce much more complexity than reduce it. Let's keep it as it is, ok?

yuchenlintt: You're right, I hadn't thought of this edge case. The adding new pools concern would certainly make initializing to INT_MAX more complicated than checking a special case of zero here.

Let's keep it as is :)

yuchenlintt: So we don't lose the other action item from this thread:

We still need to add an automatic ineligibility case for block.timestamps after `nextInterestRepayTime + gracePeriod`, otherwise the `gracePeriod` isn't used for its intended purpose.

PR #806, opened 2021-08-04
Fix low/medium severity issues

jpalmer-trust:
- [MultiFarm M2] Introduce local variable for tokens.length in claim function
- [LinearTrueDistributor L2] Mark internally unused functions 'external'
- Update Grace Period for fixed term loans to 3 days

### DIFF
--- a/contracts/truefi/distributors/LinearTrueDistributor.sol
+++ b/contracts/truefi/distributors/LinearTrueDistributor.sol
@@ -137,7 +137,7 @@ contract LinearTrueDistributor is ITrueDistributor, Ownable {
      * Distributes remaining funds before withdrawing
      * Ends current distribution
      */
-    function empty() public override onlyOwner {
+    function empty() external override onlyOwner {
         distribute();
         distributed = 0;
         totalAmount = 0;
@@ -148,7 +148,7 @@ contract LinearTrueDistributor is ITrueDistributor, Ownable {
      * @dev Change amount of tokens distributed daily by changing total distributed amount
      * @param dailyDistribution New daily distribution
      */
-    function setDailyDistribution(uint256 dailyDistribution) public onlyOwner {
+    function setDailyDistribution(uint256 dailyDistribution) external onlyOwner {
         distribute();
         uint256 timeLeft = distributionStart.add(duration).sub(block.timestamp);
         if (timeLeft > duration) {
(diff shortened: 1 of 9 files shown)


barrutko (changes requested):
As this PR addresses 3 separate things, I'll address them separately:
1. `public` -> `external`
- Looks good
2. Extracting `tokens.length` value
- I'd name the variable `tokensLength` instead of `length`
- Not sure if we need/should do this for `i`. My intuition tells me, it's not needed. Also the audit doesn't mention that. But I'm not saying it's definitely not needed - it might be a good idea. Could you please check `gasUsed` for both versions - with extracting `i` and without?
3. `LAST_MINUTE_PAYBACK_DURATION`
- Tests update is missing (they are failing for a good reason)

Also for the future. Here we have fixes of two separate issues and one non-audit-related tweak. It would be nice to have it as 3 separate PRs. Or at least 2 - one for fixes, one for constant value change.

jpalmer-trust:
@barrutko Surprisingly, using `i` inside for loop is cheaper. It's 228404 within for loop and 228409 outside of loop.
Also, the initial solution is even cheaper ... it's only 228399.
Probably because the tokens array contains only one token?

jpalmer-trust:
image

jpalmer-trust:
Testing with 100 tokens.
Result with Original Solution: 1768263
Result with both variables defined outside:  1770071
Result with only token length defined outside: 1768268

jpalmer-trust:
image

barrutko:
Cool, looks good 👍 
IMO we should keep `i` in the `for(;;)` and `tokensLength` outside. Extracting `tokensLength` improves readability and costs only flat 5 gas more. Moving `i` outside can cost way more and creates unusual code factoring, so I'd keep it inside.

We still need to:
- rename `length` to `tokensLength`
- 🚨 **fix LoanToken tests** 🚨

MiksuJak (approved):

barrutko (approved):
👍

skalermo (approved):

yuchenlintt:
LGTM

> Also for the future. Here we have fixes of two separate issues and one non-audit-related tweak. It would be nice to have it as 3 separate PRs. Or at least 2 - one for fixes, one for constant value change.

@jpalmer-trust I agree with Bartek, these changes should have been split into three separate PRs. Next time!

Extracting `uint256 tokensLength = tokens.length` was a minor positive refactor regardless of gas expenditure (even though gas was the reason brought up by auditors). You probably didn't need to take the extra effort to profile since we were just addressing audit comments on a minor refactor.

Thanks for the changes!

PR #811, opened 2021-08-09
Extract constants in credit agency

vanruch:


barrutko (approved):

yuchenlintt:
Let's follow up with a PR to global replace instances of `1e27`, `1e23`, and `10000` with this new `ADDITIONAL_PRECISION` and a `BASIS_RATIO = 10000` constant.

skalermo:
> Let's follow up with a PR to global replace instances of `1e27`, `1e23`, and `10000` with this new `ADDITIONAL_PRECISION` and a `BASIS_RATIO = 10000` constant.

@yuchenlintt I think this is a great idea! However I find these namings rather long. What do you think about using some well known concepts like `WAD` and `RAY`? We could even introduce new ones: for example basis ratio could be `BAS = 10000`. And `1e23` could be `RAS` or `BAY`.

I am eager to know your opinion and opened to suggestions.

yuchenlintt:
BASIS_RATIO = 10_000 appears to be the standard name we've adopted in other contracts, so I wouldn't change that. But I'd be up for naming WAD = 1e27. And 1e23 isn't native, it's derived from WAD / BASIS_RATIO so maybe we should just write WAD / BASIS_RATIO to make that relationship clear. The compiler should take care of turning it into a compile time constant for us.

PR #812, opened 2021-08-10
Add PoolFactory deprecatePool() and removePool() functions

yuchenlintt:
We have to upgrade PoolFactory to allow deprecation of legacy pools before we can call `createPool()` for TUSD. This is because `createPool()` reverts if the `onlyNotExistingPools()` check fails.

I searched through the repo for possible contract side effects and breakages of this PR. After deploying this change we'll have the following sequence of states:
1. Before legacy pool deprecation: No contract behavior changes.
2. After deprecation but before new pool creation: Other contracts in the repo only ever call `isPool(TUSD)`. This never touches the `pool` mapping and hence won't fail even on the legacy pool. End users who call `pool(token)` will temporarily get `address(0)`, but that doesn't break anything.
3. After new pool creation but before removal: Ditto with 1. End users who call `pool(token)` will get the new pool address.
4. After removal: `isPool(legacy TUSD)` will fail, so TrueLender2 and LoanFactory2 will refuse to create, fund, distribute, or transfer legacy tfTUSD loans.

2 and 3 should be called back-to-back when creating a new tfTUSD pool.

4 should ONLY be called after the last legacy tfTUSD loan has been repaid. Otherwise it could lock funds in legacy tfTUSD loans.

### DIFF
--- a/contracts/truefi2/PoolFactory.sol
+++ b/contracts/truefi2/PoolFactory.sol
@@ -146,6 +146,21 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
     }
 
     /**
+     * @dev Deprecate a pool from token lookups without removing it from the factory.
+     * Calling this function allows owner to create a replacement pool for the same token.
+     */
+    function deprecatePool(ITrueFiPool2 legacyPool) external onlyOwner {
+        pool[address(legacyPool.token())] = address(0);
+    }
+
+    /**
+     * @dev Remove a pool from the factory regardless of deprecation status.
+     */
+    function removePool(ITrueFiPool2 legacyPool) external onlyOwner {
+        isPool[address(legacyPool)] = false;
+    }
+
+    /**
      * @dev Create a new pool behind proxy. Update new pool's implementation.
      * Transfer ownership of created pool to Factory owner.
      * @param token Address of token which the pool will correspond to.


MiksuJak:
Very good work! I also checked the dependencies myself as a review. Just to clarify, I think that Lender's `reclaim()` would still work after calling `removePool()`, but SAFU interactions wouldn't, thus as You said we should call it after all legacyPool loans are reclaimed. I'm only wondering if we can also delete `addLegacyPool()` function in this PR. Or maybe we could delete it along with those 2 functions, once fresh pool is established and the legacy one is deleted. I think it would be just very nice, to have this contract as clean TrueFi 2.0, without any old TrueFi traces.

MiksuJak (approved):

yuchenlintt:
> I'm only wondering if we can also delete `addLegacyPool()` function in this PR. Or maybe we could delete it along with those 2 functions, once fresh pool is established and the legacy one is deleted.

If only we could do this easily. The original version of this PR deleted `addLegacyPool()` (see b2b32f0 ), but that would have required refactoring/deleting all TrueFiPool + TrueLender2 tests since we call `addLegacyPool()` beforeEach test setup. I decided to keep the TFP tests around until TFP has been removed. As you say, we can remove `addLegacyPool()` + `deprecatePool()` + `removePool()` together once TFP is gone.

vanruch (approved):
We'll have a chance to wipe all legacy pool-relaited code after old TUSD pool is retired. Let's leave it as it is for now

MiksuJak:
One more thing came to my mind. We could write some integration tests to check, if funds wouldn't get locked after pool deprecation. To be honest I wouldn't expect anything to break, but maybe it is worth doing just in case.

barrutko:
Shouldn't these two state alternations be performed in one call? We will have an inconsistent state between calling these two methods.

hal909 (approved):

PR #813, opened 2021-08-10
Allow borrowers to create 180 day term loans if credit score is at least 200

MiksuJak:


MiksuJak:
I am really open to suggestions for variable names

test/truefi2/TrueLender2.test.ts
```diff
@@ -310,13 +399,42 @@ describe('TrueLender2', () => {
         await expect(lender.connect(borrower).fund(loan1.address))
           .to.be.revertedWith('TrueLender: Loan risk is too high')
       })
+
+      it('loan term exceeds max term', async () => {
+        await rater.connect(borrower).submit(loan1.address)
+        await rater.yes(loan1.address)
+        await timeTravel(7 * DAY + 1)
+        await lender.setLongLoanMinTerm(DAY * 364)
+        await lender.setMaxLoanTerm(DAY * 364)
```
barrutko, line 408: Why 364 not 365?

MiksuJak: Because loan1 has a 1 year term and we want max term to be lower for this test

skalermo: I think it would be more obvious if it was `DAY * 365 - 1`

contracts/truefi2/TrueLender2.sol
```diff
@@ -257,7 +332,10 @@ contract TrueLender2 is ITrueLender2, UpgradeableClaimable {
 
         uint256 amount = loanToken.amount();
         (uint256 start, uint256 no, uint256 yes) = ratingAgency.getResults(address(loanToken));
+        uint256 term = loanToken.term();
 
+        require(term <= maxLoanTerm, "TrueLender: Loan's term is too long");
+        require(_isCredibleForTerm(term), "TrueLender: Credit score is too low for loan's term");
```
barrutko, line 338: Let's remove underscore

contracts/truefi2/TrueLender2.sol
```diff
@@ -257,7 +332,10 @@ contract TrueLender2 is ITrueLender2, UpgradeableClaimable {
 
         uint256 amount = loanToken.amount();
         (uint256 start, uint256 no, uint256 yes) = ratingAgency.getResults(address(loanToken));
+        uint256 term = loanToken.term();
 
+        require(term <= maxLoanTerm, "TrueLender: Loan's term is too long");
```
barrutko, line 337: Let's wrap this with function like the rest of the requirements here.

barrutko (approved):

yuchenlintt (commented):
The logic is straightforward and LGTM.

I think the naming is mostly okay too. I would probably choose `longTermLoanTermThreshold` to match `longTermLoanScoreThreshold` (but I'm sleepy enough now that I don't know if this name would still make sense if I were more awake).

PR #816, opened 2021-08-10
Introduce time-averaged base rate oracle

skalermo:


contracts/truefi2/SpotBaseRateOracle.sol
```diff
@@ -13,7 +13,7 @@ contract SpotBaseRateOracle {
         aaveLendingPool = _aaveLendingPool;
     }
 
-    function getRate(address asset) external view returns (uint256) {
+    function getRate(address asset) public view returns (uint256) {
```
MiksuJak, line 16: Why public?

contracts/truefi2/TimeAveragedBaseRateOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * @dev Update the totalsBuffer:
+     * Gets current variable borrow apy from a collateralized lending protocol
+     * for chosen asset and writes down new total running value.
+     * If the buffer is filled overwrites the oldest value
+     * with a new one and updates its timestamp.
+     */
+    function update() public offCooldown {
+        uint16 _currIndex = totalsBuffer.currIndex;
```
MiksuJak, line 83: I don't think we need `_currIndex`

contracts/truefi2/TimeAveragedBaseRateOracle.sol
```diff
@@ -0,0 +1,149 @@
...
+     * @return Average apy.
+     */
+    function calculateAverageAPY(uint16 numberOfValues) public view returns (uint256) {
+        require(numberOfValues > 0, "TimeAveragedBaseRateOracle: Number of values should be greater than 0");
+        require(numberOfValues < bufferSize(), "TimeAveragedBaseRateOracle: Number of values should be less than buffer size");
+
+        uint16 _currIndex = totalsBuffer.currIndex;
+        uint16 startIndex = (_currIndex + bufferSize() - numberOfValues) % bufferSize();
+
+        require(totalsBuffer.timestamps[startIndex] > 0, "TimeAveragedBaseRateOracle: There are fewer values stored than required");
```
vanruch, line 122: So the problem is: we can't use the oracle for a month after deployment. Maybe we should startIndex to 0 in this case? This will only be happening during the first month anyway.

test/truefi2/TimeAveragedBaseRateOracle.test.ts
```diff
@@ -0,0 +1,289 @@
...
+          await updateBufferRightAfterCooldown(timeBaseRateOracle)
+          await updateBufferRightAfterCooldown(timeBaseRateOracle)
+          await mockSpotOracle.mock.getRate.withArgs(asset.address).returns(300)
+          await updateBufferRightAfterCooldown(timeBaseRateOracle)
+          await mockSpotOracle.mock.getRate.withArgs(asset.address).returns(400)
+          await updateBufferRightAfterCooldown(timeBaseRateOracle)
+          // Borrow apys (percents): 1.0, 1.0, 1.0, 1.0, 2.0, 3.0, 4.0 probed with 1 day interval
+          // Expected avg apy is (1.0 * 3 + 2.0 * 2 + 3.0 + 4.0) / 7 = 2.0
+          // Expected weekly apy is 2.0
+          expect(await timeBaseRateOracle.calculateAverageAPY(7)).to.eq(2_00)
```
MiksuJak, line 230: 200 instead of 2_00

MiksuJak (approved):

contracts/truefi2/TimeAveragedBaseRateOracle.sol
yuchenlintt: Hmm. Yeah, I'm okay with that. We would be exposed to a little more volatility in the first month, but at least the oracle would work instead of reverting.

yuchenlintt: We can check this in and decide whether to make that change in a separate PR. Most of this PR is ready and doesn't need to wait on this decision.

yuchenlintt (approved):

PR #817, opened 2021-08-10
make strategy more conservative

hal909:


barrutko (commented):
Maybe we also want to add some sort of a setter, so we won't need more upgrades in the future?

yuchenlintt (changes requested):
I agree with Bartek, let's make this an owner-settable variable since we're already interested in changing it once.

yuchenlintt (approved):

contracts/truefi2/strategies/CurveYearnStrategy.sol
```diff
@@ -28,8 +28,6 @@ contract CurveYearnStrategy is UpgradeableClaimable, ITrueStrategy {
 
     // Number of tokens in Curve yPool
     uint8 public constant N_TOKENS = 4;
-    // Max slippage during the swap
-    uint256 public constant MAX_PRICE_SLIPPAGE = 200; // 2%
```
barrutko, line 32: Are we sure that we can remove constant from the memory? @vanruch?

barrutko: Nevermind.

yuchenlintt: My understanding is that constants do not affect contract storage. See the Solidity [docs](https://docs.soliditylang.org/en/latest/contracts.html#constants) for reference:

> It is also possible to define constant variables at the file level.
>
> The compiler does not reserve a storage slot for these variables, and every occurrence is replaced by the respective value.

barrutko:
Not sure if such conservative requirements will ever pass, though.

yuchenlintt:
> Not sure if such conservative requirements will ever pass, though.

Yeah I agree. Knowing how frequently our integration tests already fail, I don't know if setting the maxPriceSlippage to 0.75% would ever succeed.

Regardless, this PR should give us the option to adjust down _or up_, as needed.

vanruch:
Integration tests passed on first try, LGTM

PR #821, opened 2021-08-11
Add secured rate oracle integration

MiksuJak:


test/truefi2/TrueCreditAgency.test.ts
```diff
@@ -1058,6 +1096,22 @@ describe('TrueCreditAgency', () => {
...
+
+      expect(await creditAgency.interest(tusdPool.address, borrower.address)).to.be.closeTo(parseUSDC(10), 1e5) // 10 = 10% * 36500 / 365
+
+      await mockSpotOracle.mock.getRate.withArgs(tusd.address).returns(1000) // this will increase average by 1%
+      await updateRateOracle(tusdBaseRateOracle, DAY, provider)
+      await creditAgency.poke(tusdPool.address)
+      expect(await creditAgency.interest(tusdPool.address, borrower.address)).to.be.closeTo(parseUSDC(20), 1e5) // 20 = 2*(10% * 36500 / 365)
+
+      await timeTravel(DAY)
+      expect(await creditAgency.interest(tusdPool.address, borrower.address)).to.be.closeTo(parseUSDC(31), 1e5) // 21 = 2*(10% * 36500 / 365) + (11% * 36500 / 365)
```
skalermo, line 1112:
```suggestion
      expect(await creditAgency.interest(tusdPool.address, borrower.address)).to.be.closeTo(parseUSDC(31), 1e5) // 31 = 2*(10% * 36500 / 365) + (11% * 36500 / 365)
```

test/utils/updateRateOracle.ts
```diff
@@ -0,0 +1,20 @@
...
+import { DAY, timeTravelTo } from '.'
+
+export const updateRateOracle = async (oracle: TimeAveragedBaseRateOracle | TestTimeAveragedBaseRateOracle, cooldown: number, provider: MockProvider) => {
+  const [, timestamps, currIndex] = await oracle.getTotalsBuffer()
+  const newestTimestamp = timestamps[currIndex].toNumber()
+  await timeTravelTo(provider, newestTimestamp + cooldown - 1)
+  await oracle.update()
+}
+
+export const weeklyFillBaseRateOracles = async (tusdOracle: TimeAveragedBaseRateOracle, usdcOracle: TimeAveragedBaseRateOracle, provider: MockProvider) => {
```
skalermo, line 12: I think this `weeklyFillBaseRateOracles()` function is too specific to be here. And it is only used in `setupTruefi2.ts`. Consider moving it there or making it more generic.

test/utils/updateRateOracle.ts
```diff
@@ -0,0 +1,20 @@
+import { TestTimeAveragedBaseRateOracle, TimeAveragedBaseRateOracle } from 'contracts'
+import { MockProvider } from 'ethereum-waffle'
+import { DAY, timeTravelTo } from '.'
+
+export const updateRateOracle = async (oracle: TimeAveragedBaseRateOracle | TestTimeAveragedBaseRateOracle, cooldown: number, provider: MockProvider) => {
```
skalermo, line 5: I like that this function is extracted here. However I just noticed one possible issue: when there is `timeTravel()` called independently from this function it is possible that `timeTravelTo()` here will bring us to the past thus causing an error. `timeTravelTo` only if the time we are traveling to is greater than `now`.

MiksuJak: I honestly thought of it, but I really think that the person writing tests should be aware of it, and simply call `oracle.update()` if the time has passed

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -317,17 +327,25 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     }
 
     function currentRate(ITrueFiPool2 pool, address borrower) external view returns (uint256) {
-        return _currentRate(riskPremium.add(utilizationAdjustmentRate(pool)), creditScoreAdjustmentRate(pool, borrower));
+        return
+            _currentRate(
+                riskPremium.add(utilizationAdjustmentRate(pool)).add(getSecuredRate(pool)),
+                creditScoreAdjustmentRate(pool, borrower)
```
barrutko, line 333: I'd propose a following order:
`securedRate.add(riskPremium).add(util...).add(credit...)`

barrutko: And then apply the same order in other places where sum stuff.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -317,17 +327,25 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
...
+    }
+
+    function getSecuredRate(ITrueFiPool2 pool) public view returns (uint256) {
+        return baseRateOracle[pool].getWeeklyAPY();
     }
 
     /**
      * @dev Helper function used by poke() to save gas by calculating partial terms of the total rate
-     * @param partialRate risk premium + utilization adjustment rate
+     * @param bucketCommonRate risk premium + utilization adjustment rate + secured rate
```
barrutko, line 343: Maybe just `bucketRate`?

MiksuJak:
> This PR is large enough that it needs to be split into manageable pieces, some of which we can check in very quickly:
> 
> * TABRO -> ITABRO interface
> * partialRate -> poolRate naming (maybe make poolRate() an internal view function so the logic is less duplicated? you can still pass its returned value around for gas savings)
> * add ITABRO to TCA
> * setupTrueFi2 test changes

Thanks for the poolRate function idea.

As for splitting the PR into smaller pieces, the current state follows the rules we established:
1. It took less than a day to prepare
2. One board task = one PR (maybe it could have been split, but I took it as I found it)

I personally do not find the changes hard to read, but I may be biased because I'm the one who wrote them 😅. Since the changes are already implemented, I feel like dividing this process into smaller steps would only add some excessive and would lead to the same results anyway.

yuchenlintt:
Fair enough, this PR isn't _too_ large :)

I think in this kind of situation the correct action is to split the board task yourself if it looks like it doesn't fit in one PR. I'm concerned about the common situation where we're holding up PRs that touch multiple files due to changes we want to make in 1-2 of them. Or if we decide to do an unrelated refactor in an otherwise ready-to-merge PR, that could hold up the rest of the PR.

In this case, I've read through and want to approve:
- the TABRO -> ITABRO change
- the partialRate -> poolRate change (with comments)
- most of (adding ITABRO to TCA)
but I haven't reviewed and don't want to approve this yet:
- test changes

MiksuJak:
@yuchenlintt I updated the PR with `poolRate()` function refactoring. Thanks for the feedback 😁. Once You review the rest of the PR feel free to merge it. I believe having this, would ease our testing in tomorrow's tasks, because adding any new variable to the rate requires some test rewriting (as seen above).

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -317,17 +327,25 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     }
 
     function currentRate(ITrueFiPool2 pool, address borrower) external view returns (uint256) {
-        return _currentRate(riskPremium.add(utilizationAdjustmentRate(pool)), creditScoreAdjustmentRate(pool, borrower));
+        return
+            _currentRate(
+                securedRate(pool).add(riskPremium).add(utilizationAdjustmentRate(pool)),
```
yuchenlintt, line 332:
```suggestion
                _poolRate(pool),
```

yuchenlintt (approved):
The tests LGTM. Thanks for this making these changes!

I only have one (obvious) suggestion, so I'll commit that myself and merge this PR.

PR #823, opened 2021-08-11
Extract rater calculation from CreditAgency to a separate contract

vanruch:
This didn't simplify CreditAgency to an extent I hoped it will and will probably cost us some in gas cost. I have doubts if we should go on with this refactor

### DIFF
--- a/contracts/truefi2/TrueCreditAgency.sol
+++ b/contracts/truefi2/TrueCreditAgency.sol
@@ -78,8 +78,6 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
 
     ITrueFiCreditOracle public creditOracle;
 
-    mapping(ITrueFiPool2 => ITimeAveragedBaseRateOracle) public baseRateOracle;
-
     /**
      * This bitmap is used to non-empty buckets.
      * If at least one borrower with a score n has an opened credit line, the n-th bit of the bitmap is set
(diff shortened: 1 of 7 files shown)


barrutko (approved):
IMO looks good and nicely reduces the weight of TrueCreditAgency contract.

yuchenlintt (changes requested):
It's good that we're removing the contract state variables from TCA. This would at least help make this more future-compatible.

Could we also drop the delegated setters and rate views from TCA? We could call these directly from CLR.

IMO TCA public functions should be stripped down to:
- allow/status
- borrow
- repay/repayInFull/repayInterest
- poke
- etc. (I might've glossed over some other related functions)

I feel like anything related to viewing rates or changing rate params should belong as much as possible in CLR, though I don't know how much this is possible.

Also CreditLineRate should be renamed becauses it's not just credit lines (it also helps calculate rates for fixed term loans). Something like TrueRateAdjuster?

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -136,11 +127,8 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     function initialize(ITrueFiCreditOracle _creditOracle, uint256 _riskPremium) public initializer {
         UpgradeableClaimable.initialize(msg.sender);
         creditOracle = _creditOracle;
-        riskPremium = _riskPremium;
-        creditAdjustmentCoefficient = 1000;
+        rater = new CreditLineRate(_riskPremium);
```
barrutko, line 130: Let's do dependency injection instead.

PR #824, opened 2021-08-12
Simplify + automate borrower eligibility logic

yuchenlintt:
This simplifies and automates borrower credit eligibility in TFCO. More specifically, this PR:
- moves credit eligibility Status from TCA to TFCO
- allows automatic Status transitions in TFCO: Eligible -> OnHold -> Ineligible* in TFCO
- reverts TCA's `borrowerAllowedUntilTime` timestamp mapping back to an `isBorrowerAllowed` boolean mapping

*With a trick we can still _de facto_ require `owner` to manually trigger Ineligible Status. This trick works by setting the `gracePeriod` variable to something like 10 years.

Reviewer note: I found it easier to rewrite `TrueFiCreditOracle.test.ts` from scratch, so comparing the diff in this file may not make sense.


contracts/truefi2/TrueFiCreditOracle.sol
```diff
@@ -75,9 +103,9 @@ contract TrueFiCreditOracle is ITrueFiCreditOracle, UpgradeableClaimable {
      * Scores are stored as uint8 allowing scores of 0-255
      */
     function setScore(address account, uint8 newScore) public onlyManager {
+        _setEligibleUntilTime(account, Math.max(eligibleUntilTime[account], block.timestamp.add(creditUpdatePeriod)));
```
MiksuJak, line 106: 👍 Nice that You handled the case when we want to set eligibility for a longer than standard period

test/truefi2/lines-of-credit/TrueFiCreditOracle.test.ts
```diff
@@ -27,137 +31,203 @@ describe('TrueFiCreditOracle', () => {
...
+      expect(await oracle.eligibleUntilTime(borrower.address)).to.equal(timestamp.add(10 * DAY))
+      expect(await oracle.status(borrower.address)).to.equal(Status.Eligible)
     })
 
-    it('ineligibility and on hold set correctly', async () => {
-      expect(await oracle.onHold(firstAccount.address)).to.be.true
-      expect(await oracle.ineligible(firstAccount.address)).to.be.true
+    it('emits a proper event', async () => {
+      await expect(oracle.setEligibleForDuration(borrower.address, 10 * DAY))
+        .to.emit(oracle, 'EligibleUntilTimeChanged')
```
MiksuJak, line 188: We could also check `.withArgs()` here.
There is a trick to do it for time dependant storage:
`
const tx = oracle.setEligibleForDuration(borrower.address, 10 * DAY)
expect(tx).to.emit(oracle, 'EligibleUntilTimeChanged').withArgs(await oracle.status(borrower.address))
`

MiksuJak (approved):
You can add event arguments, as I commented. I approve this. I find solidity solutions very elegant.

test/truefi2/lines-of-credit/TrueFiCreditOracle.test.ts
yuchenlintt: Whoa this is really cool! Thanks for the tip, I hadn't known we could do tricks like this.

vanruch (changes requested):
I'm not completely sure we want it to behave like that tbh.
There's a problem with a danger of `Ineligible` status being set automatically which, I think, we wanted to avoid.
The grace period afaik isn't strictly connected with statuses, instead it's a time before we can call default and I'm not sure if we will always want it to be same for every LoC.

I'd like to hear @hal909 opinion on this

vanruch (approved):

PR #825, opened 2021-08-12
Reduce bucket interest calculation kludginess

yuchenlintt:
This PR adds a `_newInterestPerShare(bucket, timeNow)` view that simplifies and dedups a lot of the bucket interest calculations.


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -473,10 +471,6 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
         return Status.Eligible;
     }
 
-    /**
-     * @dev check if borrower score has been updated recently enough
-     * @return Whether block timestamp is less than last update + threshold
-     */
```
yuchenlintt, line 479: I know this docstring change is irrelevant, but prettier insisted that I delete the docstring so that `yarn lint` passes. I hope you don't mind, given this whole function is hopefully going away soon.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -422,35 +422,33 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
...
 
-        poolTotalInterest[pool] = poolTotalInterest[pool].add(
-            bucket.rate.mul(1e23).mul(bucket.totalBorrowed).mul(timeNow.sub(bucket.timestamp)).div(365 days)
-        );
-
-        bucket.cumulativeInterestPerShare = bucket.cumulativeInterestPerShare.add(
-            bucket.rate.mul(ADDITIONAL_PRECISION.div(10000)).mul(timeNow.sub(bucket.timestamp)).div(365 days)
-        );
+        uint256 newInterestPerShare = _newInterestPerShare(bucket, timeNow);
+        poolTotalInterest[pool] = poolTotalInterest[pool].add(bucket.totalBorrowed.mul(newInterestPerShare));
```
MiksuJak, line 426: In thatI'm concerned, that in this case we do multiplication after division.

vanruch: We have 10^27 additional precision, I think we should be fine with dividing by 31556952. BTW, maybe we want to extract 1e23/(seconds in year) as const as a small optimisation?

yuchenlintt: Hmm, that's a fair concern. But I think I'm not very worried about this numerical instability because we're still multiplying by 27 (well, actually 23) decimal points of precision inside the `_newInterestPerShare()` function. 23 decimal points of precision is pretty massive, about the same order of magnitude as counting the exact number of carbon atoms in 1 gram of diamond.

The loss of precision we get from dividing by 365 days is counteracted by the extra precision from `timeNow - bucket.timestamp` (probably a few hours?), which only takes away maybe 5 decimal points of precision.

contracts/truefi2/TrueCreditAgency.sol
vanruch: It's fine, just don't understand what's prettier's problem with that

vanruch (approved):

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: > BTW, maybe we want to extract 1e23/(seconds in year) as const as a small optimisation?

Not a bad idea, I'll go implement that now. If I recall correctly the Solidity compiler can optimize away simple arithmetic operations, so I might just drop the SafeMath for these constants.

MiksuJak (approved):

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: Huh, it looks like 10^23/365 days actually broke a test. I guess if we're dividing first by the seconds in a year, then integer division can truncate a lot.

Anyway, I'm splitting the division back up to regain precision. This fixed the failing test locally.

PR #827, opened 2021-08-12
Delete TrueFiPool2 exit()

yuchenlintt:


test/truefi2/TrueFiPool2.test.ts
```diff
@@ -120,12 +119,6 @@ describe('TrueFiPool2', () => {
     })
   })
 
-  it('cannot exit and join on same transaction', async () => {
-    const arbitrage = await new Pool2ArbitrageTest__factory(owner).deploy()
-    await tusd.transfer(arbitrage.address, parseEth(1))
-    await expect(arbitrage.joinExit(tusdPool.address)).to.be.revertedWith('TrueFiPool: Cannot join and exit in same block')
-  })
```
vanruch, line 127: I guess we still want to keep the test as well as the arbitrage contract

vanruch (changes requested):
I believe we only planned to remove exit() from legacy pool before doing a migration. @hal909 can you clarify?

vanruch (approved):

PR #836, opened 2021-08-12
Implement fixed term loan adjustment 2

skalermo:
I am really sorry for moving this matter from #831 here. At some point I noticed that fixing `LoanFactory2` tests polluted previous PR heavily and in fact even unnecessary as the changes doesn't need to affect LF2 yet. 

This PR should be much more readable and easy for review. I believe all important and relevant to this task changes were moved here.


yuchenlintt (approved):
Let's check this in. It looks much cleaner than before, thanks!

When we settle on the actual adjustment formula, it should be a simple one-liner code change and an update of the unit test values.

PR #863, opened 2021-08-16
Add a MultiFarm deploy script

yuchenlintt:


vanruch:
What do you think about having this in a separate file?

yuchenlintt:
> What do you think about having this in a separate file?

Yeah, that makes sense. The multifarm + distributor doesn't affect the core lending functionality, and it would help readability to make that independence clear. I'll go ahead and use this PR to move the multifarm to a new file.

This reminds me that we haven't been doing a good job of pruning deprecated contracts. I think after we delete the old farms + distributors, then `deploy/truefi2.ts` will be much smaller and more manageable.

Most of `deploy/truefi.ts` should probably also be deleted or migrated into `truefi2.ts`.

deploy/farms.ts
```diff
@@ -0,0 +1,60 @@
...
+  TrueMultiFarm,
+  TrustToken,
+} from '../build/artifacts'
+import { utils, BigNumber } from 'ethers'
+
+const DAY = 60 * 60 * 24
+
+const deployParams = {
+  mainnet: {
+    DISTRIBUTION_DURATION: 14 * DAY,
```
hal909, line 16: Can we make the distribution duration for multifarm 2 years? We don't want to have to reset this every 2 weeks

deploy/farms.ts
```diff
@@ -0,0 +1,60 @@
...
+    DISTRIBUTION_START: 1623952800,
+    // 200,000 per day for 14 days
+    STAKE_DISTRIBUTION_AMOUNT: BigNumber.from('280000000000000'),
+  },
+  testnet: {
+    DISTRIBUTION_DURATION: 180 * DAY,
```
hal909, line 22: Make this 2 years as well

deploy/farms.ts
yuchenlintt: Sorry I missed this. I had just copied over the existing values from `deploy/truefi2.ts`, from the last time we ran this script for farms.

yuchenlintt: Per offline discussion, also updating daily distribution to 330k TRU/day.

MiksuJak (approved):

vanruch (approved):

PR #869, opened 2021-08-17
Add strategy deploy script to CI

yuchenlintt:


yuchenlintt:
The strategy deploy script is failing on a `pool.token()` call in the CurveYearnStrategy `initialize()` function. For mainnet + ropsten we assume that `poolFactory.pool(token)` will return a real TrueFiPool2 because `poolFactory` was previously deployed and initialized, but that assumption fails for our CI deploy check, which does a dry run on kovan.

Closing this PR for now.

PR #870, opened 2021-08-17
Update deployments.json

MiksuJak:
Verified new contracts on Ropsten network:
1. LoanFactory2 implementation: https://ropsten.etherscan.io/address/0x1Be4462ca13266c92302983b8b3907BE0f25C409
2. Liquidator2 implementation: https://ropsten.etherscan.io/address/0x23f4DADFf7d69b8671a19DB21460563d99F9f0D2
3. SAFU implementation (UNVERIFIED): https://ropsten.etherscan.io/address/0x4C9E5914b91Bc035Bb632B9cfFd054c1EE9b040d


yuchenlintt:
Thanks!

In the future let's separate deploy script change PRs and deployments.json PRs (unless we're in a major crunch and all on a video call during a deployment). That way, there isn't a possibility of inconsistency caused by this kind of sequence:
1. deploy to ropsten/mainnet
2. code review change to deploy script
3. check in deploy script + outdated deployment in the same PR

MiksuJak:
> Thanks!
> 
> In the future let's separate deploy script change PRs and deployments.json PRs (unless we're in a major crunch and all on a video call during a deployment). That way, there isn't a possibility of inconsistency caused by this kind of sequence:
> 
> 1. deploy to ropsten/mainnet
> 2. code review change to deploy script
> 3. check in deploy script + outdated deployment in the same PR

Sure, I did not think of it, but what You say sounds like a good practice.

yuchenlintt:
I just checked and confirmed that SAFU is already included in `deploy/truefi2.ts`:

https://github.com/trusttoken/smart-contracts/blob/277642542b1f6d628a53baefccc9cc19ba6e3083/deploy/truefi2.ts#L103

We can't take it out of `deploy/truefi2.ts` because SAFU is needed to initialize `PoolFactory`, which is tightly coupled in the rest of the truefi2 deploy script. So I think we can't add a separate deploy script for SAFU like this.

If you take a look at the commit sequence in one of my previous deploys, you can see our current workaround for deploying a small piece of a larger deploy script:
https://github.com/trusttoken/smart-contracts/pull/855

1. comment out everything except for what's deployed
2. deploy + record to deployments.json
3. revert commit 1

As of commit 3, the PR diff against `dev` can be verified to only touch `deployments.json`, which provides a safety guarantee that the deployer didn't accidentally modify the larger deploy script.

I must admit I don't like this either. But until we can iron out the circular dependencies in contract initializers, we need a linear source of truth for what deployment order is safe. And the best way I see to achieve that is to have the tightly coupled deploys in a single file.

MiksuJak:
> I just checked and confirmed that SAFU is already included in `deploy/truefi2.ts`:
> 
> https://github.com/trusttoken/smart-contracts/blob/277642542b1f6d628a53baefccc9cc19ba6e3083/deploy/truefi2.ts#L103
> 
> We can't take it out of `deploy/truefi2.ts` because SAFU is needed to initialize `PoolFactory`, which is tightly coupled in the rest of the truefi2 deploy script. So I think we can't add a separate deploy script for SAFU like this.
> 
> If you take a look at the commit sequence in one of my previous deploys, you can see our current workaround for deploying a small piece of a larger deploy script:
> #855
> 
> 1. comment out everything except for what's deployed
> 2. deploy + record to deployments.json
> 3. revert commit 1
> 
> As of commit 3, the PR diff against `dev` can be verified to only touch `deployments.json`, which provides a safety guarantee that the deployer didn't accidentally modify the larger deploy script.
> 
> I must admit I don't like this either. But until we can iron out the circular dependencies in contract initializers, we need a linear source of truth for what deployment order is safe. And the best way I see to achieve that is to have the tightly coupled deploys in a single file.

Oh... sorry, I guess the task name mislead me. It is already done after all. I'll simply restrict this PR to updating deployments script then, since the changes are done.

yuchenlintt:
> Oh... sorry, I guess the task name mislead me. It is already done after all.

I think Hal didn't realize this was already done when he made the task. I should have caught this too.

I'm sorry you had to do duplicated work due to our planning mistakes.

> I'll simply restrict this PR to updating deployments script then, since the changes are done.

Sounds good!

yuchenlintt (approved):

hal909 (approved):

PR #874, opened 2021-08-17
Use compound borrowRatePerBlock function

jpalmer-trust:


contracts/truefi2/SpotBaseRateOracle.sol
```diff
@@ -13,10 +14,17 @@ contract SpotBaseRateOracle {
 
     /// @dev Aave lending pool contract
     IAaveLendingPool public aaveLendingPool;
+    uint256 public blocksPerDay;
```
MiksuJak, line 17: I'd be in favour of this being a constant

contracts/truefi2/SpotBaseRateOracle.sol
```diff
@@ -36,4 +44,10 @@ contract SpotBaseRateOracle {
         (, , , , uint128 currentVariableBorrowRate, , , , , , , ) = aaveLendingPool.getReserveData(asset);
         return uint256(currentVariableBorrowRate).div(1e23);
     }
+
+    function _getCompoundBorrowAPY(address asset) internal view returns (uint256 apy) {
+        uint256 borrowRatePerBlock = ICToken(asset).borrowRatePerBlock();
+        uint256 exponent = borrowRatePerBlock.mul(blocksPerDay).div(1e18).add(1);
```
MiksuJak, line 50: Because of the solidity math, this will always result in 1, which leads the following line to return always 0.

contracts/truefi2/SpotBaseRateOracle.sol
```diff
@@ -36,4 +44,10 @@ contract SpotBaseRateOracle {
         (, , , , uint128 currentVariableBorrowRate, , , , , , , ) = aaveLendingPool.getReserveData(asset);
         return uint256(currentVariableBorrowRate).div(1e23);
     }
+
+    function _getCompoundBorrowAPY(address asset) internal view returns (uint256 apy) {
```
MiksuJak, line 48: We'd rather avoid inplicit returns, just use `return` in the end of the function :)

MiksuJak (changes requested):
This task is mostly solidity math limitations problem.
At first let's establish, that our precision is `1e18`, meaning it is our new 1.
`borrowRatePerBlock.mul(blocksPerDay)` calculation, can remain the same and after adding to it 1e18 it is our daily ROI + 100%. 
Now the hardest part would be to take that number to the power of 365. An example of how to take a number with 1e18 precision to the power of 2:
```
a = 2e18
a*a = 4e36
hence a*a/1e18 = 4e18, which is the correct result
```

Here is what I suggest doing for optimization: 
1. Sign the daily rate to the memory variable.
2. Take it to the power of 2, then do it again, now we have `rate^4`. Assign it to another memory variable.
3. Take `(rate^4)^2 = rate^8` and assign to new variable.
4. Take `((rate^8)^2)^2 = rate^32` and assign to new variable.
5. Take `(rate^32)^2 = rate^64` and assign to new variable.
6. Take `((rate^64)^2)^2 = rate^256` and assign to new variable.
7. Why creating all those variables? Because now the only thing we need to do is multiply all of them, since `rate^256*rate^64*rate^32*rate^8*rate^4*rate^1 = rate^365`. We still need to remember about precision, so after every multiplication we need to divide by 1e18.
8. Now we have our result with 1e18 precision, so the only thing we need to do, is scale it to our 1e4 basis precision.

The second really important matter, is that we need tests to cover all of the cases and helper functions, that You create along the way. I'm open to conversation and any questions, that may occur :)

contracts/truefi2/SpotBaseRateOracle.sol
```diff
@@ -13,10 +14,17 @@ contract SpotBaseRateOracle {
 
     /// @dev Aave lending pool contract
     IAaveLendingPool public aaveLendingPool;
+    uint256 public blocksPerDay;
 
     /// @dev constructor which sets aave pool to `_aaveLendingPool`
     constructor(IAaveLendingPool _aaveLendingPool) public {
         aaveLendingPool = _aaveLendingPool;
+        blocksPerDay = 6500;
```
skalermo, line 22: Compound documentation says it is `6570`. Let's not deviate from it.

contracts/truefi2/SpotBaseRateOracle.sol
```diff
@@ -13,10 +14,17 @@ contract SpotBaseRateOracle {
...
+    }
+
+    // To-do ownable
```
barrutko, line 25: 🚨🚨🚨

barrutko (changes requested):
Tests are must have:
- unit tests with mocks for the purpose of  testing calculations
- integration tests forking mainnet

yuchenlintt:
> Now the hardest part would be to take that number to the power of 365. An example of how to take a number with `1e18` precision to the power of 2:

Thanks Mikolaj for the in-depth explanation! @jpalmer-trust if you want a reference, this is known as [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) (or double-and-add for additive groups). I would recommend extracting the exponentiation logic to a separate library contract so we don't clutter SBRO with math.

yuchenlintt:
Come to think of it, we already _have_ a fixed point math library:

https://github.com/trusttoken/smart-contracts/blob/dev/contracts/truefi2/libraries/TrueFiFixed64x64.sol

I think you could just copy the `pow(int128, uint256)` function from ABDKMath64x64 into our TrueFiFixed64x64 library:

https://github.com/abdk-consulting/abdk-libraries-solidity/blob/d8817cb600381319992d7caa038bf4faceb1097f/ABDKMath64x64.sol#L366

PR #875, opened 2021-08-17
Fork Chainlink USDC oracle for USDT and TUSD

yuchenlintt:


vanruch (commented):
Do we need this?

yuchenlintt:
> Do we need this?

I don't know. Hal mentioned yesterday that we need to deploy these for each TrueFiPool2, but I haven't dived into the code to see where the oracles are used. If they're not needed then I'm in favor of deleting. I'll take a look.

MiksuJak (approved):

yuchenlintt:
> Do we need this?

Yeah, we definitely need this (especially for the legacy TrueFiPool). I was debugging failing loan reclaims today and realized the tfTUSD failures aren't because of fee slippage (as we'd always thought). I simulated some txs and found out that they're failing on the `pool.oracle()` call _before_ we check that the swapped fee is more than the expectedDiff:

https://dashboard.tenderly.co/yuchenlintt/project/simulator/28a836fb-1a2d-4697-a535-56bf77fb73be/debugger?trace=0.0.2.7.8
https://github.com/trusttoken/smart-contracts/blob/8fe3633be4dfcdcfaa74924d7fccdc62e87b20de/contracts/truefi2/TrueLender2.sol#L445

It looks like we haven't set an oracle yet for TrueFiPool:

https://etherscan.io/address/0xa1e72267084192Db7387c8CC1328fadE470e4149#readProxyContract

This currently causes every legacy tfTUSD LoanToken2 created by TrueLender2 to be non-reclaimable. This issue doesn't show up for TrueLender tfTUSD LoanTokens because they don't pay swap fees, it doesn't show up for tfUSDC because USDC is the fee token, and it doesn't show up for tfUSDT because we set an oracle at some point.[1] The issue _only_ appears for the very few tfTUSD LoanToken2s created by TrueLender2 that have already settled... hence why it took so long for this issue to crop up.

[1] The tfUSDT oracle is not checked in to the repo(!), but it's the same as my `ChainlinkTruUsdtOracle` implementation in this PR.

https://etherscan.io/address/0x6002b1dcB26E7B1AA797A17551C6F487923299d7#readProxyContract
https://etherscan.io/address/0xec702942609fad2afc1489277d918611f0e17ceb#code

vanruch (approved):

PR #876, opened 2021-08-17
Add Chainlink oracles to deploy script

yuchenlintt:


deploy/truefi2.ts
```diff
@@ -96,7 +98,9 @@ deploy({}, (_, config) => {
   // New bare contracts
   const trueFiPool2 = contract(TrueFiPool2)
   const implementationReference = contract(ImplementationReference, [trueFiPool2])
+  const chainlinkTruTusdOracle = contract(ChainlinkTruTusdOracle)
   const chainlinkTruUsdcOracle = contract(ChainlinkTruUsdcOracle)
+  const chainlinkTruUsdtOracle = contract(ChainlinkTruUsdtOracle)
```
vanruch, line 103: You should name them differently, otherwise 2nd and 3rd deploys will be skipped

barrutko: Aren't these 3 different contracts?

vanruch: They are, missed the parallel PR. No need to change then

vanruch (approved):

deploy/truefi2.ts
yuchenlintt: You do raise a good readability point. With this naming scheme it would be easy for us to overlook a typo.

Doesn't have to be fixed now, but something to keep in mind for the future.

PR #884, opened 2021-08-19
Introduce BorrowingMutex contract

MiksuJak:


yuchenlintt (changes requested):
Thanks for the new version! I just have two naming proposals, otherwise this looks pretty good.

1. Rename `Registry` to `Mutex`

I don't think `Registry` is descriptive enough to describe what this contract does (specifically: mutual exclusion of lending protocols for a borrower address). I'm also worried that the semantics would be confusing. As a code reviewer/auditor, what would be your first impression of the following snippet?

```
function borrow(uint256 amount) external blah {
    if (registry.hasLock(msg.sender) == address(0)) {
        registry.lock(msg.sender);
    }
    require(registry.hasLock(msg.sender) == address(this), "error msg about not having lock");
    ...
}
```

For me, I would think the _registry_ (instead of TCA) owns a lock for the borrower, and it's locking the borrower into some state controlled by the registry.

`Mutex` avoids this confusion because it's a standard term in computer science with widely accepted semantics.[[1]](https://en.wikipedia.org/wiki/Reentrant_mutex)[[2]](https://en.cppreference.com/w/cpp/thread/mutex/lock) For a mutex, the `lock()` and `unlock()` functions are understood to grant the _caller_ a mutually exclusive right to modify the associated state.

2. Rename `hasLock()` to `locker()`

`hasLock()` sounds like it should return a boolean instead of an address.

This is what the above snippet would look like with these two naming changes:


```
function borrow(uint256 amount) external blah {
    if (mutex.locker(msg.sender) == address(0)) {
        mutex.lock(msg.sender);
    }
    require(mutex.locker(msg.sender) == address(this), "error msg about not having lock");
    ...
}
```

I think the intent of this code is much clearer. Other ideas I'm okay with:
- `acquire()`/`release()`
- an idempotent `lock()` that's safe to call multiple times (so caller doesn't need to check first)

Let me know what you think!

contracts/truefi2/BorrowingMutex.sol
```diff
@@ -0,0 +1,48 @@
...
+
+    function lock(address borrower, address _locker) external override {
+        require(canLock[msg.sender], "BorrowingMutex: Sender is not allowed to lock borrowers");
+        locker[borrower] = _locker;
+        emit BorrowerLocked(borrower, _locker);
+    }
+
+    function unlock(address borrower) external override {
+        address _locker = locker[borrower];
+        require(canLock[msg.sender] || _locker == msg.sender, "BorrowingMutex: Only locker or allowed addresses can unlock");
```
yuchenlintt, line 44:
```suggestion
        require(_locker == msg.sender, "BorrowingMutex: Only locker can unlock");
```

Since _any_ `canLock` address is now able to call `unlock()`, this behavior is more like a binary semaphore than a mutex (where only the original locker can call `unlock()`).

I prefer mutex over semaphore because it provides a stronger safety guarantee and allows us to reason about contract state changes _locally_ instead of _globally_. For instance, with a mutex there's no possibility for the following semaphore state violation to happen:
1. `trueLender` locks `borrower` to `loan1`.
2. `trueCreditAgency` (or `trueLender`, or some future contract/EOA that's been granted `canLock` permissions) unlocks `borrower` by exploit or accident.
3. `trueLender` sees that borrower is unlocked, and lets `borrower` take out another `loan2`.
4. A malicious `borrower` could exploit this to repeatedly withdraw past their `borrowLimit`, or take out FTL and LoC at the same time.
5. When we try to call `loan1.settle()`, our tx reverts because `loan1` is no longer the `locker` for `borrower`. Any funds returned to `loan1` are now frozen because the loan can't be settled.

It would be very easy to overlook this kind of outside interaction in an audit that's only scoped to FTL xor LoC. This is part of why designing distributed systems is really hard!

That said, I think the semantics of this contract have stabilized enough that we can check this PR in as-is to unblock other tasks. If you disagree with this suggestion, then we can have a follow-up discussion in a separate PR.

yuchenlintt (dismissed):
I think in commit 298a04d you introduced a code fragility vuln inside the `unlock()` function. But the overall contract looks okay, so if you don't agree with my suggested change then we can merge this PR now and take the defensive programming discussion to a follow-up PR.

contracts/truefi2/BorrowingMutex.sol
vanruch: I agree with @yuchenlintt point, but what about this corner case: when the mutex will be freshly introduced, the ongoing loans will not be marked in it. What's more, somebody might have several loans/LoC at that moment. This PR will also break in this case so we have yet to figure out how to do this right

vanruch (approved):

vanruch (approved):

vanruch (approved):

barrutko (approved):
LGTM

contracts/truefi2/BorrowingMutex.sol
yuchenlintt: During the upgrade transition, we can temporarily `allowLocker()` for the multisig or an EOA and manually call `lock(borrower, EOA)` and `unlock(borrower)` from the EOA.

I agree on the broader issue of figuring out how to have several FTLs across different pools. Seems like we don't have a good resolution to that yet. (It would be good to bring this up at today's biz sync.)

PR #885, opened 2021-08-20
Restrict borrower loan amount to be less than TRA borrowLimit() (FTL)

skalermo:


test/integration/TrueLender2.test.ts
```diff
@@ -115,7 +125,7 @@ describe('TrueLender2', () => {
     expect(await usdcFeePool.balanceOf(stkTru.address)).to.gt(utils.parseUnits('100', 6).mul(98).div(100))
   })
 
-  it('funds tether loan tokens', async () => {
+  xit('funds tether loan tokens', async () => {
```
vanruch, line 128: why did you skip this test?

test/integration/TrueLender2.test.ts
```diff
@@ -86,6 +93,9 @@ describe('TrueLender2', () => {
     await poolFactory.createPool(usdt.address)
     usdtLoanPool = TrueFiPool2__factory.connect(await poolFactory.pool(usdt.address), owner)
 
+    await mockCreditAgency.mock.borrowLimit.withArgs(tusdLoanPool.address, OWNER).returns(parseEth(100_000_000))
+    await mockCreditAgency.mock.borrowLimit.withArgs(usdtLoanPool.address, OWNER).returns(parseEth(100_000_000))
```
vanruch, line 97: I'm fine with using mock in tests, but we need to add at least one integration test with the real credit agency

vanruch: By integration I mean a unit test but with a real contract :)

test/truefi/TrueFiPool.test.ts
```diff
@@ -493,13 +497,35 @@ describe('TrueFiPool', () => {
     })
   })
 
+  const setupCreditAgency = async () => {
```
MiksuJak, line 500: This function takes a lot of space. Consider extracting it to utils maybe?

test/truefi2/Liquidator2.test.ts
```diff
@@ -65,6 +80,10 @@ describe('Liquidator2', () => {
     await tru.mint(otherWallet.address, parseEth(15e6))
     await tru.approve(stkTru.address, parseEth(1e7))
     await tru.connect(otherWallet).approve(stkTru.address, parseEth(1e7))
+
+    await creditAgency.allowPool(pool.address, true)
```
MiksuJak, line 84: I think we should consider adding allowing pools tusd and usdc to setupTruefi2. Maybe in a new PR.

test/utils/setUtilization.ts
```diff
@@ -0,0 +1,79 @@
...
+  TrueRatingAgencyV2,
+  TrueLender2,
+  LoanFactory2,
+} from 'contracts'
+import { BigNumber, Wallet } from 'ethers'
+import { MockProvider } from 'ethereum-waffle'
+import { DAY } from './constants'
+import { createApprovedLoan as _createApprovedLoan } from './createLoan'
+
+const includeFee = (amount: BigNumber) => amount.mul(10000).div(9975)
```
MiksuJak, line 14: I think we can get rid of it

skalermo: Yeah, you are right

test/integration/TrueLender2.test.ts
vanruch: Nevermind, didn't notice these are integration tests

test/utils/setUtilization.ts
```diff
@@ -0,0 +1,77 @@
...
+  )
+
+  const poolValue = await pool.poolValue()
+  const utilizationAmount = poolValue.mul(utilization).div(100)
+  if (utilization <= 15) {
+    const loan = await createApprovedLoan(borrower1, utilizationAmount)
+    await lender.connect(borrower1).fund(loan.address)
+    return
+  }
+  if (utilization <= 30) {
```
MiksuJak, line 48: if else?

skalermo: There is no need (unless it might improve readability which I doubt). Notice early return statement in previous *if*.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -300,7 +300,7 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
      * @param borrower Borrower to get borrow limit for
      * @return borrow limit for `borrower` in `pool`
      */
-    function borrowLimit(ITrueFiPool2 pool, address borrower) public view returns (uint256) {
+    function borrowLimit(ITrueFiPool2 pool, address borrower) public override view returns (uint256) {
```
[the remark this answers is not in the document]
skalermo: I agree, LoC's `borrowLimit()` shouldn't be also used for FTL. We have common `borrowLimit()` in TRA, and honestly, since both LoC and FTL would use it, I think TRA could use a better naming.

> Perhaps both of them could pass in their own `totalBorrowed(borrower, decimals)` values as arguments to `borrowLimit()`?

I believe this is the way to go.

PR #886, opened 2021-08-20
Move TVL calculation from TrueCreditAgency to TrueRateAdjuster

yuchenlintt:
If we move total TVL to TRA, then TrueLender2 should have access to all the info it requires to calculate `borrowLimit()` _without_ adding TrueCreditAgency as a dependency. Hopefully that resolves this discussion favorably: https://github.com/trusttoken/smart-contracts/pull/885#discussion_r693176256

### DIFF
--- a/contracts/truefi2/TrueRateAdjuster.sol
+++ b/contracts/truefi2/TrueRateAdjuster.sol
@@ -133,7 +133,7 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
     /**
      * @dev Add `pool` to TVL calculation
      */
-    function addPooltoTVL(ITrueFiPool2 pool) external onlyOwner {
+    function addPoolToTVL(ITrueFiPool2 pool) external onlyOwner {
         for (uint256 i = 0; i < tvlPools.length; i++) {
             require(tvlPools[i] != pool, "TrueRateAdjuster: Pool has already been added to TVL");
         }
--- a/contracts/truefi2/TrueRateAdjuster.sol
+++ b/contracts/truefi2/TrueRateAdjuster.sol
@@ -134,6 +134,9 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
      * @dev Add `pool` to TVL calculation
      */
     function addPooltoTVL(ITrueFiPool2 pool) external onlyOwner {
+        for (uint256 i = 0; i < tvlPools.length; i++) {
+            require(tvlPools[i] != pool, "TrueRateAdjuster: Pool has already been added to TVL");
+        }
         tvlPools.push(pool);
         emit AddPoolToTVL(pool);
     }
--- a/contracts/truefi2/TrueRateAdjuster.sol
+++ b/contracts/truefi2/TrueRateAdjuster.sol
@@ -84,7 +84,13 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
     /// @dev store borrow limit configuration
     BorrowLimitConfig public borrowLimitConfig;
 
+    /// @dev array of pools used to calculate TrueFi TVL
+    ITrueFiPool2[] public tvlPools;
+
     // ======= STORAGE DECLARATION END ============
+
+    /// @dev emit `pool` and `isIncluded` when TVL pool inclusion status changes
+    event TVLPool(ITrueFiPool2 pool, bool isIncluded);
 
     /// @dev Emit `newRate` when risk premium changed
     event RiskPremiumChanged(uint256 newRate);
@@ -119,6 +125,24 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
         creditScoreRateConfig = CreditScoreRateConfig(1000, 1);
         fixedTermLoanAdjustmentCoefficient = 25;
         borrowLimitConfig = BorrowLimitConfig(40, 7500, 1500, 1500);
+    }
+
+    /**
+     * @dev Set whether `pool` is included in TVL calculation
+     */
+    function setTVLPool(ITrueFiPool2 pool, bool isIncluded) external onlyOwner {
+        if (isIncluded) {
+            tvlPools.push(pool);
+        } else {
+            for (uint256 i = 0; i < tvlPools.length; i++) {
+                if (tvlPools[i] == pool) {
+                    tvlPools[i] = tvlPools[tvlPools.length - 1];
+                    tvlPools.pop();
+                    break;
+                }
+            }
+        }
+        emit TVLPool(pool, isIncluded);
     }
 
     /// @dev Set risk premium to `newRate`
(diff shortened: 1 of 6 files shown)


contracts/truefi2/TrueRateAdjuster.sol
```diff
@@ -121,6 +127,24 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
...
+            for (uint256 i = 0; i < tvlPools.length; i++) {
+                if (tvlPools[i] == pool) {
+                    tvlPools[i] = tvlPools[tvlPools.length - 1];
+                    tvlPools.pop();
+                    break;
+                }
+            }
+        }
+        emit TVLPool(pool, isIncluded);
+    }
```
vanruch, line 146: I think `isIncluded` name is not clear enough. Maybe we want to split this into 2 functions?
Also we should probably add check preventing adding same pool twice

yuchenlintt: These are good points, thanks! I'll make these changes.

yuchenlintt:
> We are overloading TRA with stuff that shouldn't be there. I think that also goes for `borrowLimit()`. I agree that, since total TVL will be used not only in LoC, it should be moved out of TCA. Though I am not sure if TRA is the right place.

I agree in principle. If we keep extracting code from TCA and dumping it into TRA, we'll eventually end up with the same large mess in TRA that we were trying to remove from TCA.

Ideas for where this could go instead: PoolFactory, another new contract.

[We decided in sync today to keep this in TRA for now but rename TRA to be more meaningful.]

yuchenlintt:
We just had another sync where we decided to merge all pool whitelists to a single canonical location. I'm not sure where this location should be (open for discussion).

Regardless of the discussion, we know this needs to be taken out of TCA to prevent a TL -> TCA dependency. So for now let's submit this to TRA and we can decide whether to move it again later. (No changes from my previous comment.)

contracts/truefi2/TrueRateAdjuster.sol
```diff
@@ -121,6 +130,32 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
         borrowLimitConfig = BorrowLimitConfig(40, 7500, 1500, 1500);
     }
 
+    /**
+     * @dev Add `pool` to TVL calculation
+     */
+    function addPooltoTVL(ITrueFiPool2 pool) external onlyOwner {
+        for (uint256 i = 0; i < tvlPools.length; i++) {
+            require(tvlPools[i] != pool, "TrueRateAdjuster: Pool has already been added to TVL");
```
skalermo, line 138: Can we test this?

contracts/truefi2/TrueRateAdjuster.sol
```diff
@@ -263,28 +298,43 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
...
+     * @dev Calculate total TVL in USD
+     * @param decimals Precision to return
+     * @return TVL for all pools
+     */
+    function totalTVL(uint8 decimals) public view returns (uint256) {
+        uint256 tvl = 0;
+        uint256 resultPrecision = uint256(10)**decimals;
+
+        // loop through pools and sum tvl accounting for precision
+        for (uint8 i = 0; i < tvlPools.length; i++) {
```
skalermo, line 311: I think it should be `uint256 i`.

contracts/truefi2/TrueRateAdjuster.sol
```diff
@@ -84,8 +84,17 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
...
+    ITrueFiPool2[] public tvlPools;
+
     // ======= STORAGE DECLARATION END ============
 
+    /// @dev emit `pool` when adding to TVL
+    event AddPoolToTVL(ITrueFiPool2 pool);
+
+    /// @dev emit `pool` when removing from TVL
+    event RemovePoolFromTVL(ITrueFiPool2 pool);
+
```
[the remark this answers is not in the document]
yuchenlintt: Thanks! I'll update this to your suggestion.

contracts/truefi2/TrueRateAdjuster.sol
```diff
@@ -121,6 +130,32 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
...
+    function removePoolFromTVL(ITrueFiPool2 pool) external onlyOwner {
+        for (uint256 i = 0; i < tvlPools.length; i++) {
+            if (tvlPools[i] == pool) {
+                tvlPools[i] = tvlPools[tvlPools.length - 1];
+                tvlPools.pop();
+                emit RemovePoolFromTVL(pool);
+                return;
+            }
+        }
+        revert("TrueRateAdjuster: Pool already removed from TVL");
```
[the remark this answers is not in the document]
yuchenlintt: Oops, I should've remembered to add unit tests for these functions. I'll do so now.

contracts/truefi2/TrueRateAdjuster.sol
yuchenlintt: This index was already uint8 when I cut/paste it, but I'm happy to change this to uint256.

skalermo:
Consider setting up pool-add/remove tests using `beforeEach`. Otherwise looks good.

test/truefi2/lines-of-credit/TrueRateAdjuster.test.ts
```diff
@@ -72,6 +72,89 @@ describe('TrueRateAdjuster', () => {
     })
   })
 
+  describe('addPoolToTVL', () => {
```
skalermo, line 75: I noticed this *describe* and the next one are under `initialize` *describe*. Let's move them out of there.

skalermo: Nevermind, everything's good.

yuchenlintt: Are you sure? It doesn't look like nested to me, and when I run the unit tests they don't show up under the `initializer` subtree.

skalermo (dismissed):

MiksuJak (dismissed):
Is `totalTVL` tested anywhere after these changes? Also Let's maybe rename it. FIrst of all it is literally `total total value locked` and we kind of missed it before 🙃. Then, TVL does not really reflect what the function calculates, as it is value of the chosen pools instead of total value. Maybe `riskValueLocked` as we discussed yesterday?

yuchenlintt:
> Is `totalTVL` tested anywhere after these changes?

I dropped offline for a couple hours to sleep. Thanks Roman for adding tests, although I think this could have been done in a separate PR to `dev`. _This_ PR was just supposed to be a simple move that would let us handle #885 in smaller pieces. Unit tests should have existed before the move, but now that we notice they're missing during the move, we should have filed this as a follow-up task. That way we could more quickly unblock #885.

> Also Let's maybe rename it. FIrst of all it is literally `total total value locked` and we kind of missed it before 🙃. Then, TVL does not really reflect what the function calculates, as it is value of the chosen pools instead of total value. Maybe `riskValueLocked` as we discussed yesterday?

`tvl()` sounds fine for me now (certainly better than total total value locked :)). As I recall `riskValueLocked` was only mentioned in passing yesterday, not a firm decision.

skalermo:
Sorry for 'invading' your PR @yuchenlintt. Actually tests for `totalTVL` were refactored out during the move of `totalTVL`. I was hoping for a quick merge, so I decided to bring them back in.

yuchenlintt:
No worries! You ended up being faster than my sleep cycle, so we didn't lose any additional time (apart from continuing to block your PR).

skalermo (approved):

MiksuJak (approved):

PR #888, opened 2021-08-21
Proof of Concept: Add totalBorrowed() view to TrueLender2

yuchenlintt:
Getting an array of pools from trueRateAdjuster feels like a hack, but this certainly looks like a valid way to sum all loans from a borrower. @skalermo if this approach makes sense, feel free to take over this PR and/or incorporate it into PR #885.


yuchenlintt:
Closing this PR. The function signature looks good but we don't yet know what our implementation should be. Blocked on a business decision regarding per-borrower limits on the number of FTL loans.

PR #889, opened 2021-08-23
Integrate BorrowingMutex with Fixed-term Loans

vanruch:


contracts/truefi2/LoanToken2.sol
```diff
@@ -353,6 +361,9 @@ contract LoanToken2 is ILoanToken2, ERC20 {
         require(!isRepaid(), "LoanToken2: cannot default a repaid loan");
         require(start.add(term).add(LAST_MINUTE_PAYBACK_DURATION) <= block.timestamp, "LoanToken2: Loan cannot be defaulted yet");
         status = Status.Defaulted;
+
+        borrowingMutex.unlock(borrower);
```
yuchenlintt, line 365: If our borrowers enter default, I don't think we want to unlock them. This unlock would let them take out further loans or possibly open an LOC.

I would be in favor of keeping this address permanently locked and using the BorrowingMutex as a gravestone that points to the defaulted loan. If a defaulted borrower wanted to reestablish a business relationship with TrueFi, they would need to start over with a new address, credit score, KYC, etc.

yuchenlintt (changes requested):
I don't think we should unlock a borrower if they default, otherwise this looks great!

yuchenlintt:
In case I'm not awake when you get back online: this PR LGTM after removing `unlock()` from `enterDefault()`. Assuming that's the only change (modulo downstream unit test updates), feel free to dismiss my review and merge this PR.

contracts/truefi2/LoanToken2.sol
vanruch: I was thinking about this and wasn't sure if we want the mutex to be responsible for blocking borrowers after loan default. Be kinda have other instruments like credit oracle for that. And after the loan is defaulted, well, there're no more loans taken by the borrower. But on the other hand, it will be quite convenient to have borrowers blocked automatically like this

yuchenlintt: Although we do have other ways to disqualify borrowers, I'm concerned about us making mistakes under a high task load. In an emergency situation, will we remember to perform the correct call sequence (`creditOracle.setScore(0)`, _and then_ `loan.enterDefault()`)?

Automatic locking is a nice failsafe. I wouldn't see this as 'responsible' for blocking borrowers. It's more like a layer of protection in a defense-in-depth strategy.

Since our contracts are proxy-owned, we could take our time to plan and deploy an upgrade if we _really_ wanted to unlock a specific borrower address after default.

yuchenlintt (approved):

MiksuJak (approved):

PR #898, opened 2021-08-24
Integrate TrueCreditAgency with BorrowingMutex

dymitrlubczyk:


yuchenlintt (changes requested):
Thanks for the PR! This mostly looks good.

- One change to be more defensive against invalid or unforeseen state
- Add a unit test for this scenario
- Some minor formatting notes

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -345,6 +350,11 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
         (uint8 oldScore, uint8 newScore) = _updateCreditScore(pool, msg.sender);
         require(newScore >= minCreditScore, "TrueCreditAgency: Borrower has credit score below minimum");
         require(amount <= borrowLimit(pool, msg.sender), "TrueCreditAgency: Borrow amount cannot exceed borrow limit");
+
+        if(totalBorrowed(msg.sender, 18) == 0){
+            borrowingMutex.lock(msg.sender, address(this));
+        }
+
```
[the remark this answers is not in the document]
MiksuJak: @yuchenlintt I belive there is a test for this:
`
it('fails if borrower mutex is already locked', async () => {
      await borrowingMutex.allowLocker(owner.address, true)
      await borrowingMutex.lock(borrower.address, owner.address)

      await expect(creditAgency.connect(borrower).borrow(tusdPool.address, 1000))
        .to.be.revertedWith('BorrowingMutex: Borrower is already locked')
    })
`

test/truefi2/lines-of-credit/TrueCreditAgency.test.ts
```diff
@@ -692,6 +712,12 @@ describe('TrueCreditAgency', () => {
       expect(await creditAgency.nextInterestRepayTime(tusdPool.address, borrower.address)).to.eq(0)
     })
 
+    it('sets nextInterestRepayTime to 0', async () => {
+      expect(await borrowingMutex.locker(borrower.address)).to.eq(creditAgency.address)
+      await creditAgency.connect(borrower).repayInFull(tusdPool.address)
+      expect(await borrowingMutex.locker(borrower.address)).to.eq(AddressZero)
+    })
```
skalermo, line 719: This test's name doesn't match its contents.

test/truefi2/lines-of-credit/TrueCreditAgency.test.ts
```diff
@@ -430,6 +445,11 @@ describe('TrueCreditAgency', () => {
       expect(await creditAgency.nextInterestRepayTime(tusdPool.address, borrower.address)).to.eq(timestamp.add(MONTH))
     })
 
+    it('borrows and check if mutex is locked', async () => {
```
skalermo, line 448: This name doesn't sound right to me. Maybe `borrowing locks mutex`, but I'm open to other suggestions as well.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -107,6 +108,9 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     /// @dev credit oracle
     ITrueFiCreditOracle public creditOracle;
 
+    // mutex ensuring there's only one running loan or credit line for borrower
+    IBorrowingMutex public borrowingMutex;
```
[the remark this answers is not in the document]
dymitrlubczyk: I will keep it on mind

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -345,6 +350,11 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
         (uint8 oldScore, uint8 newScore) = _updateCreditScore(pool, msg.sender);
         require(newScore >= minCreditScore, "TrueCreditAgency: Borrower has credit score below minimum");
         require(amount <= borrowLimit(pool, msg.sender), "TrueCreditAgency: Borrow amount cannot exceed borrow limit");
+
+        if(totalBorrowed(msg.sender, 18) == 0) {
+            borrowingMutex.lock(msg.sender, address(this));
```
barrutko, line 355: This will implicitly fail if the mutex is already locked, which is obviously right, but not sure if we shouldn't check it explicitly.

barrutko: With error a message like `TrueCreditAgency: Borrower cannot open two parallel debt positions`, or something like that.

dymitrlubczyk: In which way?

vanruch: You can add a similar line https://github.com/trusttoken/smart-contracts/blob/8374c789f31e37b4880979c38b5e1d7cd9853848/contracts/truefi2/TrueLender2.sol#L353

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -10,12 +10,12 @@ import {ITrueFiPool2} from "./interface/ITrueFiPool2.sol";
 import {ITrueCreditAgency} from "./interface/ITrueCreditAgency.sol";
 import {ITrueFiCreditOracle} from "./interface/ITrueFiCreditOracle.sol";
 import {ITimeAveragedBaseRateOracle} from "./interface/ITimeAveragedBaseRateOracle.sol";
+import {IBorrowingMutex} from "./interface/IBorrowingMutex.sol";
 
 interface ITrueFiPool2WithDecimals is ITrueFiPool2 {
     function decimals() external view returns (uint8);
 }
 
-// prettier-ignore
```
yuchenlintt, line 18: I would also like to remove `prettier-ignore`, but it looks like prettier deletes all our comments for this contract.

Personally, I'm in favor of deleting these comments since I prefer code to be self-documenting. (Documentation should be used sparingly to answer _why_ code is designed a particular way instead of duplicating _what_ it does, in a way that might go stale.)

However, keeping these comments is a hard requirement from above, so we can't just remove it. Until the comment deletion issue is fixed, we have to keep `prettier-ignore` here and keep all the comments.

yuchenlintt (changes requested):

skalermo (approved):

PR #903, opened 2021-08-26
Deploy TrueFiCreditOracle impl to mainnet

yuchenlintt:
This upgrade is primarily meant to allow us to set a manager address to update credit scores.

The contract code is currently UNAUDITED and contains logic that we may want to change before release.

I'm not sure yet that we want to check this in. But if we do, then we should have 3 diff checks on this PR.

### DIFF
--- a/deploy/truefi2.ts
+++ b/deploy/truefi2.ts
@@ -53,117 +53,117 @@ const deployParams = {
 }
 
 deploy({}, (_, config) => {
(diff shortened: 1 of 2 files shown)


yuchenlintt:
I'm closing this PR for now. We've decided to use the current oracle without upgrading, and just transfer ownership to a new address.

PR #905, opened 2021-08-30
Add dummy exit() and liquidExitPenalty() functions back to TrueFiPool for backward compatibility

yuchenlintt:


vanruch (approved):

barrutko (approved):
LGTM. Why not `0` instead of `10000` though?

yuchenlintt:
As I recall, 10000 represents no penalty and 0 represents 100% penalty. Thus existing contracts reading 10000 would correctly assume there is no longer any penalty.

PR #909, opened 2021-08-31
Reduce loan creation gas cost

skalermo:
As of now loan creation gas cost is very high. This PR addresses the issue by integrating [EIP-1167](https://eips.ethereum.org/EIPS/eip-1167) into current loan creation mechanism. 

Profiling results for loan creation on my local machine are as shown below:

No changes: 2,917,839 gas
Minimal proxy contract: 515,235 gas
OwnedUpgradabilityProxy: 916,955 gas

[EDIT] Modified LoanFactory2 was deployed to Kovan testnet and a sample loan token was created.
LoanFactory2: https://kovan.etherscan.io/address/0xA6C592a08FBDe619Ae487ddf1ae048f5E59197a1#code
LoanToken2: https://kovan.etherscan.io/address/0x3418fda746642ffd195dea5f292e6c381d021601#code


contracts/truefi2/LoanFactory2.sol
```diff
@@ -111,12 +117,24 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
...
         require(_term > 0, "LoanFactory: Loans cannot have instantaneous term of repay");
         require(poolFactory.isPool(address(_pool)), "LoanFactory: Pool was not created by PoolFactory");
+        require(loanTokenImplementation != address(0), "Loan token implementation should be set");
 
         uint256 apy = rate(_pool, msg.sender, _amount, _term);
 
         require(apy <= _maxApy, "LoanFactory: Calculated apy is higher than max apy");
 
-        address newToken = address(new LoanToken2(_pool, borrowingMutex, msg.sender, lender, admin, liquidator, _amount, _term, apy));
+        address newToken = Clones.clone(loanTokenImplementation);
```
vanruch, line 126: You can just call methods on new LoanToken2(newToken) without doing low-level calls

contracts/truefi2/LoanToken2.sol
```diff
@@ -150,7 +150,7 @@ contract LoanToken2 is ILoanToken2, ERC20 {
         uint256 _amount,
         uint256 _term,
         uint256 _apy
-    ) public {
+    ) external initializer {
```
barrutko, line 153: Shouldn't the contract be `Initializable` in order for that to work?

barrutko: Ok, my bad, all good 😅

barrutko (approved):

contracts/truefi2/LoanToken2.sol
vanruch: ERC20 is Initializable (it wouldn't even compile if it wasn't)

vanruch (approved):

contracts/proxy/Clones.sol
```diff
@@ -0,0 +1,86 @@
+// SPDX-License-Identifier: MIT
+// COPIED FROM https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.3.0/contracts/proxy/Clones.sol
```
yuchenlintt, line 2: I diff checked against OZ's Clones and confirmed this is (nearly) identical. Any reason we're not just importing Clones from OZ? Is it just the `pragma solidity` compiler version?

contracts/truefi2/LoanFactory2.sol
```diff
@@ -40,6 +41,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
     ITrueRateAdjuster public rateAdjuster;
     ITrueFiCreditOracle public creditOracle;
     IBorrowingMutex public borrowingMutex;
+    address public loanTokenImplementation;
```
yuchenlintt, line 44: Can we allow the compiler to typecheck this for us? We would declare it in the contract as `ILoanToken2 public loanTokenImplementation` but then need to call `Clones.clone(address(loanTokenImplementation))` when passing it to a proxy later.

This would give us more of a safeguard in our tests and deploy scripts to make sure we're not passing in (for instance) a deprecated ILoanToken or IDeficiencyToken by accident. And if we do a LoanToken2 refactor or upgrade, we won't forget to check whether that breaks the build in this contract.

Of course, declaring the correct type here doesn't prevent us from passing in the wrong address on mainnet :)

yuchenlintt (changes requested):

contracts/truefi2/LoanFactory2.sol
yuchenlintt: In case I'm not awake to re-approve, this PR LGTM after this fix.

vanruch (approved):

barrutko (approved):

PR #911, opened 2021-09-01
Update credit scores

yuchenlintt:
I don't know that score updates need to be checked in, but it was cheap for me to create a PR.

### DIFF
--- a/deploy/credit_scores.ts
+++ b/deploy/credit_scores.ts
@@ -6,20 +6,27 @@ import {
 import { utils, BigNumber } from 'ethers'
 
 const SCORES = [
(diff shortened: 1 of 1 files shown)


vanruch (approved):
I'm approving the PR but honestly I have no idea if the scores and addresses are correct. 

Ideally we need to have a single source of scores which will be controlled by the people responsible for credit scores and fetch data from there.

barrutko (approved):

yuchenlintt:
I agree. These values came from the credit team, and I don't have a way to verify their integrity either.

We're designing a better process for credit score updates, but our stopgap for now (pre-LOC launch) is to update manually.

PR #913, opened 2021-09-01
 Replace TCA isPoolAllowed() and pools[] with PF.isSupportedPool()

yuchenlintt:
Please don't merge this PR until #914 has merged to `dev` (to avoid slowing down review of #914).


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -110,6 +105,8 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     // mutex ensuring there's only one running loan or credit line for borrower
     IBorrowingMutex public borrowingMutex;
 
+    IPoolFactory public poolFactory;
+
```
skalermo, line 109: Do we need to add setters for these?

skalermo (dismissed):

contracts/truefi2/TrueCreditAgency.sol
vanruch: We absolutely should. We'll most probably deploy the TCA without this change first

vanruch: But it could be added in a separate PR

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -89,12 +90,6 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     /// @dev next payment due time per borrower for each pool
     mapping(ITrueFiPool2 => mapping(address => uint256)) public nextInterestRepayTime;
 
-    /// @dev whitelist for allowing pools to have lines of credit
-    mapping(ITrueFiPool2 => bool) public isPoolAllowed;
-
-    /// @dev array of pools with lines of credit
-    ITrueFiPool2[] public pools;
```
vanruch, line 96: Didn't we plan to deploy older audited version first? This will break the storage

contracts/truefi2/TrueCreditAgency.sol
vanruch: Unless we will deploy the different version before this one, we don't need to

vanruch:
Other than my concern with storage change, looks good

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: Thanks for catching this. I believe the team would like to launch the audited version first and then follow up later with this version.

I'll revert the deletions here and mark these as DEPRECATED.

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: Unfortunately we will have to deploy the audited version before adding this. I'll add a setter and unit tests for this.

vanruch (approved):

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -89,11 +90,9 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     /// @dev next payment due time per borrower for each pool
     mapping(ITrueFiPool2 => mapping(address => uint256)) public nextInterestRepayTime;
 
-    /// @dev whitelist for allowing pools to have lines of credit
-    mapping(ITrueFiPool2 => bool) public isPoolAllowed;
+    mapping(address => bool) private isPoolAllowed_DEPRECATED;
```
MiksuJak, line 93: Previously we used `DEPRECATED__` as a prefix, so let's stick to the naming that is already in use

contracts/truefi2/TrueCreditAgency.sol
MiksuJak: If we plan to deploy audited version first and then upgrade, I believe this would break the storage

vanruch: Why would it?

MiksuJak (approved):

vanruch (approved):

contracts/truefi2/TrueCreditAgency.sol
yuchenlintt: Thanks for fixing and checking this in!

PR #914, opened 2021-09-02
Add isSupportedPool() view to PoolFactory

yuchenlintt:


test/truefi2/PoolFactory.test.ts
```diff
@@ -327,12 +327,12 @@ describe('PoolFactory', () => {
       const pool2 = await factory.pool(token2.address)
       const pool3 = await factory.pool(token3.address)
       const pool4 = await factory.pool(token4.address)
+
+      await expect(factory.isSupportedPool(pool3)).not
       await factory.supportPool(pool3)
+      await expect(factory.isSupportedPool(pool3))
```
vanruch, line 333: Are you sure this syntax works? Never seen it being used

vanruch: I don't think it works, after removing .not the test is still passing

vanruch: The correct way would be like this
expect(await factory.isSupportedPool(pool3)).to.be.false

contracts/truefi2/PoolFactory.sol
```diff
@@ -234,6 +234,15 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
         emit BorrowerWhitelistStatusChanged(borrower, status);
     }
 
+    function isSupportedPool(ITrueFiPool2 _pool) external override view returns (bool) {
+        for (uint256 i = 0; i < supportedPools.length; i++) {
+            if (supportedPools[i] == _pool) {
```
skalermo, line 239: Let's store `supportedPools.length` in a local variable before the loop, so it doesn't query storage variable over each iteration.

vanruch: I don't think it affects gas cost in any way

skalermo: It does, but not as much as I thought. For 200 pools in `supportedPools` I'm getting around 20_000 gas saving (overall approx. 100 gas usage reduction per pool).

yuchenlintt: I think checking array.length inside the for loop is fine. If there's not much gas savings, then this would be a premature optimization at the cost of a minor drop in readability. I could go either way, but I believe the rest of our repo checks array.length in the loop so for consistency we should do so here as well.

Thanks for noticing this and benchmarking it though!

test/truefi2/PoolFactory.test.ts
```diff
@@ -328,9 +328,9 @@ describe('PoolFactory', () => {
       const pool3 = await factory.pool(token3.address)
       const pool4 = await factory.pool(token4.address)
 
-      await expect(factory.isSupportedPool(pool3)).not
+      expect(await factory.isSupportedPool(pool3)).to.be.false
       await factory.supportPool(pool3)
-      await expect(factory.isSupportedPool(pool3))
+      expect(await factory.isSupportedPool(pool3))
```
skalermo, line 333:
```suggestion
      expect(await factory.isSupportedPool(pool3)).to.be.true
```

test/truefi2/PoolFactory.test.ts
```diff
@@ -375,9 +375,9 @@ describe('PoolFactory', () => {
       const pool2 = await factory.pool(token2.address)
       const pool4 = await factory.pool(token4.address)
 
-      await expect(factory.isSupportedPool(pool2))
+      expect(await factory.isSupportedPool(pool2))
```
skalermo, line 378:
```suggestion
      expect(await factory.isSupportedPool(pool2)).to.be.true
```

skalermo (approved):

vanruch (approved):

PR #920, opened 2021-09-02
Fix inefficient looping in TrueCreditAgency

vanruch:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -257,7 +258,8 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
         uint256 resultPrecision = uint256(10)**decimals;
 
         // loop through pools and sum tvl accounting for precision
-        for (uint8 i = 0; i < pools.length; i++) {
+        uint256 poolsCount = pools.length;
+        for (uint8 i = 0; i < poolsCount; i++) {
             tvl = tvl.add(
```
skalermo, line 263: I would suggest using consistent types. It would save us a bit of gas because of casting from `uint8` to `uint256` in every iteration.

PR #921, opened 2021-09-02
TCA-01 | Remove redundant variable initialization from TVL and totalBorrowed

yuchenlintt:
This isn't my personal style preference, but let's go with this to satisfy the audit and be consistent with the repo style.

### DIFF
--- a/contracts/truefi2/TrueCreditAgency.sol
+++ b/contracts/truefi2/TrueCreditAgency.sol
@@ -252,8 +252,7 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
      * @param decimals Precision to return
      * @return TVL for all pools with lines of credit
      */
-    function totalTVL(uint8 decimals) public view returns (uint256) {
-        uint256 tvl = 0;
+    function totalTVL(uint8 decimals) public view returns (uint256 tvl) {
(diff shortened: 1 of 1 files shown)


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -250,10 +250,9 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     /**
      * @dev Calculate total TVL in USD
      * @param decimals Precision to return
-     * @return TVL for all pools with lines of credit
+     * @return tvl TVL for all pools with lines of credit
```
barrutko, line 253: ❓❓❓
```suggestion
     * @return total TVL for all pools with lines of credit
```

barrutko (dismissed):

vanruch (changes requested):
This wouldn't be consistent with repo style, we have implicit return in exactly 1 place right now

yuchenlintt:
Sorry about that. As I recall, I had articulated a 'no implicit returns' stance in my reviews. Rather hypocritical of me, huh?

I'll fix this here and in #922.

barrutko (approved):

skalermo (approved):

PR #930, opened 2021-09-02
TFC-01 | DEV | Require TFCO manager is set to nonzero

yuchenlintt:


contracts/truefi2/TrueFiCreditOracle.sol
```diff
@@ -131,6 +131,7 @@ contract TrueFiCreditOracle is ITrueFiCreditOracle, UpgradeableClaimable {
      * @dev Set new manager for updating scores
      */
     function setManager(address newManager) public onlyOwner {
+        require(address(newManager) != address(0), "TrueFiCreditOracle: Cannot set new manager to address(0)");
```
barrutko, line 134: Why do we need to cast `newManager` to `address`, while it's already an `address`?

vanruch (dismissed):

test/truefi2/lines-of-credit/TrueFiCreditOracle.test.ts
```diff
@@ -35,6 +36,11 @@ describe('TrueFiCreditOracle', () => {
         .to.be.revertedWith('Ownable: caller is not the owner')
     })
 
+    it('cannot be set to address(0)', async () => {
```
barrutko, line 39: IMO would be cleaner, but not sure.
```suggestion
    it('cannot be set to zero address', async () => {
```

contracts/truefi2/TrueFiCreditOracle.sol
yuchenlintt: Thanks for catching that!

skalermo (approved):

vanruch (approved):

PR #934, opened 2021-09-03
Migrate more TL2 functions to use supported pools

yuchenlintt:

PR #937, opened 2021-09-03
Move TVL calculation from TRA to PoolFactory supportedPools()

dymitrlubczyk:
Tests of `addPoolToTVL`, `removePoolFromTVL` and `tvl` should be moved to PoolFactory tests. I decided to do it in another PR as this one changes 14 files already.

### DIFF
--- a/contracts/truefi2/PoolFactory.sol
+++ b/contracts/truefi2/PoolFactory.sol
@@ -254,7 +254,6 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
      * @dev Add `_pool` to official support
      */
     function supportPool(ITrueFiPool2 _pool) external onlyOwner {
-        //TODO: replace addPoolToTVL with it
         require(isPool[address(_pool)], "PoolFactory: Pool not created by factory");
 
         for (uint256 i = 0; i < supportedPools.length; i++) {
@@ -268,8 +267,6 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
      * @dev Remove `_pool` from official support
      */
     function unsupportPool(ITrueFiPool2 _pool) external onlyOwner {
-        //TODO: replace removePoolFromTVL with it
-
         for (uint256 i = 0; i < supportedPools.length; i++) {
             if (supportedPools[i] == _pool) {
                 supportedPools[i] = supportedPools[supportedPools.length - 1];
--- a/contracts/truefi2/PoolFactory.sol
+++ b/contracts/truefi2/PoolFactory.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.6.10;
 
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+
 import {UpgradeableClaimable} from "../common/UpgradeableClaimable.sol";
 import {OwnedProxyWithReference} from "../proxy/OwnedProxyWithReference.sol";
 import {ERC20, IERC20} from "../common/UpgradeableERC20.sol";
@@ -19,6 +21,7 @@ import {ISAFU} from "./interface/ISAFU.sol";
  * Initially created pools hold the same implementation, which can be changed later on individually
  */
 contract PoolFactory is IPoolFactory, UpgradeableClaimable {
+    using SafeMath for uint256;
     // ================ WARNING ==================
     // ===== THIS CONTRACT IS INITIALIZABLE ======
     // === STORAGE VARIABLES ARE DECLARED BELOW ==
@@ -46,7 +49,7 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
     mapping(address => bool) public isBorrowerWhitelisted;
 
     /// @dev array of pools officially supported by TrueFi
-    ITrueFiPool2[] public supportedPools;
+    ITrueFiPool2[] public supportedPools; //TODO: replace getTVLPools with this
 
     // ======= STORAGE DECLARATION END ===========
 
@@ -251,6 +254,7 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
      * @dev Add `_pool` to official support
      */
     function supportPool(ITrueFiPool2 _pool) external onlyOwner {
+        //TODO: replace addPoolToTVL with it
         require(isPool[address(_pool)], "PoolFactory: Pool not created by factory");
 
         for (uint256 i = 0; i < supportedPools.length; i++) {
@@ -264,6 +268,8 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
      * @dev Remove `_pool` from official support
      */
     function unsupportPool(ITrueFiPool2 _pool) external onlyOwner {
+        //TODO: replace removePoolFromTVL with it
+
         for (uint256 i = 0; i < supportedPools.length; i++) {
             if (supportedPools[i] == _pool) {
                 supportedPools[i] = supportedPools[supportedPools.length - 1];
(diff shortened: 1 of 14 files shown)


contracts/truefi2/PoolFactory.sol
```diff
@@ -295,4 +298,16 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
         safu = _safu;
         emit SafuChanged(_safu);
     }
+
+    /**
+     * @dev Calculate total TVL in USD
+     * @return _tvl TVL for all supported pools
+     */
+    function tvl() public override view returns (uint256) {
+        uint256 _tvl = 0;
```
yuchenlintt, line 307: nit: Our auditors prefer this style instead.
```suggestion
        uint256 _tvl;
```

contracts/truefi2/TrueLender2.sol
```diff
@@ -382,8 +382,7 @@ contract TrueLender2 is ITrueLender2, UpgradeableClaimable {
         uint256 borrowSum;
         uint256 resultPrecision = uint256(10)**decimals;
 
-        // loop through loans and sum amount borrowed accounting for precision
-        ITrueFiPool2[] memory tvlPools = rateAdjuster.getTVLPools();
+        ITrueFiPool2[] memory tvlPools = factory.getSupportedPools();
```
yuchenlintt, line 385: Apologies for my duplicate work in PR #934. I hadn't realized you were also planning to do these changes in this task.

Some lessons for both of us for the future:
- If there's a good chance someone else is doing the same work, I should message the task owner to double-check we're not duplicating, before I start.
- If a task starts expanding to include a lot of unrelated changes, it might be a good idea to break it into smaller tasks (easier to review, less conflict with other PRs, shorter task cycle, better team visibility).

yuchenlintt (commented):
Hey there! Sorry about my duplicate work.

I'll get around to reviewing this PR properly once the PR queue gets merged in to dev and you have a chance to resolve merge conflicts.

vanruch:
Are we sure we want this change? Seems a bit arbitrary to have TVL calculation to PoolFactory

yuchenlintt:
> Are we sure we want this change? Seems a bit arbitrary to have TVL calculation to PoolFactory

I think this is still the best place to put it. Maybe PoolFactory needs a better name (PoolDirectory? PoolRegistry?), but regardless a TVL view probably belongs with the only canonical list of TVL pools in our repo.

---

A minor philosophical note: in Solidity I think there's a loose relationship between EVM contracts and Object Oriented classes. We would _like_ to think of these as pretty similar, but the gas overhead and ABI fragility of calling a separate contract is more often not worth it, whereas in a language like Java we normally wouldn't even consider the costs of a function call to another object. I think this is why our repo more naturally ends up with monolithic contracts that do a bunch of marginally related things, instead of the clean, focused classes that OO tries to encourage.

Or maybe we're just lazy hackers :)

contracts/truefi2/TrueRateAdjuster.sol
```diff
@@ -87,13 +88,10 @@ contract TrueRateAdjuster is ITrueRateAdjuster, UpgradeableClaimable {
     /// @dev array of pools used to calculate TrueFi TVL
     ITrueFiPool2[] public tvlPools;
 
-    // ======= STORAGE DECLARATION END ============
-
-    /// @dev emit `pool` when adding to TVL
-    event PoolAddedToTVL(ITrueFiPool2 pool);
+    /// @dev used for TVL calculations
+    IPoolFactory public factory;
```
MiksuJak, line 92: As we have more than one contract which serves as a factory, I'd be in favour of renaming this to `poolFactory`.

test/truefi2/lines-of-credit/TrueRateAdjuster.test.ts
```diff
@@ -3,36 +3,36 @@ import { Wallet } from 'ethers'
...
 
 import { deployMockContract, MockContract, MockProvider, solidity } from 'ethereum-waffle'
-import { ITimeAveragedBaseRateOracleJson, ITrueFiPool2WithDecimalsJson, SpotBaseRateOracleJson } from 'build'
+import {
+  ITimeAveragedBaseRateOracleJson,
+  ITrueFiPool2WithDecimalsJson,
+  SpotBaseRateOracleJson,
+} from 'build'
+
+import { MockPoolFactory__factory } from 'build/types/factories/MockPoolFactory__factory'
```
MiksuJak, line 35: This could be imported from `contracts`

test/truefi2/lines-of-credit/TrueRateAdjuster.test.ts
```diff
@@ -45,6 +45,7 @@ describe('TrueRateAdjuster', () => {
   let asset: MockErc20Token
   let mockSpotOracle: MockContract
   let oracle: TimeAveragedBaseRateOracle
+  let mockFactory: IPoolFactory
```
MiksuJak, line 48: I think importing `IPoolFactory` is not necessary. This variable can just be of type `MockPoolFactory`

contracts/truefi2/PoolFactory.sol
```diff
@@ -295,4 +298,16 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
         safu = _safu;
         emit SafuChanged(_safu);
     }
+
+    /**
+     * @dev Calculate total TVL in USD
+     * @return _tvl TVL for all supported pools
+     */
+    function tvl() public override view returns (uint256) {
```
yuchenlintt, line 306:
```suggestion
    function supportedPoolsTVL() public override view returns (uint256) {
```

We should probably disambiguate which pools we're calculating TVL from.

Rationale: we have several other pool-related mappings to track in PoolFactory. It's complicated and I wish we could simplify this, but these have already been deployed.
```
    mapping(address => address) public pool;
    mapping(address => bool) public override isPool;
    mapping(address => bool) public isAllowed;
    mapping(address => mapping(address => address)) public singleBorrowerPool;
```

I'm worried we might forget in the future and think that `tvl()` is calculated from the `isPool() == true` set.

yuchenlintt (changes requested):

vanruch (changes requested):
There're no tests for tvl() method

contracts/truefi2/TrueLender2.sol
yuchenlintt: It looks like we can revert the `TrueLender2.sol` changes in this PR. Renaming from `pools` to `tvlPools` isn't necessary since we already know in context they're coming from `getSupportedPools()`.

test/truefi2/TrueLender2.test.ts
```diff
@@ -133,13 +130,12 @@ describe('TrueLender2', () => {
...
     await timeTravel(1)
-
     loan1 = await createLoan(loanFactory, borrower, pool1, 100000, YEAR, 100)
 
     loan2 = await createLoan(loanFactory, borrower, pool2, 500000, YEAR, 1000)
 
-    await rateAdjuster.addPoolToTVL(pool1.address)
-    await rateAdjuster.addPoolToTVL(pool2.address)
+    await poolFactory.supportPool(pool1.address)
+    await poolFactory.supportPool(pool2.address)
```
yuchenlintt, line 138: We already call `poolFactory.supportPool()` in lines 109-110 above. I think this might have come from a git merge.

test/utils/setupTruefi2.ts
```diff
@@ -84,27 +84,25 @@ export const setupTruefi2 = async (owner: Wallet, provider: MockProvider, custom
...
   const standardPool = poolImplementation.attach(await poolFactory.pool(standardToken.address))
-  await poolFactory.supportPool(standardPool.address)
   await standardPool.setOracle(standardTokenOracle.address)
 
   await rateAdjuster.setBaseRateOracle(standardPool.address, standardBaseRateOracle.address)
   await rateAdjuster.setBaseRateOracle(feePool.address, feeBaseRateOracle.address)
-  await rateAdjuster.addPoolToTVL(standardPool.address)
-  await rateAdjuster.addPoolToTVL(feePool.address)
+  await poolFactory.supportPool(standardPool.address)
+  await poolFactory.supportPool(feePool.address)
```
yuchenlintt, line 105: Let's delete these two lines and keep the `poolFactory.supportPool()` calls with poolFactory instead of rateAdjuster. This probably came from a git merge.

vanruch (approved):

yuchenlintt (approved):

skalermo (approved):

PR #938, opened 2021-09-03
Fix curve calc_token_amount calculation

Vashti-w:
calc_token_amount from curve is used incorrectly

Definition from curve ```def calc_token_amount(amounts: uint256[N_COINS], deposit: bool) -> uint256:```

Now, second argument is always ```true```.
When called from ```function withdraw(uint256 minAmount) external override onlyPool``` we need to calculate amount of pool tokens (Curve.fi yDAI/yUSDC/yUSDT/yTUSD - 0xdf5e0e81dff6faf3a7e52ba697820c5e32d806a8) to be burned to get ```minAmount``` tokens back.

When curve pool is balanced, the difference in calculation is minimal and using ```.mul(1005).div(1000)``` few lines later solves most problems.
When curve pool is not balanced, ```curvePool.remove_liquidity_one_coin``` call can fail because ```minAmount``` argument is supplied but ```conservativeCurveTokenAmount``` will be too small.
This could cause temporary withdrawal DoS.

---
If this calculation is fixed, you should be able to replace ```.mul(1005).div(1000)``` with ```.mul(1001).div(1000)```, I noticed you already use ```.mul(999).div(1000)``` for deposit calculation.

### DIFF
--- a/contracts/truefi2/strategies/CurveYearnStrategy.sol
+++ b/contracts/truefi2/strategies/CurveYearnStrategy.sol
@@ -264,6 +264,14 @@ contract CurveYearnStrategy is UpgradeableClaimable, ITrueStrategy {
     }
 
     /**
+     * @dev DEPRECATED For new code, please determine which of
+     * {calcTokenDepositAmount(), calcTokenWithdrawAmount()} is appropriate.
+     */
+    function calcTokenAmount(uint256 currencyAmount) external view returns (uint256) {
+        return calcTokenDepositAmount(currencyAmount);
+    }
+
+    /**
      * @dev Expected amount of minted or burnt Curve.fi yDAI/yUSDC/yUSDT/yTUSD tokens.
      * @param currencyAmount amount to calculate for
      * @param isDeposit true for deposits=mint, false for withdrawals=burn
--- a/contracts/truefi2/strategies/CurveYearnStrategy.sol
+++ b/contracts/truefi2/strategies/CurveYearnStrategy.sol
@@ -125,7 +125,7 @@ contract CurveYearnStrategy is UpgradeableClaimable, ITrueStrategy {
         amounts[tokenIndex] = totalAmount;
 
         token.safeApprove(address(curvePool), totalAmount);
-        uint256 conservativeMinAmount = calcTokenAmount(totalAmount, true).mul(999).div(1000);
+        uint256 conservativeMinAmount = calcTokenDepositAmount(totalAmount).mul(999).div(1000);
         curvePool.add_liquidity(amounts, conservativeMinAmount);
 
         // stake yCurve tokens in gauge
@@ -143,7 +143,7 @@ contract CurveYearnStrategy is UpgradeableClaimable, ITrueStrategy {
      */
     function withdraw(uint256 minAmount) external override onlyPool {
         // get rough estimate of how much yCRV we should sell
-        uint256 roughCurveTokenAmount = calcTokenAmount(minAmount, false);
+        uint256 roughCurveTokenAmount = calcTokenWithdrawAmount(minAmount);
         uint256 yBalance = yTokenBalance();
         require(roughCurveTokenAmount <= yBalance, "CurveYearnStrategy: Not enough Curve liquidity tokens in pool to cover borrow");
         // Try to withdraw a bit more to be safe, but not above the total balance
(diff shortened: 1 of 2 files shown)


contracts/truefi2/strategies/CurveYearnStrategy.sol
```diff
@@ -125,7 +125,7 @@ contract CurveYearnStrategy is UpgradeableClaimable, ITrueStrategy {
         amounts[tokenIndex] = totalAmount;
 
         token.safeApprove(address(curvePool), totalAmount);
-        uint256 conservativeMinAmount = calcTokenAmount(totalAmount).mul(999).div(1000);
+        uint256 conservativeMinAmount = calcTokenAmount(totalAmount, true).mul(999).div(1000);
```
yuchenlintt, line 128: Our team tries to adhere to test-driven development. When we fix an issue, we typically write unit or integration tests for the fix. This has three benefits:
1. A failing test confirms the original issue exists.
2. Making the test pass confirms that our fix actually resolves the issue.
3. Leaving the test in our CI framework helps us catch regressions when we refactor in the future.

Could you write a unit test to check we're calling the correct `curve().calc_token_amount(amount, true)` under the deposit section? https://github.com/trusttoken/smart-contracts/blob/65b8699c4b44e5e18b4ff733ca4e0ccc52070ff6/test/truefi2/strategies/CurveYearnStrategy.test.ts#L102

I think all that your test case needs to do is use Waffle's `deployMockContract(owner, ICurveJson.abi)` to deploy a mock, pass this in to a new TestCurveStrategy, and check that the correct arguments get called in the mock `ICurve`.

contracts/truefi2/strategies/CurveYearnStrategy.sol
```diff
@@ -143,7 +143,7 @@ contract CurveYearnStrategy is UpgradeableClaimable, ITrueStrategy {
      */
     function withdraw(uint256 minAmount) external override onlyPool {
         // get rough estimate of how much yCRV we should sell
-        uint256 roughCurveTokenAmount = calcTokenAmount(minAmount);
+        uint256 roughCurveTokenAmount = calcTokenAmount(minAmount, false);
```
yuchenlintt, line 146: Ditto above, but checking for a call to `curve().calc_token_amount(amount, false)`, and added under the withdraw section: https://github.com/trusttoken/smart-contracts/blob/65b8699c4b44e5e18b4ff733ca4e0ccc52070ff6/test/truefi2/strategies/CurveYearnStrategy.test.ts#L136

yuchenlintt (changes requested):
Hi Vashti-w!

Thank you very kindly for finding and proposing this fix to our Curve calculations. I believe this makes you among our first community contributors! We're honored to have your help.

After reviewing your PR internally, we've determined that your change makes sense to us, and we'd like to move forward with it.

I've added some review comments to bring this PR in line with our repo style and contribution standards. Feel free to address them, or let us know if you'd like us to take over this change.

Many thanks again! With diligent community members like you, we're in a far better position to catch oversights that our code review processes and audits might overlook!


Yuchen

Vashti-w:
I have updated codestyle and added requested tests.
Feel free to update the code yourself if it is faster for you, otherwise I can do it.

---

Perhaps we want to keep `function calcTokenAmount(uint256 currencyAmount) public view returns (uint256)` in the code for backwards compatibility. Perhaps it is used by frontend?

yuchenlintt (dismissed):
Looks perfect to me, thanks again for the fix!

If we're being responsible, I suppose we should be careful about ABI compatibility. You can add an external `calcTokenAmount()` view that delegates to `calcTokenDepositAmount()`. For its docstring, maybe say something like this?
```
DEPRECATED For new code, please determine which of {calcTokenDepositAmount(), calcTokenWithdrawAmount()} is appropriate.
```

We're in a bit of a time crunch at the moment, so I'm also approving now in case you'd rather check this in quickly and skip the ABI change.

Vashti-w:
Pushed fix for ABI compatibility

Vashti-w:
The test fail looks unrelated, it succeeded before the last commit. I do not have permissions to rerun it.

yuchenlintt (approved):
LGTM, thanks!

MiksuJak (approved):

vanruch (approved):
Thanks a lot for the PR @Vashti-w LGTM

PR #940, opened 2021-09-07
Simplify yarn scripts

MiksuJak:
This PR includes 2 changes:
1. Create yarn script `deploy:dryrun`, which replaces a long bash command that is run on CircleCI. This should ease running it locally for troubleshooting, so it does not need to be copied from `config.yml`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,7 @@ jobs:
     steps:
       - attach_workspace:
           at: .
-      - run: yarn deploy:truefi --network kovan --dry-run --yes && yarn deploy:truefi2 --network kovan --dry-run --yes && yarn deploy:farms --network kovan --dry-run --yes
+      - run: yarn deploy:dryrun
   integration:
     docker:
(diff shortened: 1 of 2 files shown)


yuchenlintt:
1. LGTM, thanks for this cleanup!
2. I believe the CircleCI config is slightly different from package.json: `.mocharc.integration-ci.js` vs `.mocharc.integration.js`. This is causing the integration failure on this PR because we have some flaky tests that we don't run in CI.

MiksuJak:
> 1. LGTM, thanks for this cleanup!
> 2. I believe the CircleCI config is slightly different from package.json: `.mocharc.integration-ci.js` vs `.mocharc.integration.js`. This is causing the integration failure on this PR because we have some flaky tests that we don't run in CI.

I honestly did not notice `-ci` in the option. I'll revert last commit. Same tests fail on dev now anyways.

yuchenlintt (approved):

vanruch (approved):

PR #941, opened 2021-09-07
Add poolFactory setter to Liquidator2

MiksuJak:


yuchenlintt:
Thanks! Almost there...

After the audit we've started adding zero address checks in our setters. I think our multisig scrutiny makes the zero address checks a little bit redundant, but it's still good to keep auditors happy.

vanruch (approved):

yuchenlintt (approved):

PR #942, opened 2021-09-07
Make DebtToken functional

vanruch:
I removed all functions from IDebtToken interface that are not essential to liquidation flow. If we decide we need them, it can be added later

### DIFF
--- a/contracts/truefi2/DebtToken.sol
+++ b/contracts/truefi2/DebtToken.sol
@@ -57,6 +57,9 @@ contract DebtToken is IDebtToken, ERC20 {
      * @param _amount amount to redeem
      */
     function redeem(uint256 _amount) external override {
+        // TODO remove the require checks when Awaiting/Funded/Withdrawn enum values are removed from IDebtToken
+        require(status >= Status.Defaulted, "DebtToken: The debt has not defaulted yet");
+
         uint256 amountToReturn = _amount.mul(_balance()).div(totalSupply());
         redeemed = redeemed.add(amountToReturn);
         _burn(msg.sender, _amount);
--- a/contracts/truefi2/DebtToken.sol
+++ b/contracts/truefi2/DebtToken.sol
@@ -16,7 +16,6 @@ contract DebtToken is IDebtToken, ERC20 {
 
     address public liquidator;
     ITrueFiPool2 public override pool;
-    ERC20 public override token;
 
     uint256 public redeemed;
 
@@ -32,9 +31,8 @@ contract DebtToken is IDebtToken, ERC20 {
 
     /**
      * @dev Emitted when loan gets liquidated
-     * @param status Final loan status
      */
-    event Liquidated(Status status);
+    event Liquidated();
 
     function initialize(
         ITrueFiPool2 _pool,
@@ -46,7 +44,6 @@ contract DebtToken is IDebtToken, ERC20 {
         ERC20.__ERC20_initialize("TrueFi Debt Token", "DEBT");
 
         pool = _pool;
-        token = _pool.token();
         borrower = _borrower;
         liquidator = _liquidator;
         debt = _debt;
@@ -63,7 +60,7 @@ contract DebtToken is IDebtToken, ERC20 {
         uint256 amountToReturn = _amount.mul(_balance()).div(totalSupply());
         redeemed = redeemed.add(amountToReturn);
         _burn(msg.sender, _amount);
-        token.safeTransfer(msg.sender, amountToReturn);
+        token().safeTransfer(msg.sender, amountToReturn);
 
         emit Redeemed(msg.sender, _amount, amountToReturn);
     }
@@ -77,7 +74,7 @@ contract DebtToken is IDebtToken, ERC20 {
 
         status = Status.Liquidated;
 
-        emit Liquidated(status);
+        emit Liquidated();
     }
 
     /**
@@ -97,6 +94,14 @@ contract DebtToken is IDebtToken, ERC20 {
         return _balance();
     }
 
+    function token() public override view returns (ERC20) {
+        return pool.token();
+    }
+
+    function decimals() public override view returns (uint8) {
+        return token().decimals();
+    }
+
     function version() external override pure returns (uint8) {
         return 1;
     }
@@ -106,6 +111,6 @@ contract DebtToken is IDebtToken, ERC20 {
      * @return token balance of this contract
      */
     function _balance() internal view returns (uint256) {
-        return token.balanceOf(address(this));
+        return token().balanceOf(address(this));
     }
 }
--- a/contracts/truefi2/DebtToken.sol
+++ b/contracts/truefi2/DebtToken.sol
@@ -4,37 +4,108 @@ pragma solidity 0.6.10;
 import {ERC20} from "../common/UpgradeableERC20.sol";
 import {IDebtToken} from "./interface/ILoanToken2.sol";
 import {ITrueFiPool2} from "./interface/ITrueFiPool2.sol";
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
 
 contract DebtToken is IDebtToken, ERC20 {
-    function borrower() external override view returns (address) {}
+    using SafeMath for uint256;
+    using SafeERC20 for ERC20;
 
-    function amount() external override view returns (uint256) {}
+    address public override borrower;
+    uint256 public override debt;
 
-    function debt() external override view returns (uint256) {}
+    address public liquidator;
+    ITrueFiPool2 public override pool;
+    ERC20 public override token;
 
-    function pool() external override view returns (ITrueFiPool2) {}
+    uint256 public redeemed;
 
-    function status() external override view returns (Status) {}
+    Status public override status;
 
-    function redeem(uint256 _amount) external override {}
+    /**
+     * @dev Emitted when a DebtToken is redeemed for underlying tokens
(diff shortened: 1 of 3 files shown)


contracts/truefi2/DebtToken.sol
```diff
@@ -4,37 +4,108 @@ pragma solidity 0.6.10;
...
+
+        emit Redeemed(msg.sender, _amount, amountToReturn);
+    }
+
+    /**
+     * @dev Liquidate the loan if it has defaulted
+     */
+    function liquidate() external override {
+        require(status == Status.Defaulted, "DebtToken: Current status should be Defaulted");
+        require(msg.sender == liquidator, "DebtToken: Caller is not the liquidator");
```
yuchenlintt, line 76: Nice! I hadn't realized the `onlyDefaulted onlyLiquidator` modifiers were only used for this one function. Better to inline them like this since we're not worried about dedups.

contracts/truefi2/DebtToken.sol
```diff
@@ -4,37 +4,108 @@ pragma solidity 0.6.10;
...
+        status = Status.Defaulted;
+        _mint(_lender, _debt);
+    }
+
+    /**
+     * @dev Redeem DebtToken balances for underlying token
+     * Can only call this function after the loan is Closed
+     * @param _amount amount to redeem
+     */
+    function redeem(uint256 _amount) external override {
```
yuchenlintt, line 62: We should still check `Status.Defaulted` || `Status.Liquidated` (possibly even `Status.Settled` if we're planning to have LoanToken2 inherit from DebtToken). I'm okay with inlining the check though.

contracts/truefi2/DebtToken.sol
```diff
@@ -4,37 +4,108 @@ pragma solidity 0.6.10;
...
+        ITrueFiPool2 _pool,
+        address _lender,
+        address _borrower,
+        address _liquidator,
+        uint256 _debt
+    ) external initializer {
+        ERC20.__ERC20_initialize("TrueFi Debt Token", "DEBT");
+
+        pool = _pool;
+        token = _pool.token();
```
yuchenlintt, line 49: Would it make sense to make `token()` a view that delegates to `pool.token()`?

I'm not sure how much we care about the gas saved from caching the `token` address in the loan itself. I don't have a strong opinion on this, just noticed a possibility to reduce contract state now that we're reimplementing. Please push back if you think it still makes sense to cache.

test/truefi2/lines-of-credit/DebtToken.test.ts
```diff
@@ -0,0 +1,166 @@
...
+      expect(await debtToken.balanceOf(lender.address)).to.equal(debt)
+      expect(await debtToken.totalSupply()).to.equal(debt)
+    })
+
+    it('sets status to Defaulted', async () => {
+      expect(await debtToken.status()).to.equal(DebtTokenStatus.Defaulted)
+    })
+  })
+
+  describe('liquidate', () => {
```
yuchenlintt, line 68: We should add back the checks to make sure this reverts if Status < 4. If that's not possible without introducing needless complexity then feel free to close this comment.

test/truefi2/lines-of-credit/DebtToken.test.ts
```diff
@@ -0,0 +1,166 @@
...
+    it('sets status to liquidated', async () => {
+      await debtToken.connect(safu).liquidate()
+      expect(await debtToken.status()).to.equal(DebtTokenStatus.Liquidated)
+    })
+  })
+
+  describe('Redeem', () => {
+    beforeEach(async () => {
+      await token.mint(borrower.address, parseEth(100))
+    })
```
yuchenlintt, line 89: Same as above: We should add back tests to make sure this isn't called if Status < 4, if possible without too much hassle.

yuchenlintt (changes requested):

test/truefi2/lines-of-credit/DebtToken.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    it('reverts if not called by liquidator', async () => {
+      await expect(debtToken.liquidate())
+        .to.be.revertedWith('DebtToken: Caller is not the liquidator')
+    })
+
+    it('sets status to liquidated', async () => {
+      await debtToken.connect(safu).liquidate()
+      expect(await debtToken.status()).to.equal(DebtTokenStatus.Liquidated)
+    })
```
MiksuJak, line 83: Add event emission test

test/truefi2/lines-of-credit/DebtToken.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    it('sets pool address', async () => {
+      expect(await debtToken.pool()).to.equal(poolAddress)
+    })
+
+    it('sets borrowers debt', async () => {
+      expect(await debtToken.debt()).to.equal(debt)
+    })
+
```
MiksuJak, line 51: Add test for `debtToken.borrower`

contracts/truefi2/DebtToken.sol
```diff
@@ -4,37 +4,108 @@ pragma solidity 0.6.10;
...
+        _mint(_lender, _debt);
+    }
+
+    /**
+     * @dev Redeem DebtToken balances for underlying token
+     * Can only call this function after the loan is Closed
+     * @param _amount amount to redeem
+     */
+    function redeem(uint256 _amount) external override {
+        uint256 amountToReturn = _amount.mul(_balance()).div(totalSupply());
```
MiksuJak, line 63: Since the debt would be immutable after default, shouldn't redeeming be 1:1?

contracts/truefi2/DebtToken.sol
vanruch: DebtToken can have only 2 states now. I'd wait with adding any LoanToken-specific functionality to when we start deprecating LTs and follow YAGNI principle.

test/truefi2/lines-of-credit/DebtToken.test.ts
vanruch: I don't see a reason to add a check for a state that's not reachable. We can't even test it

contracts/truefi2/DebtToken.sol
```diff
@@ -4,37 +4,108 @@ pragma solidity 0.6.10;
...
+        address _liquidator,
+        uint256 _debt
+    ) external initializer {
+        ERC20.__ERC20_initialize("TrueFi Debt Token", "DEBT");
+
+        pool = _pool;
+        token = _pool.token();
+        borrower = _borrower;
+        liquidator = _liquidator;
+        debt = _debt;
```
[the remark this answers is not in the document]
vanruch: Not a conversion but added `decimals()` view

contracts/truefi2/DebtToken.sol
vanruch: I kept it same as LT for now, we will change it later

contracts/truefi2/DebtToken.sol
yuchenlintt: In general I agree with YAGNI, but in this case IDebtToken still has more than two possible states -- we plan to deprecate the others, but they haven't even been marked for deprecation yet.

It might happen to be that only two states are reachable in current code, but it would be easy to accidentally reintroduce the other states without noticing `redeem()` would need to guard against them. As long as it's possible for this enum to have unexpected values (i.e., 0-3), I think we should protect from those values.

For now, perhaps we could add the require and prepend with a comment, something like this?
```
// TODO remove the require checks when Awaiting/Funded/Withdrawn enum values are removed from IDebtToken
require(status() >= 4, "error log")
```

As soon as we remove the other statuses from the enum, we should notice test failures, see the TODO here, and be ready to remove these two lines.

contracts/truefi2/DebtToken.sol
```diff
@@ -4,37 +4,108 @@ pragma solidity 0.6.10;
...
+    }
+
+    /**
+     * @dev Get currency token balance for this contract
+     * @return token balance of this contract
+     */
+    function _balance() internal view returns (uint256) {
+        return token.balanceOf(address(this));
+    }
 }
```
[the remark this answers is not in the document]
yuchenlintt: Agreed. It's unclear yet exactly what mechanism litigation will use to force debt repayment (fiat to banks vs. ERC20 transfer to this LoanToken vs. ERC20 transfer to another contract -- it depends on jurisdiction + terms of settlement), so I'm inclined to avoid adding these yet.

+1 for creating tasks so we can bring these up for discussion though.

(My opinion) I'm inclined to avoid reintroducing `reclaim()` at all:
1. If a borrower accidentally overpays then we can `redeem()` and manually return funds to them.
2. During litigation, legal counsel would likely demand interest be accrued on top of the original debt amount owed. If the borrower ends up settling with an ERC20 transfer to the DebtToken, we wouldn't want them to be able to withdraw that interest via `reclaim()`.

test/truefi2/lines-of-credit/DebtToken.test.ts
yuchenlintt: SGTM, I'll close this comment then.

test/truefi2/lines-of-credit/DebtToken.test.ts
yuchenlintt: I'll go ahead and resolve this comment too since we'd have to test against an unreachable state.

However, following this comment: https://github.com/trusttoken/smart-contracts/pull/942/files#r703661740, the contract should still guard against the unreachable state with a require check to make sure it remains blocked, even if we accidentally introduced new code to violate our invariant.

yuchenlintt (approved):

MiksuJak (approved):

PR #945, opened 2021-09-08
Make reclaim return same amount of tokens as DebtTokens burnt

vanruch:
Since DebtTokens are not expected to be tradeable, we decided to make reclaim not proportional. This will allow us to reclaim partially repaid debt


barrutko (approved):

yuchenlintt:
LGTM (post-merge)

PR #947, opened 2021-09-08
 LoanFactory deploys DebtTokens

skalermo:


contracts/truefi2/LoanFactory2.sol
```diff
@@ -135,6 +151,11 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         emit LoanTokenCreated(newToken);
     }
 
+    function createDebtToken() external onlyTCA {
+        address newToken = address(new DebtToken());
+        emit DebtTokenCreated(newToken);
+    }
```
vanruch, line 157: Why don't we use minimal proxies for debt tokens?

skalermo: I was gonna do that in another PR, but I can as well here.

skalermo: We decided to do so in another PR.

contracts/truefi2/LoanFactory2.sol
```diff
@@ -135,6 +156,18 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         emit LoanTokenCreated(newToken);
     }
 
+    function createDebtToken(
+        ITrueFiPool2 _pool,
+        address _borrower,
+        uint256 _debt
+    ) external onlyTCA {
+        address newToken = address(new DebtToken());
+        DebtToken(newToken).initialize(_pool, lender, _borrower, liquidator, _debt);
```
yuchenlintt, line 165: I'm sure this will be refactored in the next PR when we switch to proxies. No need to cast to address and then back to DebtToken.

contracts/truefi2/LoanFactory2.sol
```diff
@@ -33,6 +35,9 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
     // @dev Track Valid LoanTokens
     mapping(address => bool) public override isLoanToken;
 
+    // @dev Track valid debtTokens
+    mapping(address => bool) public isDebtToken;
```
yuchenlintt, line 39: Minor refactor (task added to bottom of backlog): Let's map from DebtToken to bool to save unnecessary casts elsewhere.

yuchenlintt (approved):
LGTM! I've added backlog tasks for minor refactors, but that shouldn't block this PR.

vanruch (approved):
Agree with @yuchenlintt's comments and that it can be applied later

PR #944, opened 2021-09-08
Call default on TCA

MiksuJak:

PR #948, opened 2021-09-09
TCA default review follow up to #944

yuchenlintt:
Changes to TCA default:
- ~~per {borrower} instead of per {borrower, pool}~~
- onlyOwner
- check and unlock/lock borrowing mutex
- OR conditions instead of AND
- add {not allowed, ineligible, min score, interest overdue} conditions
- emit a descriptive event

### DIFF
--- a/contracts/truefi2/TrueCreditAgency.sol
+++ b/contracts/truefi2/TrueCreditAgency.sol
@@ -401,18 +401,22 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
         );
         if (!isBorrowerAllowed[borrower]) {
             _enterDefault(borrower);
+            return;
         }
         if (creditOracle.status(borrower) == ITrueFiCreditOracle.Status.Ineligible) {
             _enterDefault(borrower);
+            return;
         }
         if (creditOracle.score(borrower) < minCreditScore) {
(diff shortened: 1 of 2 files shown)


vanruch:
@yuchenlintt Can you elaborate why do we want it to behave that way? I don't quite understand why disallowing the borrower to borrow more or reducing the score opens the possibility of default even if the borrower keeps paying the instalments in time. 
Also should the default on a single line make the rest default as well? We probably want to avoid unnecessary defaults to save staker's and SAFU's money.

yuchenlintt:
We resolved this discussion at our standup today. Conclusions:
1. We're reverting to defaulting one pool at a time. According to Biz we should be defaulting all at once, but for granularity of control we'll launch with single pool defaults.
2. Defaulting (after a grace period) due to ineligibility, credit score drop, credit limit drop, etc. are all hard Biz requirements. These have to be kept.

This PR should be ready for review again. I'll rebase the other PRs on top of these changes, and then those should be ready as well.

If there are review comments that would only be a couple lines to fix (whether critical or non-critical), I would recommend commenting here but submitting fixes in follow-up PRs. This would reduce our burden in rebasing/merging this current queue of PRs.

vanruch (approved):

barrutko (approved):

PR #949, opened 2021-09-09
Add TCA DefaultReason.LimitExceeded

yuchenlintt:


contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -376,16 +376,16 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
...
         if (borrowed[pool][msg.sender] == 0) {
             nextInterestRepayTime[pool][msg.sender] = 0;
         }
+        pokeBorrowLimitTimer(pool, msg.sender);
 
         // transfer token from sender wallets
         _repay(pool, amount);
+        if (totalBorrowed(msg.sender) == 0) {
+            borrowingMutex.unlock(msg.sender);
+        }
```
yuchenlintt, line 388: Moving these lines isn't strictly relevant to this PR, but it _is_ an important change to make.

The original version violates checks-effects-interactions and possibly allows a reentrancy attack. Unlocking the borrowing mutex is an outside interaction that should be executed only after all other transfers have completed, so that transferees don't have an opportunity to execute borrow function calls on an unlocked borrower.

barrutko (dismissed):

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -444,6 +449,16 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
...
     }
 
+    function pokeBorrowLimitTimer(ITrueFiPool2 pool, address borrower) public {
+        if (borrowLimit(pool, borrower) > 0) {
+            overBorrowLimitTime[pool][borrower] = 0;
+            return;
+        }
+        if (overBorrowLimitTime[pool][borrower] == 0) {
+            overBorrowLimitTime[pool][borrower] = block.timestamp;
+        }
```
vanruch, line 459: Won't this also be triggered when the borrowed amount is strictly equal to max borrow limit?

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -423,6 +423,11 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
             _enterDefault(pool, borrower, DefaultReason.InterestOverdue);
             return;
         }
+        uint256 overBorrowLimit = overBorrowLimitTime[pool][borrower];
```
MiksuJak, line 426: I wouldn't create a separate variable for it, especially because its name can be confused with asset amount that is over limit. Also it is used only once in the following if statement.

skalermo: It would save some gas, because hash function will be called fewer times. Maybe more descriptive name would do? Something like `_overBorrowLimitTime`?

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -423,6 +423,11 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
...
+        if (overBorrowLimit != 0 && defaultTime >= overBorrowLimit) {
+            _enterDefault(pool, borrower, DefaultReason.LimitExceeded);
```
vanruch, line 428: I think we also need to check if it's above the limit again

MiksuJak: or maybe just reset the timer whenever borrower repays below the limit

test/truefi2/lines-of-credit/TrueCreditAgency.test.ts
```diff
@@ -915,57 +943,72 @@ describe('TrueCreditAgency', () => {
...
-      it('borrower interest is overdue', async () => {
+      it('has overdue interest', async () => {
         await creditOracle.setEligibleForDuration(borrower.address, YEAR)
         await timeTravel(MONTH + DAY * 3 + 1)
         await expect(creditAgency.enterDefault(tusdPool.address, borrower.address))
           .to.emit(creditAgency, 'EnteredDefault')
           .withArgs(borrower.address, DefaultReason.InterestOverdue)
       })
+
+      it('has exceeded borrow limit', async () => {
```
skalermo, line 1004: Same as my comment above. Imo it should be `... borrow time limit'`

contracts/truefi2/TrueCreditAgency.sol
vanruch: actually it's done it this PR so it doesn't seem to be a problem

yuchenlintt:
Thanks @MiksuJak for keeping this PR moving forward while I was asleep!

I will admit though that I'm somewhat perplexed at the lack of progress in merging the PR queue in the past 6 hours. While I typically think variable naming and edge cases are important to get right, were these concerns worth blocking our other ~8 PRs to get done now? I think we could have filed these as follow-up.

contracts/truefi2/TrueCreditAgency.sol
```diff
@@ -32,7 +32,7 @@ contract TrueCreditAgency is UpgradeableClaimable, ITrueCreditAgency {
     using SafeERC20 for ERC20;
     using SafeMath for uint256;
 
-    enum DefaultReason {NotAllowed, Ineligible, BelowMinScore, InterestOverdue}
+    enum DefaultReason {NotAllowed, Ineligible, BelowMinScore, InterestOverdue, LimitExceeded}
```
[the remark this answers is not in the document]
yuchenlintt: Hmm. I agree that `LimitExceeded` might be vague, but the phrase "time limit" has a different meaning in English that we're colliding with. "time limit" implies a purely temporal limit. For instance, you might say "your time limit for this exam is 30 minutes".

However what we want to express is that the _amount_ limit has been exceeded for some _grace period_ of time. I'll file a follow-up PR at the bottom of our backlog :)

MiksuJak (approved):

vanruch (approved):

PR #950, opened 2021-09-09
TCA default deploys DebtTokens

yuchenlintt:


vanruch (dismissed):

yuchenlintt:
I had to leave some unit tests as TODO items, but I think this PR should be checked in anyway to unblock other pending tasks.

After this is merged, we still need to follow up and fix some issues:
1. Allow borrower to enter LoC default for multiple pools. Currently we lock the mutex away for the first default. But that means we can't call `enterDefault()` again because the mutex is no longer locked to TCA. Some considerations:
    -  a) We must lock the borrowing mutex away from TCA, in order to prevent the borrower from withdrawing again from the same pool after they've just defaulted. (For instance, if a borrower defaulted due to a missed interest payment, then they could just `repay(0)` to clear the timer and start withdrawing again.)
    - b) We must also prevent the borrower from withdrawing from other pools after the first pool defaults. I admit that forcing a default on other pools may not be great for SAFU purposes, but at least they shouldn't be able to automatically withdraw again after they've already defaulted one asset.
    - c) After a default under our current implementation, a borrower's `borrowLimit()` is _increased_ by their defaulted debt. So it would be doubly dangerous to let them withdraw anything.
    - d) We shouldn't skip the mutex check/unlock()/lock() for defaults. Doing so would break our invariants and vastly complicate our assumptions. Could we now LoC default an FTL borrower, who has never opened an LoC before? Could we now default the same {pool, borrower} tuple multiple times? In the future, if we permitted borrowers to simultaneously open LoCs and FTLs (as has been proposed), would we remember to fix this code? Otherwise a malicious borrower could lock to an FTL just before we default their LoC, allow the LoC to default without moving the lock, and then unlock their FTL to be able to withdraw from their "defaulted" LoC pool again.
    - e) With mutexes locked to multiple DebtTokens, we need to ensure the borrower remains locked to the _last_ DebtToken to be settled by litigation, before they can borrow again at that address.

Point (a) can be fixed by converting the mutex to 1 LoC / {pool, borrower}. TCA would own the mutex for every {pool, borrower}, and only the mutex for the defaulted pool gets locked to the DebtToken. But that doesn't solve point (b) or point (c). Point (d) is a real headache to reason about... I think for point (e) we should never unlock the mutex from a DebtToken, so this issue might be moot. After default, a borrower should register a new address and start over with a new credit history.

2. Return DebtToken address from LoanFactory2.createDebtToken() so TCA can lock the mutex to it, instead of the placeholder `address(1)`.

3. Figure out how to capture the new DebtToken address that's emitted, so unit tests can check that the new DebtToken has the right parameters.

MiksuJak (approved):

barrutko (approved):

vanruch (approved):

skalermo (approved):

PR #951, opened 2021-09-09
Use minimal proxy when creating debt tokens

skalermo:


contracts/truefi2/LoanFactory2.sol
```diff
@@ -161,7 +164,10 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         address _borrower,
         uint256 _debt
     ) external onlyTCA {
-        address newToken = address(new DebtToken());
+        address dtImplementationAddress = address(debtTokenImplementation);
```
barrutko, line 167: If we only use it as an address, why not just store (and set) it as an address?

vanruch (approved):

barrutko (approved):

MiksuJak (approved):

PR #957, opened 2021-09-10
Strip deprecated functions + state from FTLA

skalermo:


contracts/truefi2/FixedTermLoanAgency.sol
```diff
@@ -78,16 +78,6 @@ contract FixedTermLoanAgency is IFixedTermLoanAgency, UpgradeableClaimable {
 
     // ===== Voting parameters =====
```
yuchenlintt, line 79: You can also delete this comment!

yuchenlintt (changes requested):
Thanks for the change!

Since we've deprecated `pool.exit()`, I think we can also remove FTLA's `distribute()` function (and, recursively, anything else that it calls but nothing else does). Could you see if there are any other functions or variables we can remove like this?

yuchenlintt (commented):
We should also be able to remove anything related to TrueRatingAgencyV2 and loan rating from the unit tests.

barrutko (approved):

MiksuJak (approved):

contracts/truefi2/FixedTermLoanAgency.sol
```diff
@@ -540,19 +508,6 @@ contract FixedTermLoanAgency is IFixedTermLoanAgency, UpgradeableClaimable {
...
-        address pool
-    ) internal {
-        ILoanToken2[] storage _loans = poolLoans[ITrueFiPool2(pool)];
-        for (uint256 index = 0; index < _loans.length; index++) {
-            _transferLoan(_loans[index], recipient, numerator, denominator);
-        }
-    }
-
     // @dev Transfer (numerator/denominator)*balance of loan to the recipient
     function _transferLoan(
```
yuchenlintt, line 512: We probably could have inlined this function as well. It's only called in one place, with numerator == denominator == 1. So `loan.safeTransfer(recipient, loan.balanceOf(address(this)))` would have been good enough instead of an 8 line function.

yuchenlintt (commented):
Thanks for the cleanup!

I would have preferred to delay merging this PR though. When there's a tree of PRs to merge, it's better to prioritize the base of that tree, rather than merging each individual branch into the base. The changes from this PR look great! But because these are thematically different from the base PR, a reviewer will now have a (slightly) harder time sorting through the base branch and figuring out which changes came from merging this PR and which ones originally came from the base branch.

Just a tip for the future, if this situation comes up again.

PR #958, opened 2021-09-10
Copy validity checks from LF2.createLoanToken() to FTLA.fund()

yuchenlintt:


contracts/truefi2/FixedTermLoanAgency.sol
```diff
@@ -301,32 +312,41 @@ contract FixedTermLoanAgency is IFixedTermLoanAgency, UpgradeableClaimable {
...
         require(isTermBelowMax(term), "FixedTermLoanAgency: Loan's term is too long");
         require(isCredibleForTerm(term), "FixedTermLoanAgency: Credit score is too low for loan's term");
 
         uint256 amount = loanToken.amount();
+        require(amount > 0, "LoanFactory: Loans of amount 0, will not be approved");
         require(amount <= borrowLimit(pool, loanToken.borrower()), "FixedTermLoanAgency: Loan amount cannot exceed borrow limit");
 
+        uint256 apy = rate(pool, borrower, amount, term);
+        apy = loanToken.apy();
+        require(apy <= _maxApy, "LoanFactory: Calculated apy is higher than max apy");
```
skalermo, line 343: We don't need to call `rate()`, `apy` is already calculated during creation of loan token.

yuchenlintt: Yup, I know!

This is a placeholder so that migration will be easier in a follow-up task when we need to call `rate()` from `fund()`. Essentially I'm moving all the loan APY logic from LF2 into FTLA.

I can add a TODO comment to make that more clear though.

contracts/truefi2/FixedTermLoanAgency.sol
```diff
@@ -301,32 +312,41 @@ contract FixedTermLoanAgency is IFixedTermLoanAgency, UpgradeableClaimable {
      * @param loanToken LoanToken to fund
      */
     function fund(ILoanToken2 loanToken) external {
-        require(msg.sender == loanToken.borrower(), "FixedTermLoanAgency: Sender is not borrower");
-        ITrueFiPool2 pool = loanToken.pool();
+        // TODO Replace hardcoded _maxApy value with a param
+        uint256 _maxApy = 10000;
 
+        ITrueFiPool2 pool = loanToken.pool();
         require(factory.isSupportedPool(pool), "FixedTermLoanAgency: Pool not supported by the factory");
```
skalermo, line 319: Don't know if you are debugging this right know, but it seems tests are failing because `factory` is not set.

yuchenlintt: Thanks! Yeah, I'm currently debugging this and working on a fix.

yuchenlintt: I just merged in your FTLA deletion changes, and that seems to have fixed one test, but I'm still not sure how to fix the others.

yuchenlintt: Could someone else take a look at and debug these test failures?

I have a sense for the general cause (mock contracts not set properly), but I've spent several hours and haven't tracked down exactly which lines need to change, or to what.

skalermo (dismissed):

barrutko (dismissed):

MiksuJak (approved):

skalermo (approved):

PR #970, opened 2021-09-14
POC - Deploy DebtTokens when loan defaults

vanruch:
I didn't touch interfaces yet because it appears like a change that will affect a large chunk of code

### DIFF
--- a/contracts/truefi2/LoanFactory2.sol
+++ b/contracts/truefi2/LoanFactory2.sol
@@ -117,16 +117,16 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         _;
     }
 
-    modifier onlyFTLAOrLoan() {
-        require(
-            msg.sender == address(ftlAgency) || isLoanToken[msg.sender],
-            "LoanFactory: Caller is neither the fixed term loan agency not loan token"
-        );
+    modifier onlyFTLA() {
+        require(msg.sender == address(ftlAgency), "LoanFactory: Caller is not the fixed term loan agency");
         _;
     }
 
-    modifier onlyTCA() {
-        require(msg.sender == address(creditAgency), "LoanFactory: Caller is not the credit agency");
+    modifier onlyTCAOrLoanToken() {
+        require(
+            msg.sender == address(creditAgency) || isLoanToken[msg.sender],
+            "LoanFactory: Caller is neither credit agency nor loan"
+        );
         _;
     }
 
@@ -175,7 +175,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         uint256 _amount,
         uint256 _term,
         uint256 _apy
-    ) external override onlyFTLAOrLoan returns (ILoanToken2) {
+    ) external override onlyFTLA returns (ILoanToken2) {
         return _createFTLALoanToken(_pool, _borrower, _amount, _term, _apy);
     }
 
@@ -201,7 +201,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         ITrueFiPool2 _pool,
         address _borrower,
         uint256 _debt
-    ) external override onlyTCA returns (IDebtToken) {
+    ) external override onlyTCAOrLoanToken returns (IDebtToken) {
         address dtImplementationAddress = address(debtTokenImplementation);
         require(dtImplementationAddress != address(0), "LoanFactory: Debt token implementation should be set");
 
--- a/contracts/truefi2/LoanFactory2.sol
+++ b/contracts/truefi2/LoanFactory2.sol
@@ -187,7 +187,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         require(ltImplementationAddress != address(0), "LoanFactory: Loan token implementation should be set");
 
         address newToken = Clones.clone(ltImplementationAddress);
-        LoanToken2(newToken).initialize(_pool, borrowingMutex, _borrower, lender, ftlAgency, admin, liquidator, _amount, _term, _apy);
+        LoanToken2(newToken).initialize(_pool, borrowingMutex, _borrower, lender, ftlAgency, admin, _amount, _term, _apy);
         isLoanToken[newToken] = true;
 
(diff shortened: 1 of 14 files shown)


contracts/truefi2/LoanFactory2.sol
```diff
@@ -198,7 +198,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         ITrueFiPool2 _pool,
         address _borrower,
         uint256 _debt
-    ) external override onlyTCA {
+    ) external override onlyTCA returns (address) {
```
yuchenlintt, line 201: This function should return an IDebtToken to help with compiler type-checking. Otherwise all the places we want to call this would need to cast it manually. That would be more error-prone than casting once in the factory create function.

yuchenlintt: This probably compiles right now, but it will fail to run because you're calling createDebtToken() from LoanToken instead of TCA.

A general solution discussed in yesterday's meeting would be to delegate these checks to BorrowingMutex:
- TrueFiPool2's onlyLenderOrFTLAgencyOrTrueCreditAgency
- here in LoanFactory2's onlyTCA and onlyFTLA
- LoanToken2's onlyLenderOrFTLAgency

BorrowingMutex has two mappings that I think are perfect for gating which contracts can interact with pools and borrowers:
- `canLock(msg.sender)` -> onlyCanLock guards which agencies can initiate lending from supported pools
- `msg.sender == locker(borrower)` -> onlyLocker guards existing loans from being interfered with by unrelated callers (e.g., other agencies or malicious actors).

I think we should handle these conversions in a separate PR and check it in before continuing with this PR. Sorry I didn't notice this `onlyTCA` dependency while splitting tasks.

Do you think this PR sequencing makes sense?

yuchenlintt (changes requested):

contracts/truefi2/LoanToken2.sol
```diff
@@ -369,18 +360,23 @@ contract LoanToken2 is ILoanToken2, ERC20 {
     function enterDefault() external override onlyOngoing {
         require(!isRepaid(), "LoanToken2: cannot default a repaid loan");
         require(start.add(term).add(LAST_MINUTE_PAYBACK_DURATION) <= block.timestamp, "LoanToken2: Loan cannot be defaulted yet");
```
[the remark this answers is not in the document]
yuchenlintt: Done

contracts/truefi2/LoanFactory2.sol
```diff
@@ -187,7 +187,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         require(ltImplementationAddress != address(0), "LoanFactory: Loan token implementation should be set");
 
         address newToken = Clones.clone(ltImplementationAddress);
-        LoanToken2(newToken).initialize(_pool, borrowingMutex, _borrower, lender, ftlAgency, admin, liquidator, _amount, _term, _apy);
```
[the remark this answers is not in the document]
yuchenlintt: Task filed

contracts/truefi2/LoanFactory2.sol
yuchenlintt: Actually let's check this PR in with something like `onlyTCAOrLoan` using a simple manual guard for now, and we will move to the full borrowing mutex guard later.

contracts/truefi2/LoanFactory2.sol
```diff
@@ -117,8 +117,11 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         _;
     }
 
-    modifier onlyFTLA() {
-        require(msg.sender == address(ftlAgency), "LoanFactory: Caller is not the fixed term loan agency");
+    modifier onlyFTLAOrLoan() {
```
yuchenlintt, line 120: Let's revert this hunk (see discussion below).

contracts/truefi2/LoanFactory2.sol
```diff
@@ -172,7 +175,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         uint256 _amount,
         uint256 _term,
         uint256 _apy
-    ) external override onlyFTLA returns (ILoanToken2) {
+    ) external override onlyFTLAOrLoan returns (ILoanToken2) {
```
yuchenlintt, line 178: I think this should still be `onlyFTLA`, since only FixedTermLoanAgency should be able to create loan tokens. (It doesn't make sense for loan tokens to create other loan tokens.) I believe you meant to guard `createDebtToken()` behind `onlyTCAOrLoan`?

contracts/truefi2/LoanFactory2.sol
```diff
@@ -198,7 +201,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         ITrueFiPool2 _pool,
         address _borrower,
         uint256 _debt
-    ) external override onlyTCA {
+    ) external override onlyTCA returns (IDebtToken) {
```
yuchenlintt, line 204: `onlyTCAOrLoan`?

contracts/truefi2/LoanFactory2.sol
```diff
@@ -117,8 +117,11 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         _;
     }
 
-    modifier onlyFTLA() {
-        require(msg.sender == address(ftlAgency), "LoanFactory: Caller is not the fixed term loan agency");
+    modifier onlyFTLAOrLoan() {
+        require(
+            msg.sender == address(ftlAgency) || isLoanToken[msg.sender],
+            "LoanFactory: Caller is neither the fixed term loan agency not loan token"
```
yuchenlintt, line 123: This hunk is obsolete, but just a heads up in case you copy/paste to `onlyTCAOrLoan`.:
```suggestion
            "LoanFactory: Caller is neither the fixed term loan agency nor a loan token"
```

yuchenlintt (changes requested):

vanruch:
Created PR from a different branch

PR #976, opened 2021-09-16
Make SAFU's liquidation accept a list of DebtTokens and liquidate each of them

skalermo:


contracts/truefi2/SAFU.sol
```diff
@@ -81,37 +81,38 @@ contract SAFU is ISAFU, UpgradeableClaimable {
...
+     * @param loans Loans to be liquidated
      */
-    function liquidate(IDebtToken loan) external {
-        require(loanFactory.isCreatedByFactory(address(loan)), "SAFU: Unknown loan");
-        require(loan.status() == IDebtToken.Status.Defaulted, "SAFU: Loan is not defaulted");
-
-        ITrueFiPool2 pool = ITrueFiPool2(loan.pool());
-        IERC20 token = IERC20(pool.token());
+    function liquidate(IDebtToken[] calldata loans) external {
+        for (uint256 i = 0; i < loans.length; i++) {
```
MiksuJak, line 90: I think the code would be more readable if we everything into one for loop. It sort of breaks check-effect pattern, but we already decided that in this case we should skip this rule.

skalermo: After a discussion we agreed that it may require more changes than it seemed initially, so we'll address this issue in another PR.

MiksuJak (approved):

PR #980, opened 2021-09-17
Refactor totalBorrowed in TCA

MiksuJak:
Unfortunately this refactoring, would only work when we use really simple oracles like we do now. Whenever we introduce non-stable coin pool, it would break. I believe, that this kind of optimisation is just impossible for `totalBorrowed` and we should stick with less optimised view function.


yuchenlintt:
I agree, it looks like caching totalBorrowed like this locks in previous relative oracle prices, which isn't something we want.

PR #981, opened 2021-09-17
Add a check in liquidator, if all of provided loans are of a single borrower

MiksuJak:


contracts/truefi2/Liquidator2.sol
```diff
@@ -145,6 +145,10 @@ contract Liquidator2 is UpgradeableClaimable {
      */
     function liquidate(IDebtToken[] calldata loans) external {
         require(msg.sender == SAFU, "Liquidator: Only SAFU contract can liquidate a loan");
+        require(
+            areLoansOfSingleBorrower(loans),
```
yuchenlintt, line 149: How about this naming instead?
```suggestion
            allLoansHaveSameBorrower(loans),
```
If you like this, please update the error message text as well.

contracts/truefi2/Liquidator2.sol
```diff
@@ -181,4 +185,17 @@ contract Liquidator2 is UpgradeableClaimable {
     function getDefaultedValueInUsd(uint256 defaultedValue, ITrueFiPoolOracle oracle) internal view returns (uint256) {
         return oracle.tokenToUsd(defaultedValue);
     }
+
+    function areLoansOfSingleBorrower(IDebtToken[] calldata loans) internal view returns (bool) {
+        if (loans.length == 1) {
```
yuchenlintt, line 190: If we pass in an empty list this is vacuously true.
```suggestion
        if (loans.length <= 1) {
```
I think we don't need to revert on empty lists (an empty list shouldn't hurt, it just wastes the caller's gas). But maybe we should add a unit test to confirm an empty list doesn't transfer any TRU.

yuchenlintt (changes requested):

yuchenlintt (approved):

skalermo (approved):

PR #987, opened 2021-09-20
script to save avalanche ownership

hal909:


yuchenlintt:
Since we already ran this script, let's check it in as close as possible to its original state when it was run. I'll fix lint and then approve this PR.

Let's take all other follow-up discussion to internal channels.

yuchenlintt (approved):

MiksuJak (approved):

PR #1005, opened 2021-09-24
Use ILoanToken and IDebtToken types instead of addresses in LoanFactory isLoanToken + isDebtToken mappings

MiksuJak:


contracts/truefi2/LoanFactory2.sol
```diff
@@ -200,7 +200,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
             _term,
             _apy
         );
-        isLoanToken[newToken] = true;
+        isLoanToken[ILoanToken2(newToken)] = true;
```
barrutko, line 203: Why not cast once and for good in line 189?

contracts/truefi2/LoanFactory2.sol
```diff
@@ -216,13 +216,13 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
 
         address newToken = Clones.clone(dtImplementationAddress);
         DebtToken(newToken).initialize(_pool, msg.sender, _borrower, liquidator, _debt);
-        isDebtToken[newToken] = true;
+        isDebtToken[IDebtToken(newToken)] = true;
```
barrutko, line 219: If pervious comment applied, then here we could cast in line 217.

barrutko (approved):
Two small suggestions, but not necessarily to be implemented, so approved anyway.

vanruch (dismissed):

yuchenlintt (dismissed):

yuchenlintt (approved):

vanruch (approved):

PR #1006, opened 2021-09-24
Deploy DebtToken after loan defaults

vanruch:


contracts/truefi2/LoanToken2.sol
```diff
@@ -374,17 +370,25 @@ contract LoanToken2 is ILoanToken2, ERC20 {
...
-        emit Defaulted(_balance());
+        emit Defaulted(debtToken, debt.sub(repaid()));
     }
 
-    /**
-     * @dev Liquidate the loan if it has defaulted
-     */
-    function liquidate() external override onlyDefaulted onlyLiquidator {
-        status = Status.Liquidated;
+    function claimDebtToken(address loanHolder) public onlyDefaulted {
```
vanruch, line 383: We can't really test this function because we have no pool.exit() anymore. I thought it's worth adding in case loans will be distributed in the future.

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -631,7 +631,10 @@ contract TrueFiPool2 is ITrueFiPool2, IPauseableContract, ERC20, UpgradeableClai
      * @dev CreditAgency transfers DebtToken to the pool
      */
     function addDebt(IDebtToken debtToken, uint256 amount) external override {
-        require(msg.sender == address(creditAgency), "TruePool: Only TrueCreditAgency can add debtTokens");
+        require(
+            msg.sender == address(creditAgency) || ftlAgency.loanFactory().isLoanToken(msg.sender),
```
vanruch, line 635: I'm really not a fan of using chain calls like this. This definitely breaks some clean code principle. But the alternative is to add loanFactory to the pools (so adding setters and calling them on existing pools). Do you think we can proceed like this or do refactor?

contracts/truefi2/LoanToken2.sol
yuchenlintt: Yeah, you make a good point, especially since our minimal proxies can't upgrade LoanToken2s that have already been deployed.

However, we should also prevent repeated claimDebtToken() calls on the same LoanToken2 balance. What's the best way to achieve this?
1. Burn the loanHolder's LoanToken2 balance and also subtract from the loan balance?
2. Add another mapping to track whether a lender has already claimed their debtToken balance?

If this becomes complicated to implement, I propose we don't allow outside loan claims for now. We can always introduce it in a follow-up PR.

contracts/truefi2/LoanToken2.sol
```diff
@@ -374,17 +370,25 @@ contract LoanToken2 is ILoanToken2, ERC20 {
...
+        uint256 balance = balanceOf(loanHolder);
+        uint256 debtShare = debtToken.totalSupply().mul(balance).div(debt);
+
+        debtToken.safeTransfer(loanHolder, debtShare);
+    }
 
-        emit Liquidated(status);
+    function liquidate() external override {
+        revert("LoanToken2: Liquidation not supported");
```
yuchenlintt, line 391:
```suggestion
        revert("LoanToken2: Direct liquidation has been deprecated");
```

contracts/truefi2/TrueFiPool2.sol
yuchenlintt: Let's do the refactor:
1. add loanFactory to the initializer
2. add setters and call them on existing pools

It's less painful longer term, and we have to upgrade the ImplementationReference anyway to launch this feature.

yuchenlintt (changes requested):

contracts/truefi2/LoanToken2.sol
vanruch: The safer solution would be to remove it and add later if we need to. For now I'll assume that agency holds all loan tokens

contracts/truefi2/TrueFiPool2.sol
vanruch: Let's do this in a following PR as it might be another big change

contracts/truefi2/LoanToken2.sol
```diff
@@ -374,17 +370,25 @@ contract LoanToken2 is ILoanToken2, ERC20 {
         require(!isRepaid(), "LoanToken2: cannot default a repaid loan");
         require(start.add(term).add(creditOracle.gracePeriod()) <= block.timestamp, "LoanToken2: Loan cannot be defaulted yet");
         status = Status.Defaulted;
+        debtToken = loanFactory.createDebtToken(pool, borrower, debt.sub(repaid()));
+
+        uint256 poolBalance = balanceOf(lender);
```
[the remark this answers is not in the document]
vanruch: We should also keep backwards compatibility with old loans interface. Will do in next PRs

yuchenlintt (dismissed):

contracts/truefi2/LoanToken2.sol
```diff
@@ -292,14 +284,18 @@ contract LoanToken2 is ILoanToken2, ERC20 {
     /**
      * @dev Get coupon value of this loan token in token
      * This assumes the loan will be paid back on time, with interest
-     * @param _balance number of LoanTokens to get value for
+     * @param _amount number of LoanTokens to get value for
      * @return coupon value of _balance LoanTokens in tokens
```
MiksuJak, line 288:
```suggestion
     * @return coupon value of _amount LoanTokens in tokens
```

test/truefi2/LoanToken2.test.ts
```diff
@@ -293,7 +305,7 @@ describe('LoanToken2', () => {
       await withdraw(borrower)
       await token.mint(loanToken.address, parseEth(1099))
       await timeTravel(provider, defaultedLoanCloseTime)
-      await expect(loanToken.enterDefault()).to.emit(loanToken, 'Defaulted').withArgs(parseEth(1099))
+      await expect(loanToken.enterDefault()).to.emit(loanToken, 'Defaulted')
```
skalermo, line 308: Can we test arguments of the changed event (now it returns 2 arguments)?

test/truefi2/Liquidator2.test.ts
```diff
@@ -94,8 +94,9 @@ describe('Liquidator2', () => {
       standardTokenOracle: tusdOracle,
       creditModel,
     } = await setupTruefi2(owner, _provider))
-
-    loan = await createLoan(loanFactory, borrower, usdcPool, parseUSDC(1000), YEAR, 1000)
+    const legacyLoanTokenImpl = await new LegacyLoanToken2__factory(owner).deploy()
+    await loanFactory.setLoanTokenImplementation(legacyLoanTokenImpl.address)
+    loan = await createLoan(loanFactory, borrower, usdcPool, parseUSDC(1000), YEAR, 1000) as any
```
skalermo, line 99: Is there a way to avoid using `as any`?

contracts/truefi2/LoanToken2.sol
```diff
@@ -375,16 +371,17 @@ contract LoanToken2 is ILoanToken2, ERC20 {
...
+        uint256 unpaidDebt = debt.sub(repaid());
+        debtToken = loanFactory.createDebtToken(pool, borrower, unpaidDebt);
 
-    /**
-     * @dev Liquidate the loan if it has defaulted
-     */
-    function liquidate() external override onlyDefaulted onlyLiquidator {
-        status = Status.Liquidated;
+        debtToken.approve(address(pool), unpaidDebt);
+        pool.addDebt(debtToken, unpaidDebt);
```
skalermo, line 378: Let's extract `pool` to a local variable.

contracts/truefi2/LoanToken2.sol
```diff
@@ -292,14 +284,18 @@ contract LoanToken2 is ILoanToken2, ERC20 {
...
+     * @return coupon value of _amount LoanTokens in tokens
      */
-    function value(uint256 _balance) external override view returns (uint256) {
-        if (_balance == 0) {
+    function value(uint256 _amount) external override view returns (uint256) {
+        if (_amount == 0) {
             return 0;
         }
 
+        if (status >= Status.Defaulted) {
```
skalermo, line 295: Does `>=` mean that status should be either `Defaulted` or `Liquidated` to enter if statement? If so, consider changing this to explicit comparison with those statuses for better readability.

skalermo, line 295: Consider storing `status` to a local variable, as it is used twice in this function.

contracts/truefi2/LoanToken2.sol
```diff
@@ -168,14 +168,14 @@ contract LoanToken2 is ILoanToken2, ERC20 {
...
         admin = _admin;
-        liquidator = _liquidator;
         amount = _amount;
         term = _term;
         apy = _apy;
         lender = _lender;
         ftlAgency = _ftlAgency;
         creditOracle = _creditOracle;
         debt = interest(amount);
+        loanFactory = _loanFactory;
```
skalermo, line 178: Let's move it to the `liquidator`'s place for consistency.

contracts/truefi2/LoanToken2.sol
```diff
@@ -142,7 +143,6 @@ contract LoanToken2 is ILoanToken2, ERC20 {
      * @param _borrower Borrower address
      * @param _lender Lender address
      * @param _ftlAgency FixedTermLoanAgency address
-     * @param _liquidator Liquidator address
```
skalermo, line 145: Add docstring for `_loanFactory` parameter.

contracts/truefi2/SAFU.sol
```diff
@@ -142,7 +142,8 @@ contract SAFU is ISAFU, UpgradeableClaimable {
      * @param amount Amount of deficiency tokens to be reclaimed
      */
     function reclaim(IDebtToken loan, uint256 amount) external override {
-        require(loanFactory.isLoanToken(loan), "SAFU: Unknown loan");
+        require(loanFactory.isCreatedByFactory(loan), "SAFU: Unknown loan");
```
MiksuJak, line 145: Are we sure if we want this? Later on we call:
`IDeficiencyToken dToken = deficiencyToken[loan];`
and a few lines after that:
`dToken.burnFrom(msg.sender, amount);`

I'm pretty sure it will revert any time we provide a debt token as an argument anyway.

contracts/truefi2/LoanToken2.sol
```diff
@@ -516,7 +513,7 @@ contract LoanToken2 is ILoanToken2, ERC20 {
     }
 
     function version() external override pure returns (uint8) {
-        return 6;
+        return 7;
```
MiksuJak, line 516: The current on chain version is `6`. I don't think we should bump it any higher.

contracts/truefi2/deprecated/LegacyLoanToken2.sol
```diff
@@ -0,0 +1,500 @@
+// SPDX-License-Identifier: MIT
```
MiksuJak, line 1: I think we should copy the on chain version here, instead of the version that we had in the repo once this PR was created.

package.json
```diff
@@ -25,7 +25,7 @@
     "test:true-currencies": "mocha 'test/true-currencies/*.ts'",
     "test:true-gold": "mocha 'test/true-gold/*.ts'",
     "test:truefi": "mocha 'test/truefi/**/*.ts'",
-    "test:truefi2": "mocha 'test/truefi2/**/*.ts' --exclude test/truefi2/lines-of-credit/**/*.ts test/truefi2/complete-flow/**/*.ts",
+    "test:truefi2": "NODE_OPTIONS='--max-old-space-size=5120' mocha 'test/truefi2/**/*.ts' --exclude test/truefi2/lines-of-credit/**/*.ts test/truefi2/complete-flow/**/*.ts",
```
MiksuJak, line 28: 😯

test/truefi2/LoanFactory2.test.ts
```diff
@@ -309,9 +309,9 @@ describe('LoanFactory2', () => {
     })
 
     describe('reverts if', () => {
-      it('caller is not CreditAgency', async () => {
+      it('caller is not CreditAgency or loan', async () => {
         await expect(loanFactory.connect(borrower).createDebtToken(pool.address, borrower.address, parseEth(1)))
-          .to.be.revertedWith('LoanFactory: Caller is not the credit agency')
+          .to.be.revertedWith('LoanFactory: Caller is neither credit agency nor loan')
       })
```
MiksuJak, line 316: Shouldn't we add a test, for a case when a loan (not necessarily a real one) calls `createDebtToken()`?

vanruch: We have a lot of these tests in LoanToken default tests. There's no really any other option to test it so this test would be identical to LoanToken tests where enterDefault is called

contracts/truefi2/deprecated/LegacyLoanToken2.sol
vanruch: It's not that important. The clue is to have working liquidation and these two are same

contracts/truefi2/LoanToken2.sol
vanruch: It's a very major change. Why wouldn't we bump version number here?

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -631,7 +631,10 @@ contract TrueFiPool2 is ITrueFiPool2, IPauseableContract, ERC20, UpgradeableClai
      * @dev CreditAgency transfers DebtToken to the pool
      */
     function addDebt(IDebtToken debtToken, uint256 amount) external override {
-        require(msg.sender == address(creditAgency), "TruePool: Only LineOfCreditAgency can add debtTokens");
+        require(
```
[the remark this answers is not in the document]
vanruch: Later. There's a comment for it already.

test/truefi2/Liquidator2.test.ts
vanruch: We can but this is a one-time case when createLoan returns LegacyLoanToken instead of LoanToken and it's due to changes done outside of the function. Since it's a one-off, I prefered to use as any to show exactly this.

vanruch: loan will still be typed as LegacyLoanToken

contracts/truefi2/LoanToken2.sol
vanruch: Looks like premature optimization. I'm pretty sure EVM optimizes can cache SLOADs from same location

MiksuJak (dismissed):

contracts/truefi2/LoanToken2.sol
vanruch: Ok, extracting pool would save us 190 gas out of 430k. Still not worth it

contracts/truefi2/LoanToken2.sol
vanruch: premature optimization again

MiksuJak (approved):

skalermo (approved):

PR #1009, opened 2021-09-27
Start TrueFiVault vesting at block.timestamp; replace ethers deploy script with Mars; deploy to mainnet

yuchenlintt:
I ran this script to deploy + verify an investor vault to mainnet (although I killed the script before it could initialize the vault since I don't have treasury TRU).


vanruch (dismissed):

vanruch (approved):

deploy/vault.ts
```diff
@@ -0,0 +1,49 @@
...
+  // New contract proxies
+  const trueFiVault = proxy(trueFiVault_impl, () => { })
+
+  // Contract initialization
+  runIf(trueFiVault.isInitialized().not(), () => {
+    // TODO tru.approve(trueFiVault, AMOUNT)
+    trueFiVault.initialize(BENEFICIARY, AMOUNT, tru, stkTruToken)
+
+    trueFiVault.transferOwnership(OWNER_MULTISIG)
+    // TODO trueFiVault.transferProxyOwnership(OWNER_MULTISIG)
```
MiksuJak, line 44: cannot we call it?

MiksuJak (approved):

hal909 (approved):

deploy/vault.ts
yuchenlintt: Unfortunately not. Mars proxies only have the ABI of their impl contract, but they should also have their proxy ABI.

PR #1013, opened 2021-09-28
Remove 'Skip CI' from test names

MaciejNadolski98:


yuchenlintt:
Any idea why the integration test is still flaky? The CurveYearnStrategy integration test should be deterministic* now, right?

*modulo Alchemy server uptime, of course

vanruch:
Looks like the crvBalance is not deterministic for some reason. Maybe it depends on time. And we think that 1Inch fails on slippage check when crvBalance is bigger. Increasing minAmount in data may help

yuchenlintt (approved):
I'm not sure whether this is passing now due to luck or a fix. Ostensibly moving provider initialization to beforeEach shouldn't affect whether integration tests pass.

Regardless I'm fine with checking this in and fixing tests in a follow-up if it becomes a big issue.

skalermo (approved):

PR #1014, opened 2021-09-30
Recursively remove TL2.fund()

skalermo:


contracts/truefi2/TrueLender2.sol
```diff
@@ -53,9 +52,6 @@ contract TrueLender2 is ITrueLender2, UpgradeableClaimable {
 
     mapping(ITrueFiPool2 => ILoanToken2[]) public poolLoans;
 
-    // maximum amount of loans lender can handle at once
-    uint256 public maxLoans;
```
vanruch, line 57: Wouldn't it break storage?

test/integration/TrueLender2.test.ts
```diff
@@ -27,7 +27,7 @@ import { mock1Inch_TL2 } from './data'
 
 use(solidity)
 
-describe('TrueLender2', () => {
+xdescribe('TrueLender2', () => {
```
vanruch, line 30: Why skipping the test? TL2 will be used until the last loan is repaid so we should keep testing it

vanruch, line 30: We need to rewrite this into FTLA tests in following PRs

skalermo: Task filed.

contracts/truefi2/TrueLender2.sol
```diff
@@ -88,27 +88,16 @@ contract TrueLender2 is ITrueLender2, UpgradeableClaimable {
     // minimum prediction market voting period
     uint256 public DEPRECATED__votingPeriod;
 
-    ITrueFiCreditOracle public creditOracle;
+    ITrueFiCreditOracle public DEPRECATED__creditOracle;
```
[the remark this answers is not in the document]
yuchenlintt: You can also do this for `DEPRECATED__ratingAgency` in line 66 above.

contracts/truefi2/TrueLender2.sol
```diff
@@ -399,23 +296,6 @@ contract TrueLender2 is ITrueLender2, UpgradeableClaimable {
         return borrowSum;
     }
 
-    /**
```
yuchenlintt, line 402: The `totalBorrowed()` function in line 381 was only ever called by our new credit model, so we can recursively remove that too. If you prefer, we can do this in a separate PR, although I think it fits here and should be pretty straightforward.

yuchenlintt (dismissed):
I found a couple more state vars and functions to remove, otherwise LGTM!

vanruch (approved):

MiksuJak (approved):

yuchenlintt:
LGTM (post-merge).

Confirming I diff checked. The last storage variable on the [mainnet impl](https://etherscan.io/address/0x1bbbf37a055059089d9438566fc5f100e9f08eb6#code) is `votingPeriod`, so it should be safe to delete new storage variables introduced after that point.

PR #1019, opened 2021-10-01
Recursively remove TRAV2 rating functionality

MiksuJak:

PR #1020, opened 2021-10-01
Remove lender from loan

vanruch:


contracts/truefi2/deprecated/LegacyLoanToken2.sol
```diff
@@ -156,11 +153,14 @@ contract LegacyLoanToken2 is ILoanToken2, ERC20 {
         amount = _amount;
         term = _term;
         apy = _apy;
-        lender = _lender;
         ftlAgency = _ftlAgency;
         debt = interest(amount);
     }
 
+    function setLender(address _lender) external {
```
yuchenlintt, line 160: We should guard this just in case we deploy this again.

```suggestion
    function setLender(address _lender) external onlyAdmin {
```

yuchenlintt (dismissed):
LGTM, but let's wait until #1019 is merged to see if that helps fix `test-truefi2` for this PR.

contracts/truefi2/deprecated/LegacyLoanToken2.sol
vanruch: This contract has some changes to ease testing, like removing onlyLiquidator modifier. We should not deploy it ever.

yuchenlintt: I see, I misinterpreted the intent of `LegacyLoanToken2` then. If this contract is intended only for use in tests, then let's move this contract to the `mocks/` folder and rename it to `TestLoanToken2`, `TestLegacyLoanToken2`, or something similar. (Can be discussed and done in a separate PR.)

I thought the intent of the `deprecated/` subdir was for contracts that we need to keep for compatibility and in case of urgent upgrade. But if that wasn't the intent then my bad for misinterpreting.

yuchenlintt (dismissed):

contracts/truefi2/LoanFactory2.sol
```diff
@@ -37,7 +37,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
     mapping(IDebtToken => bool) public override isLoanToken;
 
     IPoolFactory public poolFactory;
-    address public lender;
+    address DEPRECATED__lender;
```
skalermo, line 40:
```suggestion
    address private DEPRECATED__lender;
```

vanruch: isn't it the same thing?

vanruch: Ok, it's not. They are internal by default

skalermo (approved):

test/truefi2/Liquidator2.test.ts
```diff
@@ -496,7 +496,7 @@ describe('Liquidator2', () => {
 
     it('providing empty list of loans does not slash tru', async () => {
       const balanceBefore = await tru.balanceOf(stkTru.address)
-      liquidator.connect(assurance).liquidate([])
+      await liquidator.connect(assurance).liquidate([])
```
MiksuJak, line 499: good catch

MiksuJak (approved):

PR #1021, opened 2021-10-04
Remove LoanToken2 -> DebtToken inheritance

MaciejNadolski98:


contracts/truefi2/DeficiencyToken.sol
```diff
@@ -20,14 +20,14 @@ contract DeficiencyToken is IDeficiencyToken, ERC20 {
...
+     * @param _debt Defaulted loans address
      * @param _amount Amount of underlying pool token's that are owed to the pool
      */
-    constructor(IDebtToken _loan, uint256 _amount) public {
+    constructor(IDebtToken _debt, uint256 _amount) public {
         ERC20.__ERC20_initialize("TrueFi Deficiency Token", "DEF");
 
-        loan = _loan;
-        _mint(address(_loan.pool()), _amount);
+        loan = _debt;
```
MiksuJak, line 29: I feel a little inconsistency here. Both should be named either debt or loan.

contracts/truefi2/LoanFactory2.sol
```diff
@@ -34,7 +35,7 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
     // ========= IN STORAGE CORRUPTION ===========
 
     // @dev Track Valid LoanTokens
-    mapping(IDebtToken => bool) public override isLoanToken;
+    mapping(ILoanToken2 => bool) public override isLoanToken;
```
yuchenlintt, line 38: This should become ILegacyLoanToken2.

At the bottom we should also add a mapping for ILoanToken3s created by FTLA.

contracts/truefi2/TrueLender2.sol
```diff
@@ -11,7 +11,7 @@ import {UpgradeableClaimable} from "../common/UpgradeableClaimable.sol";
 import {OneInchExchange} from "./libraries/OneInchExchange.sol";
 
 import {ILoanToken2} from "./interface/ILoanToken2.sol";
-import {IDebtToken} from "../truefi2/interface/ILoanToken2.sol";
+import {ILoanToken2} from "../truefi2/interface/ILoanToken2.sol";
```
yuchenlintt, line 14: This should become `ILegacyLoanToken2`

contracts/truefi2/interface/ILoanFactory2.sol
```diff
@@ -21,7 +22,7 @@ interface ILoanFactory2 {
 
     function isCreatedByFactory(IDebtToken) external view returns (bool);
 
-    function isLoanToken(IDebtToken) external view returns (bool);
+    function isLoanToken(ILoanToken2) external view returns (bool);
```
yuchenlintt, line 25: Let's rename this to `isLoanToken(ILoanToken3)` and add a new function `isLegacyLoanToken(ILegacyLoanToken2)`

contracts/truefi2/interface/ILoanToken2.sol
```diff
@@ -5,43 +5,38 @@ import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {ERC20} from "../../common/UpgradeableERC20.sol";
 import {ITrueFiPool2} from "./ITrueFiPool2.sol";
 
-// These two interfaces have to stay in a single file due to a flattener bug that causes IDebtToken to be declared before ILoanToken2
-interface IDebtToken is IERC20 {
+interface ILoanToken2 is IERC20 {
     enum Status {Awaiting, Funded, Withdrawn, Settled, Defaulted, Liquidated}
```
yuchenlintt, line 9: I think we should break backward compatibility entirely and remove all Statuses except `{Withdrawn, Settled, Defaulted}` for ILoanToken3.

yuchenlintt (changes requested):
This might deserve a larger discussion (so hold off on implementing until you get a chance to chat with @barrutko, @vanruch, etc. about whether my suggestion make sense).

I think we should completely break backward compatibility of FTLA loans (LoanToken2 => LoanToken3) with TrueLender2 loans (LoanToken2 => LegacyLoanToken2). Here's why:
1. They work differently. TL2 loans were created by LoanFactory2 and have statuses (Awaiting, Funded, Liquidated) that are now unreachable in FTLA loans. FTLA loans are directly funded during creation and can only transition from Funded/Withdrawn* to terminal states {Settled, Defaulted}.
2. This would let us remove all the typecasting (dangerous!!). We wouldn't need to reason about ABI + behavior differences between legacy TL2 loans and new FTLA loans.
3. This makes deprecation of legacy loan logic safer: all we would need to do is (recursively) remove the functions that mention LegacyLoanToken2. Anything that touches LoanToken3 would be safe to keep.
4. This aligns better with a design philosophy of never modifying old code + only adding new modules. We're not _quite_ there since this kind of change would involve renaming old functions to Legacy/Deprecated, but the fundamental idea is that we don't want to be modifying old code when we introduce new behavior that's meant to replace the old behavior.

\* We should file a follow-up PR that transfers funds directly from the pool to the borrower instead of funding the loan and then separately having the borrower withdraw from the loan.

Feel free to push back and discuss this! I think there can be some very valid counterarguments to my proposal. And if we do decide to move forward with this proposal we may want to split the change into smaller PRs.

yuchenlintt (dismissed):
Offline decision from our standup: the changes in this PR look good as is. Let's merge conflicts and check it in soon!

We do plan to split up TrueLender2 and FTLA's LoanToken2 contracts, but we should do that in a follow-up PR, not here.

Proposed naming for the split will be based on folder structure. There already exists a deprecated LoanToken2 for TrueLender2 (open question: is LT2 version 4 still ABI compatible with TL2?):
TrueLender2: `truefi2/deprecated/LoanToken2Deprecated.sol`
FTLA: `truefi2/LoanToken`

yuchenlintt (dismissed):

contracts/truefi2/LoanFactory2.sol
vanruch: Why should it be legacy loan token? Don't see a reason why we'd want to keep old loans and new ones in different mappings.

vanruch (approved):

contracts/truefi2/LoanFactory2.sol
yuchenlintt: Separating mappings when we break backwards compatibility provides three benefits for us:
1. Type safety within our repo. We wouldn't need to typecast loans into TL2 loans and FTLA loans when we interact with a single mapping that's tracking both. This would help our compiler catch type errors that we might overlook. (Typecasting and untyping addresses are code smells that tells us we're fighting the compiler instead of letting it help us check our work.)
2. Easier on-chain monitoring of a deprecation/migration. It's much easier to check that a mapping or array is empty than to loop through all addresses, check ABIs to see whether they match the old or new version, and reduce to determine whether we're still using the old version anywhere.
3. Segregation of old and new handling logic. Breaking backwards compatibility is just creating new functions while leaving old code untouched. Deprecation is just deleting old code while leaving new code untouched.

Granted, _this_ particular mapping doesn't benefit much from points 1-3, but if we're segregating logic for TL2 loans and FTLA loans elsewhere, then point 4 below applies:
4. Consistency across a repo.

I'd be happy to keep this in a single mapping if we have a compelling case for it. One possibility I see is that `isLoanToken()` gets called in several places that we'd have to split into ugly case logic for TL2 and FTLA. But my hope is that the places where we call `isLoanToken()` vs. `isLegacyLoanToken()` (or whatever we'd like for TL2) are exactly the places we plan to split LT2 and FTLA implementations into separate functions.

contracts/truefi2/LoanFactory2.sol
```diff
@@ -178,8 +179,8 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         return newToken;
     }
 
-    function isCreatedByFactory(IDebtToken loan) external override view returns (bool) {
-        return isLoanToken[loan] || isDebtToken[loan];
+    function isCreatedByFactory(address loanOrDebt) external override view returns (bool) {
```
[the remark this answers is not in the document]
yuchenlintt: Agreed. Hopefully this will get split and typed again very soon!

contracts/truefi2/LoanFactory2.sol
```diff
@@ -178,8 +179,9 @@ contract LoanFactory2 is ILoanFactory2, Initializable {
         return newToken;
     }
 
-    function isCreatedByFactory(IDebtToken loan) external override view returns (bool) {
-        return isLoanToken[loan] || isDebtToken[loan];
+    function isCreatedByFactory(IDebtToken debt) external override view returns (bool) {
```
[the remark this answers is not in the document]
yuchenlintt: This is fine for now since we're about to split this function in a follow-up PR, but for the future I think we should consider typecasting and untyped addresses to be a code smell that indicates we're fighting the compiler. I'd rather duplicate code than leave it untyped to handle two different possible input types.

yuchenlintt (approved):

PR #1023, opened 2021-10-05
📚 Copy CONTRIBUTING, README, SECURITY from deprecated master branch

yuchenlintt:


SECURITY.md
```diff
@@ -0,0 +1,14 @@
+# Security Policy
+
+## Supported Versions
+
+Our smart contracts are under active development. As such, we only support the most current release.
```
vanruch, line 5: Should we also mention that we support code that is currently deployed? Because it's often different to main branch

vanruch (dismissed):

SECURITY.md
yuchenlintt: I clarified the wording. We'll probably want to update this once @barrutko publishes his BBBP spec.

MiksuJak (approved):

PR #1031, opened 2021-10-06
✂️ Skip LoanToken2 Funded and go straight to Withdrawn

yuchenlintt:


contracts/truefi2/LoanToken2.sol
```diff
@@ -173,6 +173,10 @@ contract LoanToken2 is ILoanToken2, ERC20 {
         loanFactory = _loanFactory;
         creditOracle = _creditOracle;
         debt = interest(amount);
+        status = Status.Withdrawn;
+        start = block.timestamp;
+        _mint(address(ftlAgency), debt);
+        emit Withdrawn(borrower);
```
vanruch, line 179: Do we need this event since there was no any withdraw performed inside LT contract?

yuchenlintt: Good call, I'll remove it!

yuchenlintt:
I've got to admit, I'm baffled. Too many hours trying to fix the unit tests for such a simple Solidity change.

I'm marking this PR as a draft WIP for now, unless someone else knows how to fix the tests.

test/truefi2/TrueRatingAgencyV2.test.ts
```diff
@@ -179,7 +183,9 @@ describe('TrueRatingAgencyV2', () => {
   describe('Claim', () => {
     const rewardMultiplier = 1
     beforeEach(async () => {
-      loanToken = await createLegacyLoan(lender, loanFactory, owner, tusdPool, parseEth(5e6), yearInSeconds * 2, 100)
+      const tx = await loanFactory.createLegacyLoanToken(tusdPool.address, owner.address, parseEth(5e6), yearInSeconds * 2, 100)
+      loanToken = await extractLegacyLoanToken(tx, owner)
+      await loanToken.setLender(lender.address)
```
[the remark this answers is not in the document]
skalermo: I agree. This pattern is used many times and it would be good to extract it to separate function.

skalermo (approved):

test/truefi2/TrueRatingAgencyV2.test.ts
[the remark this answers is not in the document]
vanruch: +1 if this doesn't break other tests

vanruch (approved):

vanruch:
What is our plan about potential hacked whitelisted borrowers? I suppose we still need to have some centralised final loan confirmation function

contracts/truefi2/LoanToken2.sol
```diff
@@ -333,16 +327,8 @@ contract LoanToken2 is ILoanToken2, ERC20 {
...
-     * Sets the status of the loan to Withdrawn
-     * @param _beneficiary address to send funds to
-     */
-    function withdraw(address _beneficiary) external override onlyBorrower onlyFunded {
-        status = Status.Withdrawn;
-        token.safeTransfer(_beneficiary, amount);
-
-        emit Withdrawn(_beneficiary);
+    function withdraw(address) external override {
+        revert("LoanToken2: Direct withdrawal has been deprecated");
```
barrutko, line 331: Same as above.

yuchenlintt:
> What is our plan about potential hacked whitelisted borrowers?

Had a talk with Ryan yesterday. The plan is that TrueFi should be more vigilant about KYC. Ryan already recommends borrowers use multisig, hardware wallet, etc. We should just step it up from a recommendation to a requirement.

In case of compromise, our plan is to help a borrower recover in good faith. However the responsibility still lies with the borrower to secure their keys + auth.

> I suppose we still need to have some centralised final loan confirmation function

One-click borrowing is on our public roadmap for this cycle. It's up there with launch of LoC + borrower staking as the three biggest things we're announcing, so we need to deliver on it.

Perhaps we might be biasing too much toward status quo? I thought loan rating was only meant to be a temporary measure when it was introduced. We engineers should be the ones advocating for capital efficiency, removing arbitrary constraints like requiring central approval, only trading during business hours, etc. Borrowing in one transaction is a better product, and it's the crypto-native way.

PR #1034, opened 2021-10-06
⛩️ Use BorrowingMutex to gate lending interactions in TrueFiPool2

skalermo:


contracts/truefi2/PoolFactory.sol
```diff
@@ -171,6 +175,7 @@ contract PoolFactory is IPoolFactory, UpgradeableClaimable {
         ftlAgency = _ftlAgency;
         safu = _safu;
         loanFactory = _loanFactory;
+        borrowingMutex = _borrowingMutex;
```
yuchenlintt, line 178: We'll also need to add a setter + tests for borrowingMutex since this is an existing contract. Let's do that in a follow-up PR.

contracts/truefi2/TrueFiPool2.sol
```diff
@@ -274,21 +279,18 @@ contract TrueFiPool2 is ITrueFiPool2, IPauseableContract, ERC20, UpgradeableClai
     /**
      * @dev only FixedTermLoanAgency, or CreditAgency can perform borrowing or repaying
```
yuchenlintt, line 280: Let's make this comment more robust to future agencies we might add. We should also clarify that this modifier is only used to gatekeep borrowing, not repaying.
```suggestion
     * @dev only agencies allowed by the borrowingMutex (like FixedTermLoanAgency) can borrow
```

yuchenlintt (changes requested):
LGTM after addressing comments!

vanruch (changes requested):
I feel like this change is too unsafe. We use this modifier to guard very critical borrow function. 
What if somewhat in the future somebody will decide to allow admin account to lock mutex for some reason and will forget that isAllowedToLock is used in the pool (because it's very much not obvious).

contracts/truefi2/PoolFactory.sol
MiksuJak: I actually think we should add it in this PR

PR #1040, opened 2021-10-08
👽 Add Mars logging to dry run deployments

MaciejNadolski98:


marsDeploy.sh
```diff
@@ -46,6 +46,8 @@ if [[ "${dry_run}" == 'false' ]]; then
         echo "Error: Build canary does not match current commit hash. Please run yarn build."
         exit 1
     fi
+else
+    args="${args} --log deploy.log"
```
vanruch, line 50: I don't think we want to always write to log. Maybe let's confine the PR to bumping mars version only?

yuchenlintt: I don't think it hurts to turn on logging by default. This is a way for us to audit what we deployed, in case we forget how far the script ran or what local changes were captured.

Happy to discuss, but my recommendations are:
- turn on logging for all deploy runs
- name logs according to network, script, dry run status, and unix timestamp in seconds, so old logs aren't immediately overwritten. For instance, `deploy-rinkeby-truefi2-dry-run-1666666666.log` or `deploy-mainnet-strategy-1666666666.log`
- move log output to the `cache/` dir so it doesn't pollute the root dir

Your thoughts?

yuchenlintt (changes requested):
LGTM, but please revert `deployments-ropstenn.json` naming back to `deployments-ropsten.json`

vanruch (approved):

PR #1050, opened 2021-10-11
🔪 Add slashing function to CollateralVault

MiksuJak:
This PR needs to be followed by a few other changes:
1. Making `LoanToken` ban borrower in `BorrowingMutex` when `enterDefault()` is called
2. Testing `CollateralVault` integration with `Liquidator`
3. Testing `CollateralVault` integration with `SAFU`

As I tried to implement it all at once, I noticed that 2 and 3 would most likely be dependent on 1, but the changes would be too big for a single PR.


contracts/truefi2/CollateralVault.sol
```diff
@@ -63,19 +65,24 @@ contract CollateralVault is ICollateralVault, UpgradeableClaimable {
...
         emit Unstaked(msg.sender, amount);
     }
 
     function slash(address borrower) external override {
-        // require(msg.sender == liquidator)
-        // require(borrowingMutex.isBanned(borrower))
-        // transfer stakedAmount() to liquidator
-        require(borrower == address(stakedToken)); // silence lint and build warnings
-        revert("Unimplemented!");
+        require(msg.sender == address(liquidator), "CollateralVault: Caller is not the liquidator");
```
vanruch, line 76: Should we keep liquidator as a plain address?

vanruch: Can't think of a scenario where this type would carry any value besides being self-documentation. But maybe it's a good enough reason to keep it to be fair

yuchenlintt (approved):
Mostly LGTM! Some nitpick changes, but if this PR blocks other tasks then we can file those as follow-up fixes.

skalermo (approved):

PR #1052, opened 2021-10-11
🚫 Ban borrower after LoanToken2 default

MaciejNadolski98:


MiksuJak (approved):

contracts/truefi2/BorrowingMutex.sol
```diff
@@ -27,11 +27,18 @@ contract BorrowingMutex is IBorrowingMutex, UpgradeableClaimable {
 
     event BorrowerUnlocked(address borrower, address locker);
 
+    event BorrowerBanned(address borrower);
```
yuchenlintt, line 30: :+1: Thanks for catching my oversight and adding this event!

contracts/truefi2/BorrowingMutex.sol
```diff
@@ -41,23 +48,19 @@ contract BorrowingMutex is IBorrowingMutex, UpgradeableClaimable {
         emit LockerAllowed(_locker, isAllowed);
     }
 
-    function ban(address borrower) external override onlyAllowedToLock {
-        _lock(borrower, BANNED);
+    function ban(address borrower) external override onlyLocker(borrower) {
```
yuchenlintt, line 51: Tracking the different access permissions for `ban()`/`lock()`/`unlock()` is starting to get rather messy:
- only a borrower's locker can ban a borrower, but if a borrower is unlocked then they can't be banned
- only FTLA and LOCA can lock a borrower... but by our design FTLA can't unlock borrowers.
- only FTLA _loans_ or LOCA can unlock a borrower.
- banning and unlocking are nearly identical operations (setting the locker address to a sentinel value). But they have different access permissions.

We should move forward with this change because it makes sense in our current lending exclusion model, but if we find an opportunity to simplify then we should take it. (Follow-up task filed.)

PR #1053, opened 2021-10-11
🧮 Add borrower score equations to CreditModel

skalermo:


contracts/truefi2/CreditModel.sol
```diff
@@ -62,6 +62,7 @@ contract CreditModel is ICreditModel, UpgradeableClaimable {
         // adjust for pool value (basis precision)
         uint16 poolValueLimitCoefficient;
         uint16 ltvRatio;
+        uint16 effectiveScorePower;
```
yuchenlintt, line 65: Now that I reflect on this some more, I've realized that `ltvRatio` and `effectiveScorePower` don't belong in the `BorrowLimitConfig`.
- `ltvRatio` will affect _both_ borrow limits and interest rates, so it doesn't make sense to update it with `BorrowLimitConfig` when we just want to alter how conservative we're being with borrower staking.
- `effectiveScorePower` _only_ affects interest rates, not borrow limits, so it definitely doesn't belong in `BorrowLimitConfig`.

These should be in a separate config (maybe named `BorrowerStakingConfig`?), since we'll most likely modify these values simultaneously.

yuchenlintt (changes requested):
The equations LGTM, but let's move borrower staking config params into their own separate struct that's independent of `BorrowLimitConfig`.

yuchenlintt (dismissed):

test/truefi2/lines-of-credit/CreditModel.test.ts
```diff
@@ -363,10 +384,59 @@ describe('CreditModel', () => {
...
+        [200, 50],
+        [100, 100],
+        [75, 100],
+        [0, 0],
+      ].map(([borrowed, result]) =>
+        it(`when borrowed amount is ${borrowed} collateral ratio is at ${result}%`, async () => {
+          expect(await creditModel.conservativeCollateralRatio(mockPool.address, parseTRU(collateral), parseEth(borrowed)))
+            .to.equal(result * 100)
+        }))
+    })
```
vanruch, line 413: Miss test of what happens when collateral is 0

test/truefi2/lines-of-credit/CreditModel.test.ts
```diff
@@ -363,10 +384,59 @@ describe('CreditModel', () => {
...
+        [223, 228],
+        [191, 201],
+        [159, 174],
+        [127, 147],
+        [95, 120],
+        [63, 93],
+        [31, 66],
+      ].map(([score, effectiveScore]) =>
+        it(`staking ${collateral} TRU increases
+        score ${score} to ${effectiveScore}`, async () => {
```
vanruch, line 436: was new line intended?

skalermo: It was, but now I see there is no need.

test/truefi2/lines-of-credit/CreditModel.test.ts
```diff
@@ -363,10 +384,59 @@ describe('CreditModel', () => {
...
+          expect(await creditModel.effectiveScore(score, mockPool.address, parseTRU(collateral), parseEth(borrowedAmount))).to.eq(effectiveScore)
+        }))
+    })
```
vanruch, line 439: Missing tests with collateral=0, with pool with different decimal count, with collateral below borrowed amount and with different score power

contracts/truefi2/CreditModel.sol
```diff
@@ -61,7 +61,11 @@ contract CreditModel is ICreditModel, UpgradeableClaimable {
         uint16 tvlLimitCoefficient;
         // adjust for pool value (basis precision)
         uint16 poolValueLimitCoefficient;
+    }
+
+    struct StakingConfig {
         uint16 ltvRatio;
+        uint16 effectiveScorePower;
```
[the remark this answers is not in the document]
yuchenlintt: Nope! Keep it as simple as possible.

The original equations spec was clear that this should be uint. I'm not sure why it got dropped in the final spec.

yuchenlintt:
> Please add more tests. These are almost pure functions so testing should be easy and we want this part of code to be as tested as it could

Agreed on more tests. But let's check this in and follow up in a separate PR. There are important Solidity tasks blocked on this API at the moment.

vanruch (dismissed):

PR #1076, opened 2021-10-19
🪚 Remove unused DebtToken's statuses

skalermo:


yuchenlintt:
> We can simplify this even further by replacing enum `Status` with bool `liquidated`, because by default DebtToken already has status `Defaulted`.

Sounds good to me. Let's call this `bool hasLiquidated()`? Boolean functions read better when written as `isX()`, `hasX()`, etc.

contracts/truefi2/DebtToken.sol
```diff
@@ -75,10 +72,10 @@ contract DebtToken is IDebtToken, ERC20 {
      * @dev Liquidate the debt if it has defaulted
      */
     function liquidate() external override {
-        require(status == Status.Defaulted, "DebtToken: Current status should be Defaulted");
+        require(!hasLiquidated, "DebtToken: Debt must not be liquidated");
```
vanruch, line 75: I'd prefer `isLiquidated`

MiksuJak: +1, that's how we do it in our code everywhere

MiksuJak (approved):

PR #1082, opened 2021-10-21
🦦 Deprecate true lender

MaciejNadolski98:


contracts/truefi2/TrueFiPool2.sol
```diff
@@ -353,6 +359,11 @@ contract TrueFiPool2 is ITrueFiPool2, IPauseableContract, ERC20, UpgradeableClai
         emit LoanFactoryChanged(_loanFactory);
     }
 
+    function setLender(ITrueLender2Deprecated _lender) external onlyOwner {
```
MiksuJak, line 362: I honestly think, that this function would make this contract TestTrueFiPool2. I don't believe we would ever deploy that version.

vanruch (approved):

PR #1096, opened 2021-10-28
💧 Pay liquidity farming rewards in stkTRU

MiksuJak:
This unfortunately increases `claim` cost from `141401` to `340174`.


vanruch (approved):

vanruch:
I'm not a fan of this idea, this will make claiming rewards rather expensive

skalermo (approved):

PR #1101, opened 2021-10-30
✅ Add minimal BorrowingMutex formal verification

MaciejNadolski98:


.gitignore
```diff
@@ -16,3 +16,7 @@ coverageEnv
 .idea
 .DS_Store
 /scripts/.ropsten-test.env
+
+# Generated by Certora
+.certora_*
+.last_confs
```
MaciejNadolski98, line 22: Certora doesn't seem to provide a way to create those files elsewhere out-of-the-box, so they will be ignored for now

spec/truefi2/BorrowingMutex.spec
```diff
@@ -0,0 +1,36 @@
+methods {
+    isUnlocked(address) returns bool envfree
+    isBanned(address) returns bool envfree
+}
+
+definition isReasonableAddress(address alice) returns bool = alice != 0 && alice != 1;
+definition isReasonableEnv(env e) returns bool = isReasonableAddress(e.msg.sender);
+
+rule onePlusTwoEqualsThree(uint one, uint two) {
```
MaciejNadolski98, line 9: This is left from previous commit for sake of an example, will be removed later.

spec/truefi2/BorrowingMutex.spec
```diff
@@ -0,0 +1,36 @@
...
+    require one == 1 && two == 2;
+
+    uint three = one + two;
+
+    assert three == 3, "One plus two does not equal three";
+}
+
+rule unlockedIntoBanned() {
```
yuchenlintt, line 17: Correct me if I'm mistaken, but I believe this rule currently says (in pseudocode)
```
(in a reasonable environment with an unlocked borrower)
if anyone calls lock:
  (with a reasonable locker)
  then borrower is not banned
else:
  any other successful function call won't ban borrower either
```

If so, then let's:
1. simplify this rule so it only verifies a single simple proof. We could use a `require(f.selector == ...)` instead of branching with an if block, so this rule would only be concerned with function calls to `lock()`.
2. give the rule a more descriptive name, like `reasonableLocksDoNotBanUnlockedBorrowers()`. A Certora convention appears to be writing rules as [declarative statements](https://prover.certora.com/output/92911/ad05c881e7b59c4af535/?anonymousKey=a03aa9ba4d959b530c5cbd427350385bf8a5e22a#envfreeFuncsAreNonpayableResults). I'm not sure whether we should make the subject of the sentence singular or plural though. (In a typical unit test, the subject would of course be singular because we're only traveling one path through code, but here we're proving _all_ subjects follow a rule.)
3. separately create other simple rules around banning, like `onlyBansOrLocksCanBanBorrower`, or `bannedBorrowersCannotBeReasonablyUnbanned()` (might be _too_ verbose? I'd be okay with dropping `Reasonably`). We should do this in separate PRs, since the purpose of this PR is to check in a proof of concept and establish coding conventions.

yuchenlintt (changes requested):
Hooray for Certora access!

Until we hook this up to CircleCI, let's establish a coding convention of adding a PR comment that links to the most recent verification report (for modified spec files), after every commit that's ready for review. I haven't had a chance to run this spec myself yet, but when I do I'll attach the verification report.

And don't forget to track tasks on our board :) I suppose I should be _adding_ tasks to our board as well.

spec/truefi2/BorrowingMutex.spec
MaciejNadolski98: The interpretation is correct, although the goal of this rule is to make sure that it's impossible to go from unlocked to banned directly, no matter what function was called exactly - it's more of a rule about possible transitions between states rather than a test of how a function affects the states
1. The `if` statement specialises locker address assumption for `lock()` call. I think this is a good presentation of how that can be done without sacrificing generality and would leave it in as is.
2. Takeaways:
- use declarative statements as rule names
- drop "reasonable". (unless we wanted to differentiate between function with open access and onlyOwner access)
- i would stick to singular, my reasoning is "(any reasonable call to a)functionDoesNotBanUnlockedBorrower"
- additionally I would consider adding comments pointing out non-trivial assumptions (it's obvious that the `msg.sender` is not going to be address(1), but a `locker` function argument could be). Certora translates their rules to Hoare triples in Audit reports, we could mimic that in some form of documentation. (added as task)
3. Agree, add more rules in a following PR.

spec/truefi2/BorrowingMutex.spec
```diff
@@ -0,0 +1,35 @@
+methods {
```
MaciejNadolski98, line 1: Verification report:
https://prover.certora.com/output/67403/1e4853bceba0c7cd83c7/?anonymousKey=7a9f73737871b861a55ebcabad823860de344431

spec/truefi2/BorrowingMutex.spec
```diff
@@ -0,0 +1,36 @@
...
+rule unlockedIntoBanned() {
+    address borrower;
+    require isReasonableAddress(borrower);
+    require isUnlocked(borrower);
+
+    method f;
+    env e;
+    require isReasonableEnv(e);
+    calldataarg args;
+    if (f.selector == lock(address, address).selector) {
```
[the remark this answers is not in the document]
yuchenlintt: Declined. Let's keep as is.

Without currying it's hard to express `require(!(f.selector == foo().selector && args[2] == bar))` constraints, and it ends up being less verbose to use if/else blocks for these.

Another option is to fix `BorrowingMutex` so it reverts when locking to `address(1)` (my preference). Can and should be done in a follow-up PR.

yuchenlintt (approved):

MiksuJak (approved):

PR #1102, opened 2021-11-02
🎃 Rename LoanToken2 to FixedTermLoan

vanruch:


contracts/truefi2/FixedTermLoanAgency.sol
```diff
@@ -353,7 +353,7 @@ contract FixedTermLoanAgency is IFixedTermLoanAgency, UpgradeableClaimable {
         uint256 apy = rate(pool, borrower, amount, term);
         require(apy <= _maxApy, "FixedTermLoanAgency: Calculated apy is higher than max apy");
 
-        ILoanToken2 loanToken = loanFactory.createLoanToken(pool, borrower, amount, term, apy);
+        IFixedTermLoan loanToken = loanFactory.createLoanToken(pool, borrower, amount, term, apy);
```
skalermo, line 356: Are we going to change function names like here, and also event and variable names which have "loanToken" in them?

vanruch: I'm against changing event, we can consider changing function name. I'd stick with current one as it pairs with createDebtToken well, but I see arguments for change too

MiksuJak: Why not rename the events? This contract hasn't even gone live yet

MiksuJak (approved):

contracts/truefi2/FixedTermLoanAgency.sol
vanruch: I think Roman meant LoanFactory event

skalermo (approved):

PR #1103, opened 2021-11-02
🚜 Add minimal RateModel formal verification

MaciejNadolski98:


spec/truefi2/BorrowingMutex.spec
```diff
@@ -0,0 +1,35 @@
+methods {
```
MaciejNadolski98, line 1: This was merged from BorrowingMutex spec PR

spec/truefi2/RateModel.spec
```diff
@@ -0,0 +1,20 @@
+methods {
```
MaciejNadolski98, line 1: Verification Report:
https://prover.certora.com/output/67403/b1f5a7ae2ab1ccee40a3/?anonymousKey=9c8c3d50b42386ae644d65c6e8ac0c4bcad14e14

vanruch (dismissed):

spec/truefi2/RateModel.spec
```diff
@@ -0,0 +1,20 @@
...
+  poolBasicRate(address, uint256) returns uint256 envfree
+
+  liquidRatio(uint256) returns uint256 => DISPATCHER(true)
```
yuchenlintt, line 4: Nota bene for other reviewers (and myself).

I found documentation for `DISPATCHER(true)` here:
https://certora.atlassian.net/wiki/spaces/CPD/pages/41058462/Function+Summarization#Calls-inside-the-code

My summary of what it does (and an opportunity for @MaciejNadolski98 to correct me if I'm mistaken):
- `DISPATCHER` is the same as `DISPATCHER(false)`. It would typically havoc a call to _any function_ in _any other contract_ with _any state_, whether or not such a contract exists, as long as it matches this signature. `DISPATCHER` can't inject havoced functions into the _current_ contract being tested, and must use any local implementations that match this signature.
- `DISPATCHER(true)` works differently. It only checks against functions that match the signature of a target contract (e.g., a contract in our repo that we've provided to the certoraRun script).

So essentially this line is saying we assume our contract can call a `liquidRatio(uint256)` function. Regardless of the _state_ of the contracts implementing `liquidRatio(uint256)`, this function must behave like it came from our provided list of contracts.

spec/truefi2/RateModel.spec
```diff
@@ -0,0 +1,20 @@
+methods {
+  poolBasicRate(address, uint256) returns uint256 envfree
```
yuchenlintt, line 2: Nota bene for other reviewers (and myself when I forget :)).

`envfree` documentation: https://certora.atlassian.net/wiki/spaces/CPD/pages/181960777/Method+Declarations

Essentially, this function doesn't access `msg.sender`, `tx.origin`, `block.timestamp`, etc.
https://certora.atlassian.net/wiki/spaces/CPD/pages/7340101/Types

spec/truefi2/RateModel.spec
```diff
@@ -0,0 +1,20 @@
...
+
+  liquidRatio(uint256) returns uint256 => DISPATCHER(true)
+  getWeeklyAPY() returns uint256 => DISPATCHER(true)
+  balanceOf() returns uint256 => DISPATCHER(true)
+}
+
+rule poolBasicRateIsMonotoneWrtAmount {
+  uint256 amount1;
+  uint256 amount2;
+  address pool;
```
yuchenlintt, line 12: style nitpick: Let's move declarations as close as possible to where they're used. So `pool` should be moved to line 16, above rate1's declaration.

It reduces the amount of working memory we need to review, and makes it clear that pool is irrelevant before the line where it's used.

Aside: I think only C still requires all variables must be declared at the start, before any of them are used.

yuchenlintt: If you have a stack of other branches based on this one, feel free to do this change in a follow-up PR after the stack is merged. I'll approve this PR as-is.

yuchenlintt (approved):
Looks great!

I'm using these PR reviews to refresh my own understanding of Certora syntax. There's a lot of subtle info to internalize!

spec/truefi2/RateModel.spec
MaciejNadolski98: Another way to think about it:
- `DISPATCHER(true)` (aka. DISPATCHER in "optimistic mode") tries to call all functions in all known (provided as certoraRun arguments) contracts with the same signature.
- `AUTO` havocs the call, as well as contracts' states depending on specific EVM opcode used: `STATICCALL`s don't affect any contract state, `CALL`s and `CREATE`s havoc all contracts states except the current one and `DELEGATECALL`s and `CALLCODE`s havoc only the current contract
- `DISPATCHER(false)` (aka. DISPATCHER in "pessimistic mode") combines the two above: first run the `DISPATCHER(true)` and if no violations are found, broaden the search space to an `AUTO` havoc.
From the three it seems obvious when we want to use the first two: `DISPATCHER(true)` for cases of interactions between our own contracts and `AUTO` (which is default, as name suggests) for cases of interaction with an untrusted contract. The third one seems strange, but my understanding of a use case is if we have our own contracts that can be called, but also allow community-defined contracts and want to make it explicit in the specification.

vanruch (approved):

PR #1106, opened 2021-11-03
🔨Refactor _rebucket

vanruch:


MiksuJak (approved):

contracts/truefi2/LineOfCreditAgency.sol
```diff
@@ -694,6 +689,11 @@ contract LineOfCreditAgency is UpgradeableClaimable, ILineOfCreditAgency {
         uint8 bucketNumber,
         address borrower
     ) internal returns (uint256 totalBorrowerInterest) {
```
skalermo, line 691: Can you please make the function return explicitly?

vanruch: Good catch, but I actually like implicit return here, works as a self documentation as it's very much not clear what is returned by the function

skalermo: I just noticed newly added `return` and wanted to find second one ... aaand failed. That's how I got to it. +1 for self documentation.

skalermo (approved):

MiksuJak (approved):

skalermo (approved):

MiksuJak (approved):

PR #1108, opened 2021-11-04
✉️ Add anyReasonableAddress()

MaciejNadolski98:


spec/truefi2/BorrowingMutex.spec
```diff
@@ -23,9 +30,8 @@ rule functionDoesNotBanUnlockedBorrower() {
     require isReasonableEnv(e);
```
MaciejNadolski98, line 30: This stays as is, as returning structs in functions is not yet supported by Certora.

spec/truefi2/BorrowingMutex.spec
```diff
@@ -6,6 +6,13 @@ methods {
 definition isReasonableAddress(address alice) returns bool = alice != 0 && alice != 1;
```
MaciejNadolski98, line 6: https://prover.certora.com/output/67403/e3982ef23f3c78f6850c?anonymousKey=fbc7c8d8088c64ca61e6b4ca70427d471381f601

spec/truefi2/BorrowingMutex.spec
```diff
@@ -6,6 +6,13 @@ methods {
...
 definition isReasonableEnv(env e) returns bool = isReasonableAddress(e.msg.sender);
 
+function anyReasonableAddress() returns address {
+    address alice;
+    havoc alice;
```
yuchenlintt, line 11: Do we need this `havoc alice;` line?

If my understanding is correct, the declaration `address alice;` already quantifies over all possible `alice`s. `havoc` is only used to forget existing constraints on a variable's values:
https://certora.atlassian.net/wiki/spaces/CPD/pages/41156829/Hooks#Havoc-Assuming

yuchenlintt (changes requested):

spec/truefi2/BorrowingMutex.spec
yuchenlintt: It's still unclear to me whether multiple invocations of `anyReasonableAddress()` will return the same `alice` or a newly quantified one each time. Perhaps that's where `havoc` might be relevant?

MaciejNadolski98: Yep, that's exactly what this `havoc` is for. Tested by running a rule:
```
rule deterministic {
    assert anyReasonableAddress() == anyReasonableAddress();
}
```
Without `havoc` this rule was not violated.

skalermo: Just curious: would this `deterministic` rule also be violated if one of the addresses were not havoced?

yuchenlintt: In my understanding that should be the case! You wouldn't even need havoc. Any two different declarations would violate `alice == bob`.

MaciejNadolski98: @skalermo I tried running an example to answer your question by implementing a following function:
```
function anyBool(bool havocIt) returns bool {
    bool ret;
    if (havocIt) {
        havoc ret assuming ret@new != ret@old;
        return ret;
    } else {
        return ret;
    }
}
```
This, however, fails with an error message:
```
Syntax error in spec file (BorrowingMutex.spec:11:5): Variable ret is defined on the then branch and used on the else branch - this is disallowed
```
my understanding of this is that `havoc`ing is disallowed in multiple execution branch scenario. I'm still trying to figure out what are the exact restrictions on the keyword.

PR #1111, opened 2021-11-04
🎶 Introduce IDebtInstrument

hal909:
I'm creating this pull request to get feedback on the DebtInstrument interface. Please review the functions and documentation and leave feedback. 

The purpose of IDebtInstrument is to represent debt universally in the TrueFi protocol. We can think of it as a standard we are setting within crypto to represent different types of debt.

Thanks to @yuchenlintt and @barrutko for helping to work on this interface!


vanruch:
What would be the point of having it as an NFT?

yuchenlintt:
> What would be the point of having it as an NFT?

Good question!

From my perspective, the aim is to improve the composability, encapsulation, and stability of our lending ecosystem. A DebtInstrument would only record a debt between a single lender (the owner of the NFT) and a single borrower. There's no mention of tradable shares because in its most stripped-down form we don't necessarily need tradability of all debts. There's no mention of default handling because, again, different lending protocols would want to handle these differently. (A simple protocol could allow instant withdrawals of your own share of repaid debt, while another could split lenders into tranches to determine repayment order. None of that needs to touch the underlying DebtInstrument implementation.)

All the complicated logic specific to particular lending schemes can be pushed up into higher-level synthetic contracts that own DebtInstruments. We would be allowed to sell a DebtInstrument directly to another lender, or transfer it to an ERC20 derivative contract which handles splitting into tradeable shares. The underlying DebtInstrument would hopefully be audited once and then fixed as a module that won't be changed any more.

```
       / borrower1
lender - borrower2
       \ borrower3

lender1 \\
lender2 == derivative - borrower
lender3 //
 
lender1 \\            / borrower1
lender2 == derivative - borrower2
lender3 //            \ borrower3
```
In the above sketch, all the single lines represent ownership of NFT DebtInstruments, and all the double lines represent ERC20 share ownership of a derivative.

contracts/truefi3/interface/IDebtInstrument.sol
```diff
@@ -0,0 +1,57 @@
...
+    function borrow(uint256 amount) external;
+
+    /// @dev Borrower calls this function to repay `amount` of tokens
+    function repay(uint256 amount) external;
+
+    /// @dev Owner funds a loan for `amount` and increases principal
+    function fund(uint256 amount) external;
+
+    /// @dev Owner calls this function to withdraw `amount` of tokens
+    function withdraw(uint256 amount) external;
```
yuchenlintt, line 49: Does `withdraw()` reduce `principal` by amount? Otherwise we'll also want to track and publish `withdrawn()` or `repaid()` views so that repayment state doesn't get lost.

This is similar to `tokenRedeemed` in our current FTLs.

contracts/truefi3/interface/IDebtInstrument.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    /// @dev Principal debt amount
+    function principal() external view returns (uint256);
+
+    /// @dev Each debt has an expiration timestamp
+    function expiry() external view returns (uint256);
+
+    /// @dev Borrower calls this function to borrow `amount` of tokens
+    function borrow(uint256 amount) external;
+
```
yuchenlintt, line 41: Is there a use case where we would want to fund a DebtInstrument but not transfer the principal to the borrower?

If not, then I would prefer to delete this function and just have `fund()` transfer directly from owner to borrower while recording the amount in `principal`. Saves an extra transaction and an extra ERC20 transfer.

contracts/truefi3/interface/IDebtInstrument.sol
```diff
@@ -0,0 +1,57 @@
...
+ *  - The borrower can withdraw funds from the loan
+ *  - The borrower calls repay() to pay back the loan
+ *  - Repayments are withdrawable by the owner
+ *  - Every DebtInstrument has an underlying token
+ *  - Only principal debt is tracked
+ *
+ **/
+interface IDebtInstrument is IERC721 {
+    /// @dev The owner address lends tokens and withdraws repayments
+    function owner() external view returns (address);
```
yuchenlintt, line 25: Is `owner` distinct from the NFT owner? That should be a call to the `ownerOf(tokenId)` view from the IERC721 spec.

yuchenlintt: Actually, shouldn't borrower(), token(), principal(), expiry(), etc. all require a `tokenId` arg? Otherwise you would force every NFT contract to be a singleton containing exactly 1 debt, when instead it should be a single NFT contract that can mint multiple NFT IDs (each representing an individual debt).

contracts/truefi3/interface/IDebtInstrument.sol
yuchenlintt: If `fund()` is restricted to only be callable by the NFT's owner (probably what we currently call agencies?), then we can put a `borrow()` function in the owner contract that delegates to `DI.fund()`.

yuchenlintt:
Since we're planning this to be a complete break from the current TrueFi ecosystem, I think it's also a good time to revisit some fundamental design choices.

For instance, we can significantly improve type safety + encapsulation of untyped uint256s by using a Newtype design pattern (see an [explanation from Rust](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html)) with library code.

Here's a concrete example using timestamps and durations:

```
struct Time {
    uint256 stamp;
}

struct Duration {
    uint256 seconds;
}
```

```
library SafeTimestamp {
    function add(Time, Duration) internal pure returns (Time);
    function sub(Time, Duration) internal pure returns (Time);
    function duration(Time, Time) internal pure returns (Duration);
}

library SafeDuration {
    function add(Duration, Duration) internal pure returns (Duration);
    function sub(Duration, Duration) internal pure returns (Duration);
    function mul(Duration, uint256) internal pure returns (Duration);
}
```

A struct wrapping a single uint256 should be syntactic sugar for the underlying uint256 in both storage and memory. Libraries with only internal functions get deployed once and referenced in multiple contracts (without costing gas in new deployments). So, if I'm not mistaken, this should provide a zero-cost abstraction that enforces correct type usage at compile-time:
```
Time time1 = Time(block.timestamp);
Duration duration = Duration(10);
Time time2 = time1.add(duration);
uint256 foo = time2; // compile error
address foo = time2; // compile error
duration = time2; // compile error
time1 = time1.add(time2); // compile error
time1 = time * time2; // compile error
```

And if we _really_ need to access the underlying info and manipulate it directly or pass it to external contracts, we can always do so explicitly:
```
Time time1 = Time(block.timestamp);
Duration duration = time1.duration(time1);
uint256 rawTimestamp = time.stamp;
uint256 rawSeconds = duration.seconds;
```

This Newtype design pattern can of course be extended to `BasisPoints`, `TokenAmount`, `Decimals`, etc. Any time we perform stoichiometry with uint256s, it would be a good idea to wrap the calculations in a Newtype struct + library to protect ourselves from accidental misuse.

Also, external contracts that don't respect our Newtypes would just use the underlying uint256 when they call functions that take or return our Newtypes. It's all compatible at a bytecode level, just errors out when compiling.

@hal909 @barrutko @vanruch @MiksuJak @skalermo @MaciejNadolski98 Thoughts on this proposal? I'm mentioning this now because we could start implementing this in IDebtInstrument as the return type of `expiry()`.

contracts/truefi3/interface/IDebtInstrument.sol
hal909: Yes, I think there is a case for using fund(). I think funds should be stored in the DebtInstrument until they are borrowed since it allows flexibility in the design on the debtInstrument (ongoing funding/borrowing).

contracts/truefi3/interface/IDebtInstrument.sol
hal909: Good question. I think we have to track the amount repaid. I'm trying *not* to have any functions regarding interest involved here.

vanruch:
@yuchenlintt This is a good idea, I was thinking about something similar. Solidity now has type aliases support https://docs.soliditylang.org/en/v0.8.9/types.html#user-defined-value-types but using structs looks cleaner to me

contracts/truefi3/interface/IDebtInstrument.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.6.10;
```
vanruch, line 2: Let's use solidity >0.8 in truefi 3

contracts/truefi3/interface/IDebtInstrument.sol
```diff
@@ -0,0 +1,57 @@
...
+ * This interface is meant to be extended to support different debt types
+ *
+ * An NFT represents some debt where:
+ *  - Each debt has an owner and a borrower
+ *  - The owner mints and funds a loan
+ *  - The borrower can withdraw funds from the loan
+ *  - The borrower calls repay() to pay back the loan
+ *  - Repayments are withdrawable by the owner
+ *  - Every DebtInstrument has an underlying token
+ *  - Only principal debt is tracked
```
vanruch, line 20: Using withdraw word for both borrower and owner is confusing. Should we replace first one with `borrow`?

contracts/truefi3/interface/IDebtInstrumentWithSchedule.sol
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @title Debt Instrument With Schedule
+ * @notice TrueFi Debt Instrument with Payment Schedule
+ * @dev This Debt Instrument stores an expected schedule of payment
+ * Calling the repay() function should store time and amount of payments
+ * Minting the token should pass in a schedule which is an array
+ * of information about expected payment times and amounts
+ **/
+interface IDebtInstrumentWithSchedule is IDebtInstrument {
```
barrutko, line 15: We could call it `IWithSchedule`.

contracts/truefi3/interface/IDebtInstrumentWithSchedule.sol
```diff
@@ -0,0 +1,29 @@
...
+    // store payment and schedule data
+    struct Payment {
+        // time of payment
+        uint256 timestamp;
+        // payment amount
+        uint256 amount;
+    }
+
+    /// @dev Get expected schedule of payments
+    function schedule() external view returns (Payment[] memory);
```
tt-marek, line 25: A few things here:
1) `tokenId` arg seems to be missing
2) we should not return an array, rather have sth like `nextPayment()`, as:
- you can't iterate it on the blockchain
- whatever calculations it does can be done on front-end

contracts/truefi3/interface/IDebtInstrument.sol
hal909: which version should we use specifically?

contracts/truefi3/interface/IDebtInstrument.sol
hal909: Yes, you are correct. the NFT contract would manage the funds, and each token has its own ID which tracks amounts debt and claimable balance for the borrower/lender

contracts/truefi3/interface/IDebtInstrumentWithSchedule.sol
hal909: We want to be able to get the full schedule to know if a debt is defaulted or not.

This is inherently not scalable past a certain number of payments and schedules. A better way might be to pass in a place in the array we want to get data for instead of returning the whole array. Thoughts?

hal909:
Updates:

- Added tokenId parameter to functions in the DebtInstrument interface
- Removed owner() function since ERC721 uses ownerOf(tokenId)
- Changed IDebtInstrumentWithSchedule to IDebtWithSchedule
- Added tokenID to IDebtWithSchedule

Please leave feedback!

contracts/truefi3/interface/IDebtInstrumentWithSchedule.sol
yuchenlintt: Passing in an array access index could be a viable compromise, although that still forces us to use arrays in the backing storage. Seemingly trivial operations like accessing the total amount currently due would be stuck behind linear iteration, and finding the first date that satisfies X property would be a logarithmic binary search. Modifying schedules (if we ever decide we need to) would be a linear expense as well.

Our largest concern here is unbounded array access. OpenZeppelin has [written a doc](https://defender.openzeppelin.com/#/advisor/docs/access-arrays-using-enumeration-and-pagination) that rates this as a Critical severity issue with a Medium effort solution. Can we can get the desired behavior via a less dangerous avenue?
- @vanruch prefers a `nextPayment()` access function for _just_ the next payment.
- @barrutko mentioned extracting this to a function that calculate deadlines based on a timestamp (so my guess is it would look like `nextPayment(uint256 timestamp)`?)

If we need to display an expensive view on the frontend, then @vanruch's proposal is to expose variables for that view, but don't add it to the core on-chain interface. @barrutko's proposal should natively be able to give all the info that a frontend would need.

hal909:
> What would be the point of having it as an NFT?

The NFT standard allows for ownership by third parties (auction contracts, single owners). Basically, we can either have DebtInstrument be an NFT where the NFT contract is the "factory" OR we can create a singleton contract where each loan is its own contract, and have a factory that deploys this.

I'm still in favor of using the ERC721 standard.

PR #1115, opened 2021-11-05
🎼 Add require(newScore > 0) to LOCA._updateCreditScore()

skalermo:


contracts/truefi2/LineOfCreditAgency.sol
```diff
@@ -727,6 +729,8 @@ contract LineOfCreditAgency is UpgradeableClaimable, ILineOfCreditAgency {
         address borrower,
         uint256 totalInterest
     ) internal {
+        require(bucketNumber > 0, "LineOfCreditAgency: Bucket of index 0 is not supported");
```
skalermo, line 732: Added this check for extra safety. It seems redundant, and I don't see a way to trigger it. Also that's why I didn't test it.

vanruch (approved):

MiksuJak (approved):

PR #1123, opened 2021-11-10
⛽ Optimise SAFU.liquidate() function

skalermo:


contracts/truefi2/SAFU.sol
```diff
@@ -127,17 +129,19 @@ contract SAFU is ISAFU, UpgradeableClaimable {
             pool.liquidateDebt(debts[i]);
             uint256 owedToPool = debts[i].debt().mul(tokenBalance(debts[i])).div(debts[i].totalSupply());
             uint256 safuTokenBalance = tokenBalance(token);
-            uint256 deficit;
             uint256 toTransfer = owedToPool;
+            DeficiencyToken _deficiencyToken;
+            uint256 deficit;
```
vanruch, line 134: Hmm, can we declare the variables outside the loop? Idk if it saves any gas but don't see a reason why not to

skalermo: Theoretically we could. I mean, once we start entering the if branch below (= no enough safu funds to cover or no funds left) we would be guaranteed to always enter it for all debts left, hence overwriting defined outside the loop variables (helps us avoid using variables from the previous loops).

I would prefer to keep it as is though.

MiksuJak (approved):

vanruch (approved):

PR #1124, opened 2021-11-12
🧶 Expose 'Unwinding condition in a loop' error

MaciejNadolski98:


spec/sandbox/A.spec
```diff
@@ -0,0 +1,6 @@
+methods {
```
MaciejNadolski98, line 1: https://prover.certora.com/output/67403/16e4fa83b44eb08d6f9b/?anonymousKey=5ec76de53a47717af433355f890c25ab3a539481

spec/sandbox/harness/A.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity 0.6.10;
```
MaciejNadolski98, line 1: Rewriting the `_name` variable is done in a loop, hence the error

PR #1125, opened 2021-11-12
🧰 Fix skipped integration tests

skalermo:
This should (hopefully) fix 'tether flush' integration test. I unskipped it as well as one other test. These are still prone to get timedout though.


MiksuJak:
Was this failing, because the wallet that we used for flushing did not have enough ETH to cover gas costs?

skalermo:
> Was this failing, because the wallet that we used for flushing did not have enough ETH to cover gas costs?

No, it was because the caller wasn't the owner. I believe by default it is `powner`.

vanruch:
But why was it failing so randomly then?

MiksuJak (approved):

vanruch (approved):

PR #1126, opened 2021-11-12
🕰️ Fix randomly failing TABRO/TATPO tests

skalermo:
These tests failed about 6 times since the merge of TABRO. The problem is that they test calculations on timestamps and sometimes an extra second slips in causing a fail. I suspect it to be a ganache issue, but I'm not sure.

Since they are in 'truefi2' test folder they are required to pass, and we don't want to get blocked on CI for another 20 or so minutes because of some randomness.

Also I stumbled upon this one import "ganache-time-traveler", I don't think it is being used so I removed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "eslint-plugin-no-only-tests": "^2.6.0",
     "ethereum-mars": "^0.1.7",
     "ethers": "5.0.18",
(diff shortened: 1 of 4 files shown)


vanruch (approved):
TABRO/TATPO goes too far with acronyms

barrutko (approved):

PR #1130, opened 2021-11-16
🃏 Add minimal Fixed Term Loan verification

MaciejNadolski98:


spec/truefi2/FixedTermLoan.spec
```diff
@@ -0,0 +1,20 @@
+using ERC20 as token
```
MaciejNadolski98, line 1: https://prover.certora.com/output/67403/63216e3ee42389ee9ff0?anonymousKey=0a9666a7234eb25a481392955f8c467276696749

spec/truefi2/FixedTermLoan.spec
```diff
@@ -0,0 +1,20 @@
...
+    transferFrom(address, address, uint256) returns (bool) => DISPATCHER(true)
+    approve(address, uint256) returns (bool) => DISPATCHER(true)
+    unlock(address) => DISPATCHER(true)
+}
+
+invariant debtNotGreaterThanSumOfBalanceAndUnpaidDebt()
+    debt() <= token.balanceOf(currentContract) + unpaidDebt()
+    filtered { f -> 
+            f.selector != redeem().selector &&
+            f.selector != initialize(address, address, address, address, address, address, address, uint256, uint256, uint256).selector
```
MaciejNadolski98, line 19: This is our solution for making sure the contract is initialized right now. We are thinking about introducing an extra definition for helping with that, it requires some fixes though, so this will do for now.

spec/truefi2/FixedTermLoan.spec
```diff
@@ -0,0 +1,20 @@
...
+    
+    transfer(address, uint256) returns (bool) => DISPATCHER(true)
+    transferFrom(address, address, uint256) returns (bool) => DISPATCHER(true)
+    approve(address, uint256) returns (bool) => DISPATCHER(true)
+    unlock(address) => DISPATCHER(true)
+}
+
+invariant debtNotGreaterThanSumOfBalanceAndUnpaidDebt()
+    debt() <= token.balanceOf(currentContract) + unpaidDebt()
+    filtered { f ->
```
skalermo, line 17: Can you explain what `invariant` and `filtered` keywords mean?

MaciejNadolski98: Sure, in general, an invariant is a property that is always satisfied for a contract. Important note here is that it needs to be inductive, think of mathematical induction:
1. It needs to be satisfied initially.
2. It must be satisfied after an action assuming it's satisfied beforehand:
```
rule invariant_is_satisfied() {
  require P();
  
  method f;
  calldataarg args;
  env e;
  f(e, args);

  assert P();
}
```

Let's have an example: there is a contract with 4 expressible states - `A`, `B`, `C`, `D`. The initial state is `A` and transitions are:
`A -> B`
`B -> A`
`C -> D`
`D -> C`
This means that property `not D` is always satisfied. However, it's not a proper invariant, because state `C` satisfies the property and yet it transits to `D` which violates it. The proper invariant in this case would be `not C and not D`.


Now, the `filtered` keyword filters which functions to actually take into account when proving those transitions. The same keyword can be used in rules and it's because of how `method` variables are treated: they generate copies of the original rule with the `method` replaced with an actual function of the contract. Therefore, a `require f.selector = redeem().selector` would create a vacuous check, meaning that it would reduce to `require false`.

yuchenlintt (approved):

spec/truefi2/scripts/FixedTermLoan.sh
```diff
@@ -0,0 +1,7 @@
+certoraRun contracts/truefi2/FixedTermLoan.sol \
+  contracts/truefi2/BorrowingMutex.sol \
+  contracts/common/UpgradeableERC20.sol:ERC20 \
+  --verify FixedTermLoan:spec/truefi2/FixedTermLoan.spec \
+  --link FixedTermLoan:token=ERC20 \
+  --short_output \
+  --optimistic_loop
```
MaciejNadolski98, line 7: You may notice that `--rule_sanity` is absent here. That's because sanity checks are giving a false-negative error on a fallback function. There is a task to fix that in a followup PR.

skalermo (approved):

PR #1131, opened 2021-11-16
🥞 Add tfBUSD integration test

vanruch:


yuchenlintt (approved):

test/integration/TrueFiPool2.test.ts
```diff
@@ -55,4 +56,21 @@ describe('TrueFiPool2', () => {
 
     await expect(usdtPool.connect(configGnosis).flush(10000000)).not.to.be.reverted
   })
+
+  it('tfBUSD flow', async () => {
+    const BUSD = '0x4Fabb145d64652a948d72533023f6E7A623C7C53'
+    const holder = provider.getSigner(BUSD_HOLDER)
+    const tfBUSD = '0x1Ed460D149D48FA7d91703bf4890F97220C09437'
+    const busd = Erc20Mock__factory.connect(BUSD, holder)
+    const pool = TrueFiPool2__factory.connect(tfBUSD, holder)
```
skalermo, line 65: This test fails on my local machine. The pool value is non-zero on creation. Any idea why?

hal909 (approved):

test/integration/TrueFiPool2.test.ts
yuchenlintt: That's probably because this links directly against [tfBUSD](https://etherscan.io/address/0x1Ed460D149D48FA7d91703bf4890F97220C09437), which has $60M in it right now.

We can do a before/after check of pool amounts instead of assuming it starts at 0.

PR #1132, opened 2021-11-16
🔐 Unlock mutex when loan is reclaimed from FTLA + remove mutex `locker` delegation to loans

MiksuJak:


contracts/truefi2/FixedTermLoanAgency.sol
```diff
@@ -419,14 +423,19 @@ contract FixedTermLoanAgency is IFixedTermLoanAgency, UpgradeableClaimable {
...
      */
     function totalBorrowed(address borrower, uint8 decimals) public view returns (uint256) {
-        IFixedTermLoan loan = IFixedTermLoan(borrowingMutex.locker(borrower));
-        if (!loanFactory.isLoanToken(loan)) {
+        IFixedTermLoan _loan;
+        if (borrowerLoans[borrower].length > 0) {
+            _loan = borrowerLoans[borrower][0];
+        } else {
+            return 0;
+        }
```
vanruch, line 431: I suggest a small refactor:
```
if (borrowerLoans[borrower].length == 0) { return 0; }

_loan = borrowerLoans[borrower][0];

if (!loanFactory.isLoanToken(_loan)) {
    return 0;
}
```

contracts/truefi2/FixedTermLoanAgency.sol
```diff
@@ -92,6 +92,8 @@ contract FixedTermLoanAgency is IFixedTermLoanAgency, UpgradeableClaimable {
 
     IStakingVault public stakingVault;
 
+    mapping(address => IFixedTermLoan[]) public borrowerLoans;
```
yuchenlintt, line 95: I just remembered that there's a max of one FTL per borrower, so all we would need to track is the single FTL that a borrower might have.

I wonder if there's a standard Bidirectional map implementation for Solidity...

PR #1134, opened 2021-11-17
🗯 Move methods to rule arguments

MaciejNadolski98:


spec/truefi2/BorrowingMutex.spec
```diff
@@ -22,11 +22,10 @@ rule onePlusTwoEqualsThree(uint one, uint two) {
     assert three == 3, "One plus two does not equal three";
```
MaciejNadolski98, line 22: https://prover.certora.com/output/67403/690c3a6742331b91076f?anonymousKey=29827b2881a06762aee10a6966ff992c0d29f5d8

spec/truefi2/BorrowingMutex.spec
```diff
@@ -22,11 +22,10 @@ rule onePlusTwoEqualsThree(uint one, uint two) {
...
 }
 
-rule functionDoesNotBanUnlockedBorrower() {
+rule functionDoesNotBanUnlockedBorrower(method f) {
```
MaciejNadolski98, line 25: This change serves two purposes:
1. This allows the use of `filtered` keyword (mentioned in #1130)
2. This divides variables nicely into two categories - `method f` works differently than `address alice` in that the former generates an extra rule for each possible method, while the latter translates into "for any address alice" in the underlying SMT solver.

spec/truefi2/scripts/BorrowingMutex.sh
```diff
@@ -1,3 +1,2 @@
 certoraRun contracts/truefi2/BorrowingMutex.sol \
-  --verify BorrowingMutex:spec/truefi2/BorrowingMutex.spec \
-  --rule_sanity
```
MaciejNadolski98, line 3: As mentioned in #1130, the `rule_sanity` checks are giving a false-negative error right now, hence temporarily removed.

yuchenlintt (approved):

vanruch (approved):

PR #1135, opened 2021-11-17
🌲 Set up solc and Certora versions in local env

yuchenlintt:


MaciejNadolski98 (approved):

spec/run.sh
```diff
@@ -1,4 +1,54 @@
...
+
+setup_solc() {
+    echo "Setting up solc..." >&2
+    local solc_version="$(extract_version "$(solc --version)")"
+    local compiler_version="$(extract_version "$(extract_json .waffle.json compilerVersion)")"
+
+    echo "Found solc version ${solc_version} and Waffle compiler version ${compiler_version}." >&2
+    if [[ "${solc_version}" != "${compiler_version}" ]]; then
+        echo "Updating solc version to ${compiler_version}..." >&2
+        pip3 install solc-select
```
skalermo, line 27: You might want to consider installing it in python [virtual environment](https://docs.python.org/3/tutorial/venv.html). Same for certora-cli.

skalermo (approved):

MaciejNadolski98 (approved):

spec/run.sh
yuchenlintt: Thanks! I'll look into this in a follow-up PR. Task created.

PR #1136, opened 2021-11-17
✅ Add Certora to CircleCI

yuchenlintt:


.circleci/config.yml
```diff
@@ -66,6 +66,13 @@ jobs:
       - attach_workspace:
           at: .
       - run: yarn lint
+  verify:
+    docker:
+      - image: circleci/python:3-node
```
yuchenlintt, line 71: Any idea how to get a docker image with python3, java, and node? I couldn't find one, at a cursory glance.

yuchenlintt:
After commit dfc459c, I can confirm that Certora is now hooked up to CircleCI, using a proper API key:
https://app.circleci.com/pipelines/github/trusttoken/smart-contracts/4515/workflows/92284ca8-5e35-48d2-8259-ba512f50d0e9/jobs/25222

The verification failures are a known issue mentioned and temporarily bypassed here: https://github.com/trusttoken/smart-contracts/pull/1130#discussion_r751115019 and here: https://github.com/trusttoken/smart-contracts/pull/1134#discussion_r751142262. We'll reintroduce `--rule_sanity` once we can refactor our specs to filter out fallback functions.

Installing `java` would speed up spec CVL syntax checking, but it's not a bottleneck yet so we can ignore it for now.

MaciejNadolski98 (approved):

.circleci/config.yml
skalermo: I couldn't find one either (at least among pre-built ones). According to CircleCI docs you could use [custom docker image](https://circleci.com/docs/2.0/custom-images/), but that seems rather complicated.

skalermo (approved):

PR #1137, opened 2021-11-17
🪓 Hack together Certora solc compiler options from waffle config

yuchenlintt:


spec/run.sh
```diff
@@ -48,12 +53,38 @@ setup_certora() {
     echo "Found CERTORAKEY environment variable." >&2
```
skalermo, line 53: Just letting you know that attempt to read unset CERTORAKEY fails with `spec/run.sh: line 49: CERTORAKEY: unbound variable` (I believe it has something to do with `set -eu`). Not that it matters since we do want to terminate if CERTORAKEY is empty.

spec/run.sh
```diff
@@ -3,10 +3,15 @@ set -eu
 
 extract_json() {
     local json_file="$1"
-    local key="$2"
+    local keys="$2"
 
-    local python_command="import json,sys;obj=json.load(sys.stdin);print(obj['${key}'])"
-    python3 -c "${python_command}" < "${json_file}"
+    python3 <<EOF
+import json,sys
```
skalermo, line 9:
```suggestion
import json
```

skalermo (approved):

spec/run.sh
yuchenlintt: Yep! It's not the most helpful error message, but this was intended behavior to make sure we fail early before sending a spec to Certora when we don't have a CERTORAKEY.

@skalermo I've put in a request to grant you an API key.

yuchenlintt:
Closing POC. Could retry this in Ragnarok repo.

PR #1138, opened 2021-11-18
🪃 Check for fallback functions in rules

MaciejNadolski98:


spec/truefi2/BorrowingMutex.spec
```diff
@@ -34,6 +34,8 @@ rule functionDoesNotBanUnlockedBorrower() {
         address locker = anyReasonableAddress();
```
MaciejNadolski98, line 34: https://prover.certora.com/output/67403/aebc1e73f163fd7941bc?anonymousKey=3923171e43a2f72efa42f42cc27897ebb8965c7a

spec/truefi2/BorrowingMutex.spec
```diff
@@ -49,7 +51,11 @@ rule functionDoesNotUnbanBorrower {
     env e;
     require isReasonableEnv(e);
     calldataarg args;
-    f(e, args);
+    if (f.isFallback) {
+        f@withrevert(e, args);
```
MaciejNadolski98, line 55: We've decided on this pattern for fallback functions, but also for deprecated functions that will always fail and functions whose failure is determined by assumptions. This is for the sanity checks which require that assertions are reachable in all rules.

spec/truefi2/BorrowingMutex.spec
yuchenlintt: 👍 

Once #1136 gets checked in to main, we won't need to manually link this any more.

yuchenlintt (approved):

vanruch (approved):

PR #1143, opened 2021-11-23
⛷️ Avalanche TUSD: make mintPauser and mintRatifiers set by onlyOwner

hal909:
The owner should set mint ratifiers and mint pausers in the TUSD system, not the registry admin. This change fixes the permissions for setting these values on the Avalanche TUSD controller and adds an improved script that when run by the owner will upgrade the Avalanche TUSD contract and set 2 new mint ratifiers.


scripts/avalanche_script.ts
```diff
@@ -0,0 +1,83 @@
...
+  const controller = AvalancheTokenController__factory.connect('0xc3a247acE92A6A36FB69F70873A85fB8a66aDC1c', wallet)
+  const proxy = OwnedUpgradeabilityProxy__factory.connect('0xc3a247acE92A6A36FB69F70873A85fB8a66aDC1c', wallet)
+  const token = TrueUsd__factory.connect('0x1C20E891Bab6b1727d14Da358FAe2984Ed9B59EB', wallet)
+  const tokenProxy = OwnedUpgradeabilityProxy__factory.connect('0x1C20E891Bab6b1727d14Da358FAe2984Ed9B59EB', wallet)
+
+  // deploy, upgrade, set ratifiers
+  controller = await deployController(wallet)
+  await upgradeController(proxy, controller.address)
+  await setIsMintRatifier('0x083B59F244f8BAcbc79282Cdd623686324C962AC', 1, controller)
+  await setIsMintRatifier('0xb47DeA8731Fd846065294771Ab68B3Ee1FC6E880', 1, controller)
```
yuchenlintt, line 37: I see on Etherscan that these addresses are used to call `TUSD.ratifyMint()` on mainnet. Do we have independent attestation that these are correct? I couldn't find them in the usual places where we track addresses internally.

contracts/avalanche/AvalancheTokenController.sol
```diff
@@ -449,11 +449,11 @@ contract AvalancheTokenController {
         registryAdmin = admin;
     }
 
-    function setIsMintPauser(address account, bool status) external onlyRegistryAdmin {
+    function setIsMintPauser(address account, bool status) external onlyOwner {
         isMintPauser[account] = status;
     }
 
-    function setIsMintRatifier(address account, bool status) external onlyRegistryAdmin {
+    function setIsMintRatifier(address account, bool status) external onlyOwner {
```
yuchenlintt, line 456: Are these changes we'd like to backport to mainnet TUSD? If so, let's create a task and/or submit a separate PR.

yuchenlintt (dismissed):
LGTM! Just two questions meant for follow-up.

scripts/avalanche_script.ts
hal909: I can get additional confirmation from these addresses by having them send me AVAX

contracts/avalanche/AvalancheTokenController.sol
hal909: Yes. I looked over some of the contracts and there are some issues in mainnet TUSD as well.

scripts/avalanche_script.ts
hal909: confirmed. resolving this thread

yuchenlintt (dismissed):

vanruch (approved):

scripts/avalanche_script.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
yuchenlintt, line 1: The correct response to failing lint shouldn't be disabling lint for the file...

But let's check this PR in now and resolve that later. Putting out a fire is more important than cleaning up lint.

yuchenlintt (approved):

PR #1149, opened 2021-11-24
🦹 Transfer fee to the manager

vanruch:


vanruch:
As per @yuchenlintt proposal https://github.com/trusttoken/smart-contracts/pull/1111#issuecomment-962095576

test/ragnarok/ManagedPortfolio.test.ts
```diff
@@ -44,6 +45,8 @@ describe('ManagedPortfolio', () => {
       token.address,
       bulletLoans.address,
       YEAR,
+      1000,
```
krzysztof-jelski, line 48:
```suggestion
      TEN_PERCENT,
```

contracts/ragnarok/ManagedPortfolio.sol
```diff
@@ -22,17 +25,23 @@ contract ManagedPortfolio is IERC721Receiver, ERC20, Ownable {
...
 
     event BulletLoanCreated(uint256 id);
 
     constructor(
         IERC20WithDecimals _underlyingToken,
         BulletLoans _bulletLoans,
-        uint256 _duration
+        uint256 _duration,
+        BIP _managerFee,
+        address _manager
```
krzysztof-jelski, line 38: I feel like we assumed that `manager` is just the `msg.sender` (the wallet that creates the portfolio). But perhaps it's better to make it explicit.

contracts/ragnarok/types/BIP.sol
```diff
@@ -0,0 +1,14 @@
...
+pragma solidity 0.8.10;
+
+type BIP is uint256;
+
+library BIPMath {
+    function add(BIP a, BIP b) internal pure returns (BIP c) {
+        c = BIP.wrap(BIP.unwrap(a) + BIP.unwrap(b));
+    }
+
+    function mul(BIP a, uint256 b) internal pure returns (uint256 c) {
```
vanruch, line 11: I worry about having division in multiplication that is totally not obvious. This may lead to precision loss. Maybe it should return BP as well and at the and you are should be forced to call something like `normalise` that will divide by 10000?

.solhintignore
```diff
@@ -0,0 +1 @@
+contracts/ragnarok/types/**/*.sol
```
MiksuJak, line 1: Isn't it too much of a shortcut?

vanruch: solhint can't parse this type alias syntax and it's likely that all contracts in this folder will have one

test/ragnarok/ManagedPortfolio.test.ts
```diff
@@ -72,11 +77,45 @@ describe('ManagedPortfolio', () => {
       const creationTimestamp = (await provider.getBlock(deployTx.blockHash)).timestamp
       expect(await portfolio.endDate()).to.equal(creationTimestamp + YEAR)
     })
+
+    it('sets owner', async () => {
```
MiksuJak, line 81: This test is duplicated

vanruch: looks like somebody moved it to top and it merged

MiksuJak (approved):

PR #1152, opened 2021-11-25
⚙️ Introduce protocol config

vanruch:


contracts/ragnarok/PortfolioConfig.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
+import {BP, BPMath} from "./types/BP.sol";
+import {Ownable} from "@openzeppelin/contracts4/access/Ownable.sol";
```
skalermo, line 5: Can you swap these lines?

test/utils/constants.ts
```diff
@@ -7,3 +7,5 @@ export const MAX_BURN_BOUND = constants.MaxUint256.sub(constants.MaxUint256.mod(
 export const MAX_APY = 100_000
 export const DAY = 60 * 60 * 24
 export const YEAR = 365 * DAY
+
+export const ONE_PERCENT = 100
```
skalermo, line 11: I think for consistency we should do something with `TEN_PERCENT` const introduced in #1149: either replace it with 10 * ONE_PERCENT or add new const here.

skalermo (approved):

PR #1153, opened 2021-11-25
👮🏽‍♂️ Verify KYL signature

MiksuJak:


contracts/ragnarok/ManagedPortfolio.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.10;
 
+import "@openzeppelin/contracts4/utils/cryptography/ECDSA.sol";
```
skalermo, line 4:
```suggestion
import {ECDSA} from "@openzeppelin/contracts4/utils/cryptography/ECDSA.sol";
```

test/ragnarok/ManagedPortfolio.test.ts
```diff
@@ -320,8 +332,31 @@ describe('ManagedPortfolio', () => {
...
+    })
+
+    it('returns false for invalid message', async () => {
+      const signature = await signMessage(lender, 'other message')
+      expect(await portfolio.isSignatureValid(lender.address, signature)).to.be.false
+    })
+
+    it('returns false for invalid signer', async () => {
+      const signature = await signMessage(manager, DEPOSIT_MESSAGE)
+      expect(await portfolio.isSignatureValid(lender.address, signature)).to.be.false
```
skalermo, line 348: Can we also have a test where signature is signed by lender, but isSignatureValid() called for e.g. manager?

MiksuJak: `isSignatureValid` is a view function, that does not depend on `msg.sender` in any way

vanruch (approved):

PR #1154, opened 2021-11-26
📅 Manager can change loan status

skalermo:


contracts/ragnarok/ManagedPortfolio.sol
```diff
@@ -16,17 +16,25 @@ interface IERC20WithDecimals is IERC20 {
 contract ManagedPortfolio is IERC721Receiver, ERC20, Manageable {
     using BPMath for BP;
 
+    enum LoanStatus {
+        Active,
```
MiksuJak, line 20:
```suggestion
        Issued,
```

contracts/ragnarok/ManagedPortfolio.sol
```diff
@@ -16,17 +16,25 @@ interface IERC20WithDecimals is IERC20 {
 contract ManagedPortfolio is IERC721Receiver, ERC20, Manageable {
     using BPMath for BP;
 
+    enum LoanStatus {
```
vanruch, line 19: I think this should be kept in BL (as per UML)

PR #1171, opened 2022-02-24
🍼 Deploy TrueUSDNoDelegate to mainnet

yuchenlintt:

PR #1176, opened 2022-05-16
🧾 Chainlink Proof of Reserve

pkuchtatt:


test/true-currencies/TrueCurrencyWithPoR.test.ts
```diff
@@ -0,0 +1,184 @@
...
+  it('should mint successfully when feed decimals < TrueCurrency decimals', async () => {
+    // Re-deploy a mock aggregator with fewer decimals
+    const currentTusdSupply = await token.totalSupply()
+    const validReserve = currentTusdSupply.div(exp(1, 12)).add(AMOUNT_TO_MINT)
+    const mockV3AggregatorWith6Decimals = await new MockV3Aggregator__factory(owner).deploy('6', validReserve)
+    // Set feed and heartbeat on newly-deployed aggregator
+    await token.setChainReserveFeed(mockV3AggregatorWith6Decimals.address)
+    expect(await token.chainReserveFeed()).to.equal(mockV3AggregatorWith6Decimals.address)
+    await token.setChainReserveHeartbeat(ONE_DAY_SECONDS)
+    expect(await token.chainReserveHeartbeat()).to.equal(ONE_DAY_SECONDS)
```
tt-krzysztof, line 99: We don't need to verify every time that setting the heartbeat works. It's enough to verify it in one test.

test/true-currencies/TrueCurrencyWithPoR.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    // Mint TUSD
+    const balanceBefore = await token.balanceOf(owner.address)
+    await token.mint(owner.address, AMOUNT_TO_MINT)
+    expect(await token.balanceOf(owner.address)).to.equal(balanceBefore.add(AMOUNT_TO_MINT))
+  })
+
+  it('should revert if TrueCurrency supply > proof-of-reserves', async () => {
+    // Re-deploy aggregator with fewer TUSD in reserves
```
tt-krzysztof, line 140: We'r not re-deploying here, just updating the oracle's answer.

test/true-currencies/TrueCurrencyWithPoR.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    expect(await token.balanceOf(owner.address)).to.equal(balanceBefore)
+  })
+
+  it('should revert if the feed is not updated within the heartbeat', async () => {
+    // Set heartbeat to 1 day
+    await token.setChainReserveHeartbeat(ONE_DAY_SECONDS)
+    expect(await token.chainReserveHeartbeat()).to.equal(ONE_DAY_SECONDS)
+
+    // Heartbeat is set to 1 day, so fast-forward 2 days
+    await network.provider.send('evm_increaseTime', [2 * ONE_DAY_SECONDS])
```
tt-krzysztof, line 163: There's a helper `timeTravel` function in `utils`, which you can use here.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,98 @@
...
+     */
+    function _mint(address account, uint256 amount) internal virtual override {
+        if (chainReserveFeed == address(0)) {
+            super._mint(account, amount);
+            return;
+        }
+
+        // Get latest proof-of-reserves from the feed
+        (, int256 answer, , uint256 updatedAt, ) = IChainlinkAggregatorV3(chainReserveFeed).latestRoundData();
+        require(answer > 0, "TrueCurrency: Invalid answer from PoR feed");
```
tt-krzysztof, line 41: That got me thinking: in theory `0` is still a valid answer (eg. there are no reserves, every TUSD has been redeemed). But then, if it's `0`, we can't mint anyway. So I believe it's ok in the end.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,98 @@
...
+
+        // Check the answer is fresh enough (i.e., within the specified heartbeat)
+        uint256 oldestAllowed = block.timestamp.sub(chainReserveHeartbeat, "TrueCurrency: Invalid timestamp from PoR feed");
+        require(updatedAt >= oldestAllowed, "TrueCurrency: PoR answer too old");
+
+        // Get required info about total supply & decimals
+        uint256 currentSupply = totalSupply();
```
tt-krzysztof, line 48: Move 2 lines down (closer to usage).

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,98 @@
...
+        require(newHeartbeat <= MAX_AGE, "TrueCurrency: PoR heartbeat greater than MAX_AGE");
+
+        // Allowable scenarios:
+        //  - heartbeat is not initialised (0 instead of default MAX_AGE); OR
+        //  - new heartbeat is different AND
+        //    new heartbeat is not resetting to default while current heartbeat is already set to the default
+        require(
+            chainReserveHeartbeat == 0 ||
+                (newHeartbeat != chainReserveHeartbeat && !(newHeartbeat == 0 && chainReserveHeartbeat == MAX_AGE)),
+            "TrueCurrency: new chainReserveHeartbeat must be different to current heartbeat"
```
tt-krzysztof, line 92: Add test cases covering this condition.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,98 @@
...
+        );
+
+        emit NewChainReserveHeartbeat(chainReserveHeartbeat, newHeartbeat);
```
tt-krzysztof, line 95: Consider adding tests for event emission.

yuchenlintt:
🚨 I haven't had a chance to review most of this code yet, but a critical first step is that we need to first merge #1170 into `main`, followed by merging #1171 into `main`.

- #1170 fixes a critical vuln in `mainnet` TUSD that OpenZeppelin reported to us a few months ago.
- #1171 deploys this fixed contract to `mainnet`. This upgrade has already happened, so for this PR we can skip the final diff check that we'd typically do for changes to `deployments-mainnet.json`.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,99 @@
...
+ *  This contract implements an additional check against a Proof-of-Reserves feed before
+ *  allowing tokens to be minted.
+ */
+abstract contract TrueCurrencyWithPoR is TrueCurrency, IPoRToken {
+    using SafeMath for uint256;
+
+    uint256 public constant MAX_AGE = 7 days;
+
+    constructor() public {
+        chainReserveHeartbeat = MAX_AGE;
```
yuchenlintt, line 22: If we're doing an upgrade of TUSD, then this constructor won't get called. We'll have to tell Techteryx to queue `setChainReserveHeartbeat(7 days)` from the multisig after the upgrade call.

contracts/true-currencies/TrueCurrencyWithPoR.sol
yuchenlintt: If this is a redundant `require()`, then I would prefer to remove it. Simpler code is better!

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,99 @@
...
+        super._mint(account, amount);
+    }
+
+    /**
+     * @notice Sets a new feed address
+     * @dev Admin function to set a new feed
+     * @param newFeed Address of the new feed
+     */
+    function setChainReserveFeed(address newFeed) external override onlyOwner returns (uint256) {
+        require(newFeed != chainReserveFeed, "TrueCurrency: new chainReserveFeed must be different to current feed");
```
yuchenlintt, line 72: Do we need this require? I don't think it hurts to set `chainReserveFeed` to the same value.

contracts/true-currencies/TrueCurrencyWithPoR.sol
yuchenlintt: Do we need this require? It shouldn't hurt to set chainReserveHeartbeat to the same value as before, and if we want an explicit `MIN_AGE == 1` then we should add that as a constant -- although personally I don't see much difference between 1 second of heartbeat and 0 seconds for Chainlink's update frequency (at most once per block).

yuchenlintt (commented):
(Possibly in a follow-up PR) it's a good idea to add a simulation test that forks the `mainnet` contract address `0x0000000000085d4780B73119b644AE5ecd22b376`, performs the upgrade, and checks that a user is still able to mint, redeem, etc.

contracts/common/interface/IPoRToken.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.6.10;
```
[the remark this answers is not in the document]
yuchenlintt: Unfortunately probably not, since the existing contract has a lot of legacy code that we would need to inspect carefully for upgrade compatibility.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,99 @@
...
+        require(currentSupply + amount <= reserves, "TrueCurrency: total supply would exceed reserves after mint");
+        super._mint(account, amount);
+    }
+
+    /**
+     * @notice Sets a new feed address
+     * @dev Admin function to set a new feed
+     * @param newFeed Address of the new feed
+     */
+    function setChainReserveFeed(address newFeed) external override onlyOwner returns (uint256) {
```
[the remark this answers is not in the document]
yuchenlintt: I would recommend against a zero address check, since `address(0)` in this current implementation acts as a sentinel value that indicates proof of reserve is disabled.

This might be necessary to call in case, for instance, the Chainlink aggregator goes down for a period that is longer than `MAX_AGE`.

test/true-currencies/TrueCurrencyWithPoR.test.ts
pkuchtatt: Removed redundant checks.

test/true-currencies/TrueCurrencyWithPoR.test.ts
pkuchtatt: Removed the comment.

contracts/true-currencies/common/ProxyStorage.sol
```diff
@@ -56,6 +56,10 @@ contract ProxyStorage {
     mapping(address => bool) isBlacklisted;
     mapping(address => bool) public canBurn;
 
+    // PoR feed-related variables
+    address public chainReserveFeed;
+    uint256 public chainReserveHeartbeat;
```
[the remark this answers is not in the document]
yuchenlintt: Nice catch!

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Done.

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Done.

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Added tests covering the condition.

@yuchenlintt That's a good question. The code was originally written by a Chainlink eng. It seems for some reason he thought that not allowing to set `chainReserveHeartbeat` to the same value was important. I can see this would lead to emission of `NewChainReserveHeartbeat(<n>, <n>)` events, which would be a little silly, but we could as well skip emitting instead of reverting in this case. What do you think?

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Added tests.

contracts/true-currencies/common/ProxyStorage.sol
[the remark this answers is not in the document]
pkuchtatt: Done.

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: As mentioned earlier, I think MAX_AGE is an arbitrary constant, I don't think anybody knows for sure what a really useful/practical MAX_AGE is at the moment. I think it's probably best not to constrain the possible values of `chainReserveHeartbeat` in the code and remove `MAX_AGE` altogether. Please do let me know if you think otherwise.

We need to be able to disable temporarily PoR if Armanino, Chainlink or one of our banking partners' API (most likely) is down for an extended period of time, this will be done be setting `chainReserveFeed` to `0`. We should be able to do it quickly, e.g. in case of a VIP mint, and I think calling `setChainReserveFeed` should be quicker and cheaper than upgrading the contract. Do you agree?

Also, we plan to deploy it on Ethereum and Avalanche, which, I assume, will have different feed addresses, so I think hardcoding `CHAIN_RESERVE_FEED` in the code is probably not a good idea.

contracts/true-currencies/TrueCurrencyWithPoR.sol
yuchenlintt: I think it's fine to emit an event that says nothing changed. Anyone who consumes the event should be able to handle this without any issues.

In smart contracts I'd much prefer to reason about simple code than a bunch of edge cases. And in this example I don't see any failure modes that would force us to add this require (which might instead backfire and prevent us from newHeartbeat == 0 if for some reason we need to, for instance).

pkuchtatt:
> 🚨 I haven't had a chance to review most of this code yet, but a critical first step is that we need to first merge #1170 into `main`, followed by merging #1171 into `main`.
> 
> * [✂️ Disable delegation in TUSD #1170](https://github.com/trusttoken/contracts-pre22/pull/1170) fixes a critical vuln in `mainnet` TUSD that OpenZeppelin reported to us a few months ago.
> * [🍼 Deploy TrueUSDNoDelegate to mainnet #1171](https://github.com/trusttoken/contracts-pre22/pull/1171) deploys this fixed contract to `mainnet`. This upgrade has already happened, so for this PR we can skip the final diff check that we'd typically do for changes to `deployments-mainnet.json`.

@yuchenlintt Thank you for the info! Please let us know when this is ready. Speaking about the repo, I noticed CI always fails on the `verify` task (certora). Do you have an idea why? If not, do you think we can remove the task from CI?

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,99 @@
...
+ * @title TrueCurrencyWithPoR
+ * @dev TrueCurrencyPoR is an ERC20 with blacklist & redemption addresses.
+ *  Please see TrueCurrency for the implementation that this contract inherits from.
+ *  This contract implements an additional check against a Proof-of-Reserves feed before
+ *  allowing tokens to be minted.
+ */
+abstract contract TrueCurrencyWithPoR is TrueCurrency, IPoRToken {
+    using SafeMath for uint256;
+
+    uint256 public constant MAX_AGE = 7 days;
```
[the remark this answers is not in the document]
yuchenlintt: Agree, I don't think it's necessary to have a MAX_AGE set. The setter is only callable by `owner` anyway, which for TUSD is the same as the `proxyOwner` that can override `MAX_AGE` via upgrade.

contracts/true-currencies/TrueCurrencyWithPoR.sol
yuchenlintt: Having the option to set `chainReserveFeed` to 0 sounds to me like enough reason to keep this as a storage variable! And yeah, I'd overlooked that Ethereum mainnet and Avalanche will have different addresses. Definitely shouldn't hardcode in that case.

yuchenlintt:
> @yuchenlintt Thank you for the info! Please let us know when this is ready.

These PRs have been ready to merge since February, but no one's approved them :)

> I noticed CI always fails on the verify task (certora). Do you have an idea why? If not, do you think we can remove the task from CI?

My guess is our dev environment is probably out of date? Calls to Certora go to an external service, so if we haven't updated our client version then there might be some errors. Feel free to remove this check from CI -- we'd only gotten around to verifying code that has since become obsolete.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,98 @@
...
+    function _mint(address account, uint256 amount) internal virtual override {
+        if (chainReserveFeed == address(0)) {
+            super._mint(account, amount);
+            return;
+        }
+
+        // Get latest proof-of-reserves from the feed
+        (, int256 signedReserves, , uint256 updatedAt, ) = IChainlinkAggregatorV3(chainReserveFeed).latestRoundData();
+        require(signedReserves > 0, "TrueCurrency: Invalid answer from PoR feed");
+        uint256 reserves = uint256(signedReserves);
```
cds95, line 42: do we need to check if `signedReserves` isn't too large before converting to `uint256`?

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,98 @@
...
+     */
+    function _mint(address account, uint256 amount) internal virtual override {
+        if (chainReserveFeed == address(0)) {
+            super._mint(account, amount);
+            return;
+        }
+
+        // Get latest proof-of-reserves from the feed
+        (, int256 signedReserves, , uint256 updatedAt, ) = IChainlinkAggregatorV3(chainReserveFeed).latestRoundData();
+        require(signedReserves > 0, "TrueCurrency: Invalid answer from PoR feed");
```
cds95, line 41: Just a small suggestion to save gas but maybe we can shorten the `require` messages

contracts/common/interface/IChainlinkAggregatorV3.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.6.10;
+
+interface IChainlinkAggregatorV3 {
```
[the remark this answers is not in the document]
pkuchtatt: Done.

contracts/true-currencies/TrueCurrencyWithPoR.sol
yosriady: I would consider skiping the storage write and event emission if it's set to the same value: 

```
if (newFeed != chainReserveFeed) {
  ...
}
```

contracts/true-currencies/TrueCurrencyWithPoR.sol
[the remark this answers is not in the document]
pkuchtatt: Reworked the code so `chainReserveHeartbeat` is initialised in the constructor to 7 days but can be later reset to any value.

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Removed require and added ifs as suggested to avoid writing to storage and emitting events when no change.

bryanjowers:
Thanks @yosriady and @cds95 for taking a quick look over this PR. @pkuchtatt  Glad we could provide a few helpful bits of feedback, but want to call out this shouldn't be considered a complete review or audit. excited about this project!

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,99 @@
...
+            return;
+        }
+
+        // Get latest proof-of-reserves from the feed
+        (, int256 answer, , uint256 updatedAt, ) = IChainlinkAggregatorV3(chainReserveFeed).latestRoundData();
+        require(answer > 0, "TrueCurrency: Invalid answer from PoR feed");
+
+        // Check the answer is fresh enough (i.e., within the specified heartbeat)
+        uint256 oldestAllowed = block.timestamp.sub(chainReserveHeartbeat, "TrueCurrency: Invalid timestamp from PoR feed");
+        require(updatedAt >= oldestAllowed, "TrueCurrency: PoR answer too old");
```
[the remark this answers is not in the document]
pkuchtatt: Good point! It made me realise that we do have to constrain values of `chainReserveHeartbeat`. Added `MAX_CHAIN_RESERVE_HEARTBEAT` and set it to 30 days, it's hopefully more than we will never need but still not cause the overflow here. Note, `INITIAL_CHAIN_RESERVE_HEARTBEAT` is set to 7 days now.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,90 @@
...
+     * @dev Admin function to set the heartbeat
+     * @param newHeartbeat Value of the age of the latest update from the feed
+     */
+    function setChainReserveHeartbeat(uint256 newHeartbeat) external override onlyOwner returns (uint256) {
+        require(newHeartbeat <= MAX_CHAIN_RESERVE_HEARTBEAT, "TrueCurrency: PoR heartbeat too long");
+
+        if (newHeartbeat != chainReserveHeartbeat) {
+            emit NewChainReserveHeartbeat(chainReserveHeartbeat, newHeartbeat);
+            chainReserveHeartbeat = newHeartbeat;
+        }
```
yuchenlintt, line 88: Same comment as above re no-op events:
```suggestion
        emit NewChainReserveHeartbeat(chainReserveHeartbeat, newHeartbeat);
        chainReserveHeartbeat = newHeartbeat;
```

yuchenlintt (changes requested):

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Committed suggestion.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,90 @@
...
+ * @title TrueCurrencyWithPoR
+ * @dev TrueCurrencyPoR is an ERC20 with blacklist & redemption addresses.
+ *  Please see TrueCurrency for the implementation that this contract inherits from.
+ *  This contract implements an additional check against a Proof-of-Reserves feed before
+ *  allowing tokens to be minted.
+ */
+abstract contract TrueCurrencyWithPoR is TrueCurrency, IPoRToken {
+    using SafeMath for uint256;
+
+    uint256 public constant MAX_CHAIN_RESERVE_HEARTBEAT = 30 days;
```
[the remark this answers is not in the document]
pkuchtatt: So, I realised here https://github.com/trusttoken/contracts-pre22/pull/1176#discussion_r879425181 that we should constrain the values of `chainReserveHeartbeat` because it can potentially cause an underflow in this code:

```
require(updatedAt >= block.timestamp.sub(chainReserveHeartbeat), "TrueCurrency: PoR answer too old");
```

I think it's better to revert with a clear error message that will help in debugging rather than potentially allow the code to fail with an underflow. What do you think?

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * @notice Sets the feed's heartbeat expectation
+     * @dev Admin function to set the heartbeat
+     * @param newHeartbeat Value of the age of the latest update from the feed
+     */
+    function setChainReserveHeartbeat(uint256 newHeartbeat) external override onlyOwner returns (uint256) {
+        require(newHeartbeat <= MAX_CHAIN_RESERVE_HEARTBEAT, "TrueCurrency: PoR heartbeat too long");
+
```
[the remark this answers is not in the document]
pkuchtatt: As explained above, too big `maxChainReserveHeartbeat` can cause an underflow in one of the `require`s.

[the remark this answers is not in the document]
yuchenlintt: Here's my reply: https://github.com/trusttoken/contracts-pre22/pull/1176#discussion_r921271248

In brief: the only possible way for my proposal to underflow would be if `block.timestamp < updatedAt`, which should not happen assuming `block.timestamp` is monotonic increasing. If this change means we can cut out extra code, I'd be happy to accept a slightly less legible error in that situation.

contracts/true-currencies/TrueCurrencyWithPoR.sol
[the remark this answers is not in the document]
pkuchtatt: Great idea! `updatedAt` is an answer from chainlink aggregator contract, I imagine they may be scenarios where we get a wrong value here, e.g. when owner sets `chainReserveFeed` to a wrong/malicious contract, and it's causing an underflow, however we can and probably should protect ourselves from this with something like
```
require(block.timestamp >= updatedAt)
```

Will implement this shortly, sorry about the delay, was setting up a new workstation and also found that an integration test is failing on the main branch, here is a PR with a fix: https://github.com/trusttoken/contracts-pre22/pull/1180 Can you please review and merge?

README.md
```diff
@@ -1,6 +1,6 @@
 # 💰 TrueFi Smart Contracts
 
-[![Coverage Status](https://coveralls.io/repos/github/trusttoken/smart-contracts/badge.svg?branch=main)](https://coveralls.io/github/trusttoken/smart-contracts?branch=main)
+[![Coverage Status](https://coveralls.io/repos/github/trusttoken/contracts-pre22/badge.svg?branch=main)](https://coveralls.io/github/trusttoken/contracts-pre22?branch=main)
```
yuchenlintt, line 3: nit: revert to original for a concise PR
```suggestion
[![Coverage Status](https://coveralls.io/repos/github/trusttoken/smart-contracts/badge.svg?branch=main)](https://coveralls.io/github/trusttoken/smart-contracts?branch=main)
```

README.md
```diff
@@ -32,9 +32,9 @@ Smart contracts implementing True Gold - a synthetic gold on Ethereum blockchain
 # 🧰 Installation
 In order to compile the contracts first clone the repository. In order to do that run
 ```
-git clone git@github.com:trusttoken/smart-contracts.git
+git clone git@github.com:trusttoken/contracts-pre22.git
```
yuchenlintt, line 35:
```suggestion
git clone git@github.com:trusttoken/smart-contracts.git
```

README.md
```diff
@@ -32,9 +32,9 @@ Smart contracts implementing True Gold - a synthetic gold on Ethereum blockchain
...
 ```
-Then enter `smart-contracts` directory and install all necessary dependencies by running
+Then enter `contracts-pre22` directory and install all necessary dependencies by running
```
yuchenlintt, line 37:
```suggestion
Then enter `smart-contracts` directory and install all necessary dependencies by running
```

yuchenlintt (commented):
LGTM, thanks for carrying this through!

For the sake of a cleaner commit history, I suggest we revert the irrelevant changes and handle them in a separate PR. I'll commit these as suggested changes and then approve this PR.

yuchenlintt (approved):

tt-krzysztof (approved):

yuchenlintt:
#1170 and #1171 have been merged, so this should be ready as well now!

PR #1186, opened 2022-08-26
🟰 Require PoR feed decimals to be equal token's decimals

pkuchtatt:
Fixed potential vulnerability in the Proof of Reserve implementation by requiring that chain reserve feed's decimals are exactly the same as token's decimals.

Removed no longer needed normalisation code.

Also, fixed CI by disabling linting on `TrueUSDNoDelegate.sol`, recently merged file that must not change as it was used to verify previous deployment of the contract.


pkuchtatt:
Temporarily closed, will be reopened.

.prettierignore
```diff
@@ -1,3 +1,4 @@
 contracts/governance/GovernorAlpha.sol
 contracts/governance/Timelock.sol
 contracts/governance/VoteToken.sol
+contracts/true-currencies/tokens/TrueUSDNoDelegate.sol
```
yuchenlintt, line 4: What's the .prettierignore reason? Can we fix it during this upgrade or would it be too much effort?

.solhintignore
```diff
@@ -0,0 +1 @@
+contracts/true-currencies/tokens/TrueUSDNoDelegate.sol
```
yuchenlintt, line 1: Ditto here: can we fix the .solhintignore reason during this upgrade? Okay to decline if this would be too large a change.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -49,13 +49,11 @@ abstract contract TrueCurrencyWithPoR is TrueCurrency, IPoRToken {
         // Get required info about total supply & decimals
         uint8 trueDecimals = decimals();
         uint8 reserveDecimals = AggregatorV3Interface(chainReserveFeed).decimals();
+
+        // Decimals of the Proof of Reserve feed must be the same as the token's.
+        require(reserveDecimals == trueDecimals, "Unexpected decimals of PoR feed");
```
yuchenlintt, line 54: Since we no longer need to use the `trueDecimals` and `reserveDecimals` local vars, I would recommend shortening this to a single line to reduce clutter and improve clarity (adding a local variable means auditors will search for all possible abuses of this variable).
```suggestion
        // Decimals of the Proof of Reserve feed must be the same as the token's.
        require(decimals() == AggregatorV3Interface(chainReserveFeed).decimals(), "Unexpected decimals of PoR feed");
```

yuchenlintt: Actually, let's move this decimals check up above the `latestRoundData()` call to align with the [checks-effects-interactions](https://docs.soliditylang.org/en/develop/security-considerations.html?highlight=check%20effects#use-the-checks-effects-interactions-pattern) pattern. Now that we don't consume decimals here, variable proximity-to-use is no longer a concern.

(To expand on this a little more: `decimals()` is a cheaper check than calling `latestRoundData()` to get a struct. Also, unlike `latestRoundData()`, `decimals()` should never change between invocations, so in some sense having a contract with the right decimals is a precondition for having correct round data from that contract.)

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -49,13 +49,11 @@ abstract contract TrueCurrencyWithPoR is TrueCurrency, IPoRToken {
...
+
         uint256 currentSupply = totalSupply();
```
yuchenlintt, line 56: Can we inline `totalSupply()` into the check below, now that the `currentSupply` variable is no longer used elsewhere?

yuchenlintt (changes requested):
I reopened this PR now that the prelim report is back!

Some minor refactoring to simplify `_mint()` even further, as well as an open question (the answer is possibly a large, out-of-scope refactor).

yuchenlintt:
> Also, fixed CI by disabling linting on `TrueUSDNoDelegate.sol`, recently merged file that must not change as it was used to verify previous deployment of the contract.

Ah, I see now! Actually now that this file has been committed to `main`, we can change it arbitrarily or even delete it 😛 

The important invariant to maintain was that the exact version of this file that was deployed to `mainnet` was already merged in to the repo at a commit where the `deployments-mainnet.json` file was updated. This way, anyone can run `git blame` on the deployments file to:
- find the commit where this address was deployed
- build the repo at that commit, and
- confirm the deployed bytecode matches the local build at that commit.

This is why it was so important to merge the two no-delegate hotfix PRs _before_ the proof of reserve PRs. Otherwise the commit history mangles the build for a past deployment of TUSD. Hope this helps!

tl;dr Feel free to delete `TrueUSDNoDelegate.sol` from the repo, now that it's been merged in the correct order.

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Done.

.prettierignore
pkuchtatt: Thanks for the explanation later on, makes sense. Removed `TrueUSDNoDelegate.sol` and references to it in the linters' ignore files.

.solhintignore
pkuchtatt: Done.

contracts/true-currencies/TrueCurrencyWithPoR.sol
yuchenlintt: Thanks for the first fix (inlining)!

Could you also move this require and its comments up above the `latestRoundData()` call? The order shouldn't affect the conditions for when this function call would revert, but it's a more defensive code style and improves clarity:
- aggregator decimals should never change, so we want to fail earlier if this ever happens
- aggregator latestRoundData() often changes, so only _some_ calls to it should revert

Concretely, if we're in a situation where _both_ the updatedAt heartbeat is stale _and_ decimals are wrong, then we should revert with the reason that decimals are wrong, because it indicates a more serious problem that needs to be addressed. Thanks!

pkuchtatt: Done, sorry I forgot about the reordering earlier.

yuchenlintt (dismissed):
Thanks for the fixes! LGTM now

test/true-currencies/TrueCurrencyWithPoR.test.ts
```diff
@@ -83,38 +83,40 @@ describe('TrueCurrency with Proof-of-reserves check', () => {
     expect(await token.balanceOf(owner.address)).to.equal(balanceBefore.add(AMOUNT_TO_MINT))
   })
 
-  it('should mint successfully when feed decimals < TrueCurrency decimals', async () => {
+  it('should revert mint when feed decimals < TrueCurrency decimals', async () => {
     // Re-deploy a mock aggregator with fewer decimals
```
tt-krzysztof, line 87: Move comment next to deploy statement.

test/true-currencies/TrueCurrencyWithPoR.test.ts
```diff
@@ -83,38 +83,40 @@ describe('TrueCurrency with Proof-of-reserves check', () => {
...
-    })
-    expect(await token.balanceOf(owner.address)).to.equal(balanceBefore.add(AMOUNT_TO_MINT))
+    })).to.be.revertedWith('Unexpected decimals of PoR feed')
+    expect(await token.balanceOf(owner.address)).to.equal(balanceBefore)
   })
 
-  it('should mint successfully when feed decimals > TrueCurrency decimals', async () => {
-    // Re-deploy a mock aggregator with more decimals
+  it('should revert mint when feed decimals > TrueCurrency decimals', async () => {
+    // Re-deploy a mock aggregator with fewer decimals
```
tt-krzysztof, line 105:
```suggestion
    // Re-deploy a mock aggregator with more decimals
```

test/true-currencies/TrueCurrencyWithPoR.test.ts
```diff
@@ -83,38 +83,40 @@ describe('TrueCurrency with Proof-of-reserves check', () => {
...
     const currentTusdSupply = await token.totalSupply()
-    const validReserve = currentTusdSupply.mul(exp(1, 2)).add(AMOUNT_TO_MINT)
-    const mockV3AggregatorWith20Decimals = await new MockV3Aggregator__factory(owner).deploy('20', validReserve)
+    const validReserve = currentTusdSupply.div(exp(1, 12)).add(AMOUNT_TO_MINT)
+
+    const mockV3AggregatorWith6Decimals = await new MockV3Aggregator__factory(owner).deploy('20', validReserve)
```
tt-krzysztof, line 109:
```suggestion
    const mockV3AggregatorWith20Decimals = await new MockV3Aggregator__factory(owner).deploy('20', validReserve)
```

tt-krzysztof (approved):

yuchenlintt (approved):

PR #1188, opened 2022-09-08
⏸️ Add Proof of Reserve pausing

pkuchtatt:


contracts/true-currencies/TokenController.sol
```diff
@@ -86,6 +86,8 @@ contract TokenController {
     // pausing the contract upgrades the proxy to this implementation
     address public constant PAUSED_IMPLEMENTATION = 0x3c8984DCE8f68FCDEEEafD9E0eca3598562eD291;
```
yuchenlintt, line 87: I just noticed there exists this other pause semantic.

Could we rename pause/unpause PoR to enable/disable PoR? This way there's no possible confusion between pausing TUSD and pausing PoR on TUSD.

contracts/true-currencies/TokenController.sol
```diff
@@ -86,6 +86,8 @@ contract TokenController {
...
 
+    address public proofOfReservePauser;
```
yuchenlintt, line 89: The `public constant` fields above don't take up any storage slots (constants are compiled into the contract source code), so let's move this address up, just after `gasRefunder`.

contracts/true-currencies/TrueCurrencyWithPoR.sol
```diff
@@ -30,7 +28,7 @@ abstract contract TrueCurrencyWithPoR is TrueCurrency, IPoRToken {
      * @param amount The amount of tokens to mint
      */
     function _mint(address account, uint256 amount) internal virtual override {
-        if (chainReserveFeed == address(0)) {
+        if (chainReserveFeed == address(0) || !proofOfReserveEnabled) {
             super._mint(account, amount);
             return;
         }
```
yuchenlintt, line 34: nitpick: it's probably better to keep only one canonical way to enable/disable PoR.
```suggestion
        if (!proofOfReserveEnabled) {
            super._mint(account, amount);
            return;
        }
        require(chainReserveFeed != address(0), "TODO err msg");
```

yuchenlintt (changes requested):
Approach looks great! Also flexible: we can decide later what address can enable/disable PoR.

pkuchtatt:
> Approach looks great! Also flexible: we can decide later what address can enable/disable PoR.

Thanks. To clarify, the intention is to set the `proofOfReservePauser` (or `proofOfReserveEnabler` if we do the suggested renaming from "pause" to "enable" consistent in the code) to a Gnosis Safe multi-sig wallet, as discussed on slack, so we can let people working on TUSD operations to pause or temporarily disable PoR. They will be able to disable and enable PoR safely, without a risk of setting `chainReserveFeed` to a wrong address.

However, the initial method of pausing PoR, calling `setChainReserveFeed(0)` and then `setChainReserveFeed(<feed address>)` has not been eliminated. It is still in place, though still can be executed only by the owner of the contract.

contracts/true-currencies/TokenController.sol
pkuchtatt: Good point about the inconsistent naming! It is an effect of discussing the feature with various people: operations people preferred `pause`, whereas when discussing it with engineers `proofOfReserveEnabled` seemed a more natural name of a variable that needs to be used to implement the feature. To be consistent I will also rename `proofOfReservePauser` to `proofOfReserveEnabler`, if that's OK.

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: When I first saw the suggested change I thought it was really neat! However, after thinking about it a bit more I came to the following conclusions.

1. When we realised that disabling (temporarily) PoR will be needed often and will be performed by operations people that manage minting as well as USD funds allocation across escrow accounts on a daily basis, rather than the owner account only, which is of the highest security and normally only used for upgrading the contract, we came up with an *additional*, safe method to disable temporarily PoR.
2. The original method of disabling PoR by setting `chainReserveFeed` to `0` stays in place. Still, only the owner of the contract can use it.
3. The suggested change does not eliminate any of the two available methods to disable PoR, it only causes the mint to be reverted if the two methods were used inconsistently: the owner of TUSD has not enabled PoR, yet the other method was used to enable PoR - perhaps by one of the "PoR enablers" or by the owner himself, by a mistake. In this situation I think it is right to accept authority of the owner and consider PoR as disabled, rather than consider it a situation of an inconsistent state of the contract that justifies reverting mints.

So, I think `chainReserveFeed == address(0) || !proofOfReserveEnabled` evaluated to `true` correctly means that PoR is disabled. 

I admit it doesn't feel right that the state of the contract can be somewhat inconsistent when `chainReserveFeed` is set to a non-zero address and `proofOfReserveEnabled` is `true` so I propose that:
- in `setChainReserveFeed` when the new address is `0` we also set `proofOfReserveEnabled` to `false`,
- in `_,mint` when we detect the inconsistent state we emit an event rather than revert, so it is easier to notice the situation.

What do you think?

yuchenlintt: > In this situation I think it is right to accept authority of the owner and consider PoR as disabled, rather than consider it a situation of an inconsistent state of the contract that justifies reverting mints.

I don't have a strong opinion on this :) Your reasoning here makes sense to me, so I think we can stay with the current implementation and have two ways to pause.

> in setChainReserveFeed when the new address is 0 we also set proofOfReserveEnabled to false,
> in _,mint when we detect the inconsistent state we emit an event rather than revert, so it is easier to notice the situation.

An alternate proposal:
- `setChainReserveFeed(0)` -> `proofOfReserveEnabled = false;`
- `enableProofOfReserve() { require(chainReserveFeed != address(0)); ... }`

This way it shouldn't ever be possible to have `proofOfReserveEnabled == true` while `chainReserveFeed` is 0. How does that look?

yuchenlintt: Ah, I see the impl has already changed to this! Looks good.

contracts/true-currencies/TokenController.sol
```diff
@@ -82,6 +82,8 @@ contract TokenController {
     bytes32 public constant IS_MINT_RATIFIER = "isTUSDMintRatifier";
     bytes32 public constant IS_REDEMPTION_ADMIN = "isTUSDRedemptionAdmin";
 
+    address public proofOfReserveEnabler;
```
yuchenlintt, line 85: Looks like you didn't move this quite high enough. There are still constants mixed in with the storage variables here. Could you move this between `gasRefunder` and `IS_MINT_PAUSER`?

yuchenlintt (changes requested):
Just one more nitpick! Otherwise I think this LGTM.

contracts/true-currencies/TokenController.sol
pkuchtatt: Done.

contracts/true-currencies/TrueCurrencyWithPoR.sol
pkuchtatt: Yep! Can you please review again and approve if you're ok with all changes?

msieczko:
There is still a problem in this PR. We are applying our changes to `TokenController` contract (v1) and on mainnet `TokenControllerV2` is used. I believe we should add `TokenControllerV2` contract to the repo, rename it to `TokenControllerV3` and reapply our changes. @yuchenlintt does that sound good to you as well?

msieczko:
Also, just discovered that [Avalanche TUSD](https://snowtrace.io/token/0x1C20E891Bab6b1727d14Da358FAe2984Ed9B59EB) uses it's own version of [`AvalancheTokenController`](https://snowtrace.io/address/0x0008fad7b0eac3f863498a864c037dc4a6a113e1#code). So before deploying to Avalanche we need to reaply changes there as well.

msieczko:
I noticed that the [currently used TUSD implementation](https://etherscan.io/address/0xd8d59c59ab40b880b54c969920e8d9172182ad7b#code) differs from the `TrueUSDWithPoR` contract that we plan to deploy. Let's do an analysis what changes between those version to make sure that new TrueUSD implementation does not break things.

contracts/true-currencies/TokenController.sol
pkuchtatt: Renamed "pause/unpause" to "enable/disable"

contracts/true-currencies/TokenController.sol
pkuchtatt: Done.

yuchenlintt (dismissed):
I'm okay with checking this in for now, provided in follow-up PRs we propagate all relevant changes to TokenControllerV2/3 and make sure any diffs against deployed contracts are exactly as expected.

Approving!

yuchenlintt:
> Which expected functions do you have in mind? I think we should test any function that changes with the new implementation.

Yeah, that sounds good to me! Another area of concern would be to check behavior related to storage variables that are within 2-3 slots of anything we change here. For this TokenController implementation, that would include functions related to `registryAdmin` and `gasRefunder` (I'm not sure what these do...), just in case there's an unexpected shift in slots caused by an undocumented deletion/addition of a storage variable at some point in the past.

PR #1194, opened 2022-09-23
⚙️ Build and test contracts-por package on CI

pkuchtatt:


.circleci/config.yml
```diff
@@ -48,6 +48,7 @@ jobs:
           at: .
       - run: yarn build
       - run: yarn typecheck
+      - run: for d in packages/*; do cd $d && yarn build; done
```
msieczko, line 51: You usually use tools like [wsrun](https://www.npmjs.com/package/wsrun) instead of looping like this. In this case I'd simply use:
```suggestion
      - yarn workspace @trusttoken-smart-contracts/contracts-por build
```

.circleci/config.yml
```diff
@@ -66,6 +67,13 @@ jobs:
       - attach_workspace:
           at: .
       - run: yarn lint
+  test-packages:
+    docker:
+      - image: cimg/node:16.1.0
+    steps:
+      - attach_workspace:
+          at: .
+      - run: for d in packages/*; do cd $d && yarn test; done
```
msieczko, line 76:
```suggestion
      - yarn workspace @trusttoken-smart-contracts/contracts-por test
```

msieczko (changes requested):

msieczko (approved):

PR #1207, opened 2022-10-26
🧪 Add TokenController tests to package contracts-por

pkuchtatt:
Add tests for old TokenController functionality by adapting and fixing relevant tests from repo's root.


msieczko (changes requested):
See comments. Otherwise LGTM.

packages/contracts-por/test/TokenController.test.ts
```diff
@@ -0,0 +1,592 @@
...
+    await registry.setAttribute(ratifier2.address, formatBytes32String('isTUSDMintRatifier'), 1, notes, { gasLimit: 5_000_000 })
+    await registry.setAttribute(ratifier3.address, formatBytes32String('isTUSDMintRatifier'), 1, notes, { gasLimit: 5_000_000 })
+    await registry.setAttribute(pauseKey.address, formatBytes32String('isTUSDMintPausers'), 1, notes)
+    await controller.setMintThresholds(parseEther('200'), parseEther('1000'), parseEther('1001'))
+    await controller.setMintLimits(parseEther('200'), parseEther('300'), parseEther('3000'))
+    await controller.refillMultiSigMintPool()
+    await controller.refillRatifiedMintPool()
+    await controller.refillInstantMintPool()
+    await controller.instantMint(thirdWallet.address, parseEther('100'))
+    await controller.setMintLimits(0, 0, 0)
```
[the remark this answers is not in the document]
pkuchtatt: Used `beforeEachWithFixture` fixture from the root package. It uses `hardhatCallHistoryProvider` which doesn't work in the contracts-por, for some reason, so I commented out the call history, as it doesn't seem to be essential (looks like a debugging tool). Tests do pass quite a bit faster!

packages/contracts-por/package.json
```diff
@@ -5,13 +5,20 @@
...
+    "lint:fix": "prettier 'contracts/**/*.sol' --write --loglevel error && yarn lint:ts --fix",
     "prebuild": "yarn clean",
     "build:hardhat": "hardhat compile",
     "build:typechain": "typechain --target ethers-v5 --out-dir build/types 'build/*.json'",
     "build": "yarn build:hardhat && yarn build:typechain && mars",
     "preflatten": "rm -rf custom_flatten",
     "flatten": "waffle flatten .waffle.json",
-    "test": "mocha 'test/**/*.test.ts'"
+    "test": "mocha 'test/**/*.test.ts'",
+    "checks": "yarn lint && yarn test && yarn test:integration"
```
msieczko, line 21: There's no `test:integration` script in this `package.json file`. Looks like a copy-paste error.

packages/contracts-por/package.json
```diff
@@ -5,13 +5,20 @@
   "private": true,
   "scripts": {
     "clean": "rm -rf ./build && hardhat clean",
+    "lint": "yarn lint:sol && yarn lint:ts",
+    "typecheck": "tsc --noEmit",
+    "lint:sol": "solhint 'contracts/**/*.sol' && prettylint 'contracts/**/*.sol'",
+    "lint:js": "eslint '{test,scripts}/**/*.js'",
```
msieczko, line 11: We only code in TypeScript, so linting JavaScript is redundant.

packages/contracts-por/package.json
```diff
@@ -5,13 +5,20 @@
   "private": true,
   "scripts": {
     "clean": "rm -rf ./build && hardhat clean",
+    "lint": "yarn lint:sol && yarn lint:ts",
+    "typecheck": "tsc --noEmit",
+    "lint:sol": "solhint 'contracts/**/*.sol' && prettylint 'contracts/**/*.sol'",
```
msieczko, line 10: `solhint`, `prettylint` and `prettier` tools are used but not listed on the list of `devDependencies`. They only work because they are installed by the parent `package.json` file. If someone removes or changes the version there, `contracts-por` package breaks. Always declare all used dependencies to avoid such problems.

packages/contracts-por/test/utils/hardhatPatches.ts
```diff
@@ -1,10 +1,11 @@
+// @ts-ignore
```
msieczko, line 1: Including hardhat.config.ts in files that are looked at by the TypeScript compiler solves the problem with import. Source: https://github.com/NomicFoundation/hardhat/issues/1627#issuecomment-872713053.

packages/contracts-por/contracts/Registry.sol
```diff
@@ -0,0 +1,185 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.6.10;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import {IRegistryClone as RegistryClone} from "./interface/IRegistryClone.sol";
+
+contract Registry {
```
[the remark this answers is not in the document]
msieczko: What about `IRegistryClone.sol`? I'd move it as well.

packages/contracts-por/test/TokenController.test.ts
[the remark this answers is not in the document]
msieczko: `hardhatCallHistoryProvider` is not needed as you noticed, so let's remove it. Never leave commented-out code.

msieczko (changes requested):

packages/contracts-por/test/TokenController.test.ts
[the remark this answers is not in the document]
pkuchtatt: Removed `hardhatCallHistoryProvider`.

packages/contracts-por/contracts/Registry.sol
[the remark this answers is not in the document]
pkuchtatt: Done.

packages/contracts-por/package.json
pkuchtatt: Done.

packages/contracts-por/package.json
pkuchtatt: Removed now.

packages/contracts-por/test/utils/hardhatPatches.ts
pkuchtatt: Done, thanks!

msieczko (approved):

PR #1210, opened 2022-11-04
[A4] Remove redundant address casting

msieczko:
@yuchenlintt Can you take a look at this one? We're not sure what was the purpose of `address(uint160(address(token)))` casting. Auditor hinted that _it is used to cast the address of token to a payable one_. We don't see a point in doing that though or maybe we are missing something?


yuchenlintt (approved):
Yeah, I don't see the point of quadruple-casting either. Two casts should be enough to do an external function call on `token`. Good catch!

pkuchtatt (approved):

pkuchtatt (approved):

PR #1218, opened 2022-11-22
5️⃣ Implement audit recommendation A5

pkuchtatt:


packages/contracts-por/contracts/TrueCurrencyWithProofOfReserve.sol
```diff
@@ -39,7 +39,7 @@ abstract contract TrueCurrencyWithProofOfReserve is TrueCurrency, IProofOfReserv
         uint256 reserves = uint256(signedReserves);
 
         // Sanity check: is chainlink answer updatedAt in the past
-        require(block.timestamp >= updatedAt);
+        require(block.timestamp >= updatedAt, "TrueCurrency: invalid PoR updatedAt");
```
msieczko, line 42: We're missing a test for this require statement.

msieczko (changes requested):

packages/contracts-por/contracts/TrueCurrencyWithProofOfReserve.sol
pkuchtatt: Added test.

msieczko (approved):

PR #1234, opened 2023-03-08
🏄‍♂️ Use XC-20 standard for Watr `TrueCurrency`

abam-iksde:


packages/contracts-watr/test/verifyDeployment/verifyDeployment.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    await waitFor(tokenController.instantMint(deployer.address, parseEther('1')))
+
+    await waitFor(token.transfer(otherAddress, parseEther('0.5')))
+
+    const balancesAfter = [
+      await token.balanceOf(deployer.address),
+      await token.balanceOf(otherAddress),
+    ]
+
+    expect(balancesAfter[0].sub(balancesBefore[0]).toString()).to.eq(parseEther('0.5').toString())
```
tt-krzysztof, line 81: We can use [waffle balance matchers](https://ethereum-waffle.readthedocs.io/en/latest/matchers.html#change-token-balance-multiple-accounts) (here and in the previous test).

packages/contracts-watr/test/verifyDeployment/verifyDeployment.test.ts
```diff
@@ -0,0 +1,88 @@
+import { Contract, ethers } from 'ethers'
+import { JsonRpcProvider } from '@ethersproject/providers'
+import { expect } from 'chai'
+import { unknown as deployments } from '../../deployments-watr_local.json'
+import { TokenControllerV3__factory, TrueUSD__factory } from 'contracts'
+import { parseEther } from '@ethersproject/units'
+
+describe('verify deployment', () => {
+  const provider = new JsonRpcProvider('http://127.0.0.1:8822', 688)
+  const ownableInterface = [
```
tt-krzysztof, line 10: I wanted to suggest we use one of the Typechain-generated factories, but apparently, there are none like this. Which perhaps suggests we have a Solidity interface missing somewhere? 🤔

packages/contracts-watr/test/verifyDeployment/verifyDeployment.test.ts
```diff
@@ -0,0 +1,88 @@
+import { Contract, ethers } from 'ethers'
+import { JsonRpcProvider } from '@ethersproject/providers'
+import { expect } from 'chai'
+import { unknown as deployments } from '../../deployments-watr_local.json'
+import { TokenControllerV3__factory, TrueUSD__factory } from 'contracts'
+import { parseEther } from '@ethersproject/units'
+
+describe('verify deployment', () => {
+  const provider = new JsonRpcProvider('http://127.0.0.1:8822', 688)
```
tt-krzysztof, line 9: Extract to constants.

packages/contracts-watr/contracts/interface/IERC20Plus.sol
```diff
@@ -0,0 +1,127 @@
...
+    function mint(address beneficiary, uint256 amount) external returns (bool);
+
+    /**
+     * @dev Burns by up to the specified amount of asset from the target.
+     * This operation will increase decrease the total supply.
+     * Only usable by asset admin.
+     * Selector: 9dc29fac
+     */
+    function burn(address who, uint256 amount) external returns (bool);
+}
```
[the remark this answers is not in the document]
abam-iksde: that would mean we have to modify the Watr itself, I wanted to do add that in a possible follow up PR

packages/contracts-watr/test/verifyDeployment/verifyDeployment.test.ts
tt-krzysztof: Yes, I guess it can stay like this ✅

packages/contracts-watr/contracts/interface/IERC20Plus.sol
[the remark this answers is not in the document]
tt-krzysztof: Fair enough.

packages/contracts-watr/contracts/common/ProxyStorage.sol
```diff
@@ -61,6 +61,10 @@ contract ProxyStorage {
     address public chainReserveFeed;
     bool public proofOfReserveEnabled;
 
+    // XC20Wrapper variables
+    address public nativeToken;
+    mapping(address => mapping(address => uint256)) allowances;
```
tt-krzysztof, line 66: We could keep using the ERC20 `_allowances` field (L39).

abam-iksde: I updated it as you suggested, although since we're deploying the contracts to a new chain we could consider stripping the `ProxyStorage` of all the legacy stuff

packages/contracts-watr/contracts/TrueCurrency.sol
```diff
@@ -126,8 +130,9 @@ abstract contract TrueCurrency is BurnableTokenWithBounds {
         require(!isBlacklisted[recipient], "TrueCurrency: recipient is blacklisted");
 
         if (isRedemptionAddress(recipient)) {
-            super._transfer(sender, recipient, amount.sub(amount.mod(CENT)));
-            _burn(recipient, amount.sub(amount.mod(CENT)));
+            uint256 _amount = amount.sub(amount.mod(CENT));
```
akoszowski, line 133: Good change, though we don't really need SafeMath here as it it always the case that for `uint256` that `amount >= amount mod CENT `. Not sure if we also want to avoid changes like this 🤔

packages/contracts-watr/scripts/deployment/deployToken.ts
```diff
@@ -14,7 +15,7 @@ export function deployToken<T extends Token>(tokenArtifact: ArtifactFrom<T>, con
     proxy.transferProxyOwnership(controller)
   })
   const proxy = tokenProxy(implementation, (token) => {
-    token.initialize()
+    token.initialize(generatePrecompileAddress(1983))
```
akoszowski, line 18: Maybe we could move this `1983` to some `WATR` config file or so? Are we sure it this number will be our `assetId`?

packages/contracts-watr/contracts/common/XC20Wrapper.sol
```diff
@@ -0,0 +1,108 @@
...
+        address sender,
+        address recipient,
+        uint256 amount
+    ) internal {
+        require(IERC20Plus(nativeToken).balanceOf(sender) >= amount, "XC20: amount exceeds balance");
+        IERC20Plus(nativeToken).burn(sender, amount);
+        IERC20Plus(nativeToken).mint(recipient, amount);
+    }
+
+    function name() public pure virtual returns (string memory) {
```
akoszowski, line 101: Can it be just ?
```
function name() public virtual pure returns (string memory);
```

packages/contracts-watr/contracts/common/XC20Wrapper.sol
akoszowski, line 1: General things: it would be easier to read this code if it were consistent with `IERC20` interface when it comes to the order of functions

packages/contracts-watr/contracts/common/XC20Wrapper.sol
```diff
@@ -0,0 +1,108 @@
...
+        emit Transfer(address(0), account, amount);
+    }
+
+    function _burn(address account, uint256 amount) internal virtual {
+        IERC20Plus(nativeToken).burn(account, amount);
+        emit Transfer(account, address(0), amount);
+    }
+
+    function decimals() public view virtual returns (uint8) {
+        return IERC20Plus(nativeToken).decimals();
```
akoszowski, line 27: It is later overwritten by `18` in `TrueUSD.sol` is that okay? Maybe we don't need to call this method on `nativeToken` or should not override in `TrueUSD.sol`?

packages/contracts-watr/contracts/common/XC20Wrapper.sol
```diff
@@ -0,0 +1,108 @@
...
+    function totalSupply() public view virtual override returns (uint256) {
+        return IERC20Plus(nativeToken).totalSupply();
+    }
+
+    function allowance(address owner, address spender) external view override returns (uint256) {
+        return _allowances[owner][spender];
+    }
+
+    function approve(address spender, uint256 amount) external virtual override returns (bool) {
+        _approve(msg.sender, spender, amount);
```
akoszowski, line 39: Shouldn't we use `_msgSender()`?

packages/contracts-watr/contracts/common/XC20Wrapper.sol
```diff
@@ -0,0 +1,108 @@
...
+
+    function decimals() public view virtual returns (uint8) {
+        return IERC20Plus(nativeToken).decimals();
+    }
+
+    function totalSupply() public view virtual override returns (uint256) {
+        return IERC20Plus(nativeToken).totalSupply();
+    }
+
+    function allowance(address owner, address spender) external view override returns (uint256) {
```
akoszowski, line 34: Has not `virtual`

PR #1242, opened 2023-03-22
⭐  feat: Add IBEP20 and Change ERC20 to BEP20

qiwihui:

PR #1245, opened 2023-03-22
🍚  Fix: Manage claimable ownership of TUSD

qiwihui:


packages/contracts-bsc/contracts/BscTokenController.sol
```diff
@@ -70,16 +71,13 @@ contract BscTokenController {
 
     ITrueCurrency public token;
     IRegistry public registry;
-    address public fastPause; // deprecated
-    address public trueRewardManager; // deprecated
```
yuchenlintt, line 74: Just like with #1242, let's separate {removal of deprecated storage variables} and {adding new features} into different PRs.

I would have been fine approving and merging the removal, but (re)adding functionality here has more scrutiny.

packages/contracts-bsc/contracts/BscTokenController.sol
```diff
@@ -533,10 +534,10 @@ contract BscTokenController {
...
 
     /**
-     * @dev Claim ownership of an arbitrary HasOwner contract
-
+     * @dev Claim ownership of an arbitrary IHasOwner contract
+     */
     function issueClaimOwnership(address _other) public onlyOwner {
-        HasOwner other = HasOwner(_other);
+        IHasOwner other = IHasOwner(_other);
         other.claimOwnership();
```
yuchenlintt, line 541: I assume this function exists to claim impl ownership of `token`? If so, then I have some suggestions to improve the original code:
- for consistency with the `proxy functions` section above, I would rename this function to `claimTrueCurrencyImplOwnership()` and move it into its own section immediately below `proxy functions`:
```
    /*
    ========================================
    impl functions
    ========================================
    */
```
- drop the `address _other` argument. There's no need to support other arbitrary addresses if the controller will only ever claim `token` impl ownership. Ideally, the body of this function should just be `token.claimOwnership()`.
- no other functions in `BscTokenController.sol` call this, so this can safely be marked `external`.

I would also suggest making these changes in a separate PR so unrelated changes can be merged.

packages/contracts-bsc/contracts/BscTokenController.sol
```diff
@@ -533,10 +534,10 @@ contract BscTokenController {
...
     }
```
yuchenlintt, line 543: If you're bringing back `claimTrueCurrencyImplOwnership()`, then, for symmetry, I would also bring back the commented out function below (and move it to an `impl functions` section, see above comment). Something like this? Should be done in the same PR.

```
function transferTrueCurrencyImplOwnership(address _newOwner) external onlyOwner {
    token.transferOwnership(_newOwner);
}
```

packages/contracts-bsc/contracts/BscTokenController.sol
```diff
@@ -610,6 +611,7 @@ contract BscTokenController {
      */
     function setCanBurn(address burner, bool canBurn) external onlyRegistryAdminOrOwner {
         token.setCanBurn(burner, canBurn);
+        emit CanBurn(burner, canBurn);
```
yuchenlintt, line 614: I noticed that you added an event here in `BscTokenController`. Based on the coding style of these contracts, I believe it would be more appropriate to emit the event from inside `TrueCurrency` than here, for two reasons:
- other `token` composition functions in `BscTokenController` do not emit events, so this would be inconsistent style
- no direct contract storage gets changed for `BscTokenController` due to this function, but contract storage _does_ get changed in `TrueCurrency` (which doesn't currently emit an event for `token.setCanBurn()`).

I would also recommend making this change in a separate PR.

yuchenlintt (changes requested):

packages/contracts-bsc/contracts/BscTokenController.sol
qiwihui: will separate it to a new pr

packages/contracts-bsc/contracts/interface/IHasOwner.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.6.10;
+
+interface IHasOwner {
```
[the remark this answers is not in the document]
qiwihui: `IHasOwner` interface is used for `issueClaimOwnership`, which will be used when using TokenController to claim the ownership of TrueUSD contract when first deployed on a new chain.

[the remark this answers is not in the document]
qiwihui: Since it was deleted in por code but we need `issueClaimOwnership`, so I added `IHasOwner` interface according to POR changes.

packages/contracts-bsc/contracts/BscTokenController.sol
qiwihui: There is a pair of functions `issueClaimOwnership` and `transferChild` to manage the ownership of `token`(or some other token maybe, I have no idea about the original design). I think it would be better to keep the name. But I will comment out transferChild in case we need it.

packages/contracts-bsc/contracts/BscTokenController.sol
qiwihui: will separate it

qiwihui:
#1233

packages/contracts-bsc/contracts/BscTokenController.sol
yuchenlintt: Thanks for splitting it!

However, I noticed that your new PR #1248 still has changes that come from this PR. Ideally PRs capture independent changes, so we can review and merge one PR without necessarily waiting for another PR to pass review. As of now, I'm unwilling to begin reviewing another PR that requires this one to be merged first, especially if there are unresolved comments in this PR whose fixes will need to be propagated.

packages/contracts-bsc/contracts/BscTokenController.sol
yuchenlintt: I am also uncertain about the original intent -- I wasn't at Archblock when this design decision was made.

However, what I can say is that functions should have descriptive names that are consistent with the local style. My best judgment is that `issueClaimOwnership()` + `transferChild()` must have been remnants of previous naming refactors, but were skipped because they were commented out.

However, as of _now_, these two functions are:
- no longer descriptive (`child` does not appear to be used anywhere else in this contract, so it's not obvious from the name what it refers to -- we can _infer_ that it refers to `token`, but, if so, then we should definitely _not_ keep the name `child`),
- nor consistent with local style (no other function is named `issueX()` -- there are still functions named `requestX()` for `token.X()` and `X()` for `token.X()`, but none named `issueX()`).

If we want to adhere to descriptiveness and consistency, then we should name these two functions something like `claimTrueCurrencyImplOwnership()` + `transferTrueCurrencyImplOwnership()`, and, as I mention in [this comment](https://github.com/trusttoken/contracts-pre22/pull/1245#discussion_r1145470568), move these functions to a new `impl functions` section below `proxy functions` so that they correspond with the existing `claimTrueCurrencyProxyOwnership()` and `transferTrueCurrencyProxyOwnership()` functions.

packages/contracts-bsc/contracts/interface/IHasOwner.sol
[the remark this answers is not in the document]
yuchenlintt: The missing `HasOwner` contract comes from commented-out code, which may have survived several naming refactors that left it out-of-date with the current repo.

I would not base a _current_ naming decision on _previous_ names that originally came from commented-out code.

Instead, I would advise that we determine our current needs (claiming/transferring impl ownership of a TrueCurrency `token` contract) and write descriptive and consistent code to directly address these needs.

We can delete the commented-out code; it is no longer relevant and should not inform our naming decisions.

yuchenlintt:
Thanks for bearing with me! Also, a reminder to update the name of this PR to reflect the changes that are made in here.

packages/contracts-bsc/contracts/BscTokenController.sol
qiwihui: Thanks for your great advice.

packages/contracts-bsc/contracts/BscTokenController.sol
```diff
@@ -223,6 +225,28 @@ contract BscTokenController {
...
+     * @dev Claim ownership of token contract
+     */
+    function claimTrueCurrencyImplOwnership() external onlyOwner {
+        token.claimOwnership();
+    }
+
+    /**
+     * @dev Transfer ownership of token to _newOwner.
+     * Can be used e.g. to upgrade this TokenController contract.
+     * @param _newOwner new owner/pending owner of _child
```
yuchenlintt, line 244:
```suggestion
     * @param _newOwner new owner/pending owner of token
```

packages/contracts-bsc/contracts/BscTokenController.sol
```diff
@@ -109,14 +110,15 @@ contract BscTokenController {
...
+
     /// @dev Emitted when ownership of controller was transferred
     event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
     /// @dev Emitted when ownership of controller transfer procedure was started
     event NewOwnerPending(address indexed currentOwner, address indexed pendingOwner);
     /// @dev Emitted when new registry was set
     event SetRegistry(address indexed registry);
     /// @dev Emitted when owner was transferred for child contract
-    // event TransferChild(address indexed child, address indexed newOwner);
+    event TransferChild(address indexed child, address indexed newOwner);
```
yuchenlintt, line 121:
```suggestion
```

yuchenlintt (approved):
Looks great now, thank you for updating this!

Just two minor suggestions that I think you might've overlooked.

packages/contracts-bsc/contracts/BscTokenController.sol
```diff
@@ -109,14 +110,15 @@ contract BscTokenController {
...
         _;
     }
+
     /// @dev Emitted when ownership of controller was transferred
     event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
     /// @dev Emitted when ownership of controller transfer procedure was started
     event NewOwnerPending(address indexed currentOwner, address indexed pendingOwner);
     /// @dev Emitted when new registry was set
     event SetRegistry(address indexed registry);
     /// @dev Emitted when owner was transferred for child contract
```
yuchenlintt, line 120:
```suggestion
```

packages/contracts-bsc/contracts/BscTokenController.sol
yuchenlintt, line 121:
```suggestion
```

qiwihui: Cool. Thanks.

yuchenlintt (approved):

yuchenlintt (approved):

